PR #8, opened 2026-06-29
add OMTD core library

mirek190:
## Summary

Adds the OMTD core library as a standalone output-modality boundary for future generated-media backends.

OMTD is the output-modality counterpart to MTMD: it keeps generated-media output routing behind a separate companion path instead of overloading input-media projector semantics.

## Notes

- Scope is OMTD core only.
- Model-specific backends are intentionally left for follow-up patches.
- This branch was prepared from `mirek190/llama.cpp` OMTD core commit `06c714efbfc5cc544a3dbee134c1e3368adabc05`, adapted onto `audio.cpp:release-0.1` by omitting llama.cpp aggregate CMake files that do not exist in audio.cpp.

## Validation

- Confirmed final branch diff contains only OMTD docs/source/test files.
- No audio.cpp CI has been run locally for this branch.

### DIFF
--- a/docs/omtd.md
+++ b/docs/omtd.md
@@ -1,0 +1,117 @@
+﻿# Output Multimodal Transfer/Decoder
+
+OMTD is a proposed output-side companion library for llama.cpp.
+
+It mirrors MTMD conceptually, but runs in the opposite direction:
+
+```text
+MTMD: media input -> embeddings/tokens -> language model
+OMTD: language model output -> output companion -> generated media
+```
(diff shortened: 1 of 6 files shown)


0xShug0 (approved):
Currently, it’s unclear to me how OMTD interacts with the audio.cpp framework. I haven't decided whether to integrate llama.cpp for AR audio models yet. But since the changes are additive and mostly placeholders, I’ll merge them.

PR #14, opened 2026-07-01
Add VibeVoice 7B support and decoder LoRA merging

justinjohn0306:
Support the VibeVoice 7B model alongside the 1.5B:
- Add the vibevoice_7b model-manager package, reusing the shared Qwen2.5 tokenizer bundle.
- Handle the 7B config.json, tolerating its upstream "acostic_vae_dim" key and reading top-level tie_word_embeddings.
- Load a separate lm_head.weight when word embeddings are untied and bind the decoder logits head to it.

Add optional PEFT LoRA merging for the decoder:
- Merge lora_A/lora_B into the targeted linear weights at load time via a tensor-source decorator, so it composes with the weight-type quantization options at no per-step cost.
- Expose it through the vibevoice.lora / vibevoice.lora_scale load options and document it in the README and docs/tts.md.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,7 @@ add_compile_options(
 add_library(engine_runtime STATIC
     src/framework/assets/resource_bundle.cpp
     src/framework/assets/tensor_source.cpp
+    src/framework/assets/torch_bin.cpp
     src/framework/core/module.cpp
     src/framework/core/backend.cpp
     src/framework/core/deferred_tensor_writer.cpp
@@ -413,6 +414,24 @@ add_executable(miocodec_wavlm_parity
 )
 
 target_link_libraries(miocodec_wavlm_parity PRIVATE engine_runtime ggml)
+
+add_executable(torch_bin_parity
+    tests/vibevoice/torch_bin_parity.cpp
+)
+
+target_link_libraries(torch_bin_parity PRIVATE engine_runtime ggml)
+if (ENGINE_ENABLE_OPENMP)
+    target_link_libraries(torch_bin_parity PRIVATE OpenMP::OpenMP_CXX)
+endif()
+
+add_executable(vibevoice_finetune_overlay_check
+    tests/vibevoice/finetune_overlay_check.cpp
+)
+
+target_link_libraries(vibevoice_finetune_overlay_check PRIVATE engine_runtime ggml)
+if (ENGINE_ENABLE_OPENMP)
+    target_link_libraries(vibevoice_finetune_overlay_check PRIVATE OpenMP::OpenMP_CXX)
+endif()
 
 if (ENGINE_BUILD_WARMBENCH)
     function(add_engine_warmbench target_name source_file)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,6 +219,7 @@ add_library(engine_runtime STATIC
     src/models/vibevoice/diffusion_sampler.cpp
     src/models/vibevoice/generator.cpp
     src/models/vibevoice/loader.cpp
+    src/models/vibevoice/lora.cpp
     src/models/vibevoice/scheduler.cpp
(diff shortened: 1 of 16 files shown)


0xShug0:
@justinjohn0306 Thank you for your PR! Could you add extra logs to confirm the LoRA loading is woring with 7B model and share the run log? 

Reviewed by Codex:
[docs/tts.md:392](https://github.com/0xShug0/audio.cpp/pull/14/commits/981f9df27cac5e3cf9139f1cea85432164793d46) and README news say to use --load-option vibevoice.lora, but the implementation only reads vibevoice.lora from SessionOptions in src/models/vibevoice/session.cpp:34. The CLI stores --load-option separately in load_request.options, while session_options.options only receives --session-option. Since the VibeVoice loader does not consume load_request.options, users following the docs will run without any LoRA merge and get no error

justinjohn0306:
## LoRA loading confirmed on VibeVoice-7B + `--load-option` fix

Thanks for the review — addressed both points, and extended the adapter support to match `infer.py`.

### 1. `--load-option vibevoice.lora` now actually merges (Codex finding)

Previously the adapter was only read from `SessionOptions`, so the **documented** `--load-option` path silently no-op'd. The loader now consumes `request.options` at load time via a shared `apply_vibevoice_finetune_options(...)` helper, and the same helper still handles `--session-option` for backward compatibility. A guard rejects passing it through both at once. So `docs/tts.md` / README (`--load-option vibevoice.lora=<dir>`) are now correct.

### 2. Merge logging + 7B run log

Added `--log` output confirming each component. Running **VibeVoice-7B** with the fine-tune via `--load-option`:

```
[info][vibevoice] applying fine-tune adapter: ...\finetune_mp1\lora
[info][vibevoice] LM LoRA: merged 196 decoder modules (scale 4.000000)
[info][vibevoice] diffusion head: overrode 26 tensors
[info][vibevoice] acoustic connector: overrode 5 tensors
[info][vibevoice] semantic connector: overrode 5 tensors
```

### 3. Confirmed it changes the output toward the fine-tuned speaker

A/B with a fixed seed and identical reference audio — base weights vs the fine-tune adapter — the two outputs correlate **0.956** (not identical), difference RMS ≈ 30% of the signal: same words/prosody, but the voice timbre shifts onto the target speaker. It audibly matches the reference and the Python `infer.py` output with the same adapter.

### Bonus: full fine-tune overlay (not just the LM LoRA)

To mirror `infer.py`'s `apply_lora`, the overlay now applies all four trained components when present in the adapter dir:

- **language-model LoRA** → delta-merged into the decoder linears
- **fine-tuned diffusion head** → replaces `model.prediction_head.*`
- **acoustic + semantic connectors** → replace `model.{acoustic,semantic}_connector.*`

The connector / head `.bin` files are read by a new **pure-C++** Torch-pickle reader (`torch_bin.*`) — no Python at inference or preprocessing. It's parity-checked bit-for-bit against `torch.load` (`torch_bin_parity`), and an overlay check (`vibevoice_finetune_overlay_check`) confirms the merged tensors equal the adapter's and that the merge reaches the forward pass (verified: `set_backend_tensor → require_tensor → require_tensor_data`, which the overlay overrides).

Docs and README updated to describe the full fine-tune behavior.

0xShug0:
Thanks! I will do a final test tomorrow and then merge it.

0xShug0:
@justinjohn0306 Code merged! Thanks a lot!

0xShug0:
@justinjohn0306 I did some changes to improve the lora performance. Just pushed acde132974d4e93c86558599ba74626296ced26c.

Metric | Before | After | Improvement
-- | -- | -- | --
LoRA decoder merge | 25,922 ms | 11,693 ms | 54.9% faster, 2.22x
Decoder weights load | 39,001 ms | 14,463 ms | 62.9% faster, 2.70x
Output WAV | baseline | byte-identical | no change

PR #25, opened 2026-07-03
Feature/server multipart transcriptions

dkruyt:
## Summary

Two small additions to `audiocpp_server` that make it a better drop-in for OpenAI-API-compatible clients (Open WebUI, llama-swap's playground, etc.):

- **Accept multipart/form-data uploads on `POST /v1/audio/transcriptions`.** The real OpenAI Whisper API (and every client built against it — Open WebUI included) uploads audio as a `multipart/form-data` `file` part. `audiocpp_server` currently only accepts a JSON body with a server-local path (`audio`/`audio_path`/`file` as a string). That's fine for local scripting, but it means the endpoint can't be used as-is behind a client that does real file uploads. This adds a small multipart parser (`app/server/multipart.{h,cpp}`) and routes requests with a `multipart/form-data` `Content-Type` through it: the uploaded bytes are spooled to a temp file and handed to the existing JSON-based request builder, so the original JSON/path request shape keeps working unchanged.

- **Add `GET /v1/audio/voices?model=<id>`.** Clients that build a voice picker (llama-swap's playground does this) have no way to discover which `cached_voice_id` values a TTS model actually supports, so they fall back to generic placeholder names (`alloy`, `nova`, ...) that don't exist as real presets and 500 at inference time. For families that keep voice presets under `model_root/embeddings/*.safetensors` (`pocket_tts` today — see `assets.cpp`), this lists the available ids. Families without an embeddings directory, or an unknown/missing `model` query param, just report an empty list, matching how the frontend already falls back gracefully. This also required capturing the raw query string on `HttpRequest`, which `read_http_request` was previously discarding entirely.

Neither change touches the existing JSON-only request shapes — both are purely additive.

## Testing

- New unit test exercising the multipart parser directly: boundary extraction (unquoted/quoted), multi-part splitting, and byte-exact handling of binary payloads with embedded NUL bytes.
- Built `audiocpp_server` with CUDA (`-DENGINE_ENABLE_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=native -DGGML_NATIVE=ON`) and ran it against real weights (`Qwen3-ASR-0.6B`, `kyutai/pocket-tts`), on an actual GPU:
  - `POST /v1/audio/transcriptions` with a JSON `{"model","audio"}` body (existing behavior) — unchanged, verified no regression.
  - `POST /v1/audio/transcriptions` with a real `multipart/form-data` upload (`-F file=@... -F model=...`, with and without `language`) — new behavior, correct transcript returned, temp file cleaned up afterward.
  - Missing `file`/`model` multipart fields — errors out same as other malformed requests (500, consistent with existing error handling style).
  - `GET /v1/audio/voices?model=pocket-tts` — returns the real cached voice ids (`alba`, `cosette`, `marius`, ...).
  - `GET /v1/audio/voices?model=qwen3-asr` (no embeddings dir), `?model=<unknown>`, and no `model` param — all return `{"voices":[]}`.
  - `POST /v1/audio/speech` — unchanged, verified no regression after both changes.
- Deployed behind [llama-swap](https://github.com/mostlygeek/llama-swap) in front of Open WebUI and confirmed both endpoints work end-to-end through a real client: STT via multipart upload, and TTS with the voice picker now populated from `/v1/audio/voices` instead of guessing.

## Notes for reviewers

- `handle_transcription` now dispatches on `Content-Type` (`multipart/form-data` → new path, otherwise → existing JSON path) rather than replacing the JSON handling.
- The multipart parser is intentionally minimal (no external dependency) and mirrors the existing hand-rolled HTTP parsing style already used in `http.cpp`.
- Voice discovery is scoped to "does `model_root/embeddings` exist" rather than hardcoding `pocket_tts`, so it should pick up other families automatically if they adopt the same layout.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -610,6 +610,17 @@ if (ENGINE_BUILD_TESTS)
         NAME encoder_module_test
         COMMAND encoder_module_test
     )
+
+    add_executable(server_multipart_test
+        tests/unittests/test_server_multipart.cpp
+        app/server/multipart.cpp
+    )
+    target_include_directories(server_multipart_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/app/server)
+
+    add_test(
+        NAME server_multipart_test
+        COMMAND server_multipart_test
+    )
 endif()
 
 if (ENGINE_BUILD_EXAMPLES)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -440,6 +440,7 @@ add_executable(audiocpp_server
     app/server/main.cpp
     app/server/config.cpp
     app/server/http.cpp
+    app/server/multipart.cpp
     app/server/runtime.cpp
     app/cli/args.cpp
     app/cli/request.cpp
--- a/app/server/README.md
+++ b/app/server/README.md
@@ -111,6 +111,29 @@ curl http://127.0.0.1:8080/v1/audio/transcriptions \
   }'
 ```
 
+Also accepts a `multipart/form-data` upload, matching the OpenAI Whisper API convention used by real clients (e.g. Open WebUI). The request is routed to the multipart path based on the `Content-Type` header; the JSON path above still works unchanged.
+
+```bash
+curl http://127.0.0.1:8080/v1/audio/transcriptions \
+  -F model=qwen3-asr \
+  -F language=en \
+  -F file=@/path/to/input.wav
+```
+
+`file` and `model` are required; `language` is optional. The uploaded bytes are spooled to a temporary file for the duration of the request and removed afterward.
+
+### `GET /v1/audio/voices?model=<id>`
+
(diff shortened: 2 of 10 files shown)


0xShug0:
@dkruyt Thank you for your contribution! I’ll test the PR, and if everything looks good, I’ll merge it.

0xShug0:
@dkruyt Two small follow-ups before merge:
1. Could you update `app/server/README.md` to document multipart transcription uploads and `GET /v1/audio/voices?model=<id>`?
2. The PR description mentions a new multipart parser unit test, but I do not see a test file or CTest target in the diff. Could you add it?

dkruyt:
Both addressed:

- `app/server/README.md` now documents the multipart `/v1/audio/transcriptions` upload (with a `curl -F` example) alongside the existing JSON path, and adds a section for `GET /v1/audio/voices?model=<id>` with an example request/response.
- Added `tests/unittests/test_server_multipart.cpp` plus a `server_multipart_test` CTest target (wired into `CMakeLists.txt` next to the other unit tests).

0xShug0:
@dkruyt merged. Thanks!

PR #45, opened 2026-07-12
Implement portable GGUF core support

mirek190:
## What changed

- adds a reusable `TensorSource` GGUF reader and `audiocpp_gguf` streaming converter
- supports direct conversion from safetensors and sharded `model.safetensors.index.json` sources
- preserves logical tensor names and exact original ranks, including trailing singleton dimensions
- embeds recursive, binary-safe model sidecars and materializes them into a fingerprinted temporary model directory when loading standalone GGUF files
- exposes `prepare_model_directory` so model families share the standalone-GGUF preparation path
- integrates standalone GGUF loading with Qwen3 ASR, Qwen3 Forced Aligner, Qwen3 TTS, Nemotron 3.5 ASR, VibeVoice-ASR, and Higgs Audio v3 STT
- adds `--no-sidecars` for tensor-only containers and documents conversion/compatibility behavior
- forwards server transcription language and context through the standard runtime transcript input

## How GGUF loading works

The converter reads tensors through the common `TensorSource` abstraction, streams and quantizes eligible tensors into GGUF, and records audio.cpp logical tensor names plus exact-rank metadata. By default it recursively packages non-weight model assets up to the configured safety limit into binary GGUF metadata. At load time, `prepare_model_directory` detects an embedded standalone GGUF, extracts those assets into a cache directory keyed by the GGUF path, size, and modification time, and returns both that prepared model root and the original GGUF tensor source. Older tensor-only GGUF layouts remain supported with external sidecars.

## Server request path

The OpenAI-compatible transcription endpoint receives `language` and optional context in the request's `text` field. Previously, the server copied only `language` into the generic options map and did not create the runtime `Transcript` input used by model sessions. The server now constructs `Transcript{text, language}` whenever either field is present and assigns it to `TaskRequest::text_input`. It also keeps `options[language]` populated for backward compatibility. As a result, CLI and server execution deliver the same language hint and context prompt to the ASR model runtime, including when the model weights come from GGUF.

## Validation

- built `gguf_tensor_source_test`, `audiocpp_gguf`, and `audiocpp_server` with the Windows CPU configuration
- ran `gguf_tensor_source_test` successfully
- manually tested Q8 GGUF models in the ASR application

GGUFs for test
https://huggingface.co/mirek190/audio.cpp/tree/main

example to use ASR
```
  .\audiocpp_cli.exe `
    --task asr `
    --family qwen3_asr `
    --model .\models\Qwen3-ASR-0.6B-Q8_0.gguf`
    --backend cuda `
    --audio .\audio\example_1.wav`
    --text-out .\outputs\qwen06_transcript.txt `
```
example to use ASR with a text timecodes
```
  .\audiocpp_cli.exe `
    --task asr `
    --family qwen3_asr `
    --model .\models\Qwen3-ASR-0.6B-Q8_0.gguf`
    --backend cuda `
    --audio .\audio\example_1.wav `
    --text-out .\outputs\qwen06_transcript.txt `
    --words-out .\outputs\qwen06_words.json `
    --session-option "qwen3_asr.forced_aligner_model_path=.\models\Qwen3-ForcedAligner-0.6B-Q8_0.gguf"    
```


mirek190:
First path

From my tests Q8 GGUFs are 50% faster than ASR safetensor models with my RTX 3090

0xShug0:
@mirek190 Would you like to take a look at the macOS error?

mirek190:
> @mirek190 Would you like to take a look at the macOS error?

Already fixed

The problem was with timestamp code not explicitly converted to long long for Xcode’s libc++.

0xShug0:
> > @mirek190 Would you like to take a look at the macOS error?
> 
> Already fixed

Thanks! I will run some end to end tests tomorrow.

0xShug0:
@mirek190 I tried the conversion tool first, and it generally works well. I have two comments:

1. **Rank-0 scalar tensors**

When converting IndexTTS2 from https://huggingface.co/mlx-community/index-tts2-mlx, the conversion failed on `campplus.safetensors`:

```text
error: GGUF supports tensor ranks from 1 to 4: speaker_encoder.head.bn1.num_batches_tracked
```

This tensor is a rank-0 scalar in the source safetensors file. The current converter in the PR rejects empty-shape tensors before writing them to GGUF.

This does not appear to be specific to IndexTTS2. Safetensors checkpoints can contain scalar tensors, such as BatchNorm tracking counters. The PR already exposes `TensorSource::require_i64_scalar()`, and the GGUF tensor source supports reading scalars, but the converter rejects rank-0 tensors before they can be represented.

2. **Packing multiple components into a single GGUF**

One follow-up that would make GGUF much more convenient for multi-component models such as IndexTTS2 would be support for packing multiple tensor sources, along with required sidecar files, into a single GGUF.

The current converter maps one safetensors source to one GGUF, which works well for models with a single weight file. However, models such as VeVo2, VoxCPM, and IndexTTS2 contain multiple components, including GPT, S2Mel, a semantic codec, CAMPPlus, BigVGAN, Qwen emotion, W2V-BERT, and small matrix or statistics files.

Another example is higgs-audio-v3-stt. After convert I can't run the GGUF from a random path directly, because it expects whisper-large-v3/preprocessor_config.json. 

Converting each component into a separate GGUF works mechanically, but from a user’s perspective, it still behaves like a directory containing many weight files.

A single packed GGUF with namespaced tensor groups, such as `gpt/`, `s2mel/`, `bigvgan/`, and `qwen_emotion/`, would preserve clear internal component boundaries while still providing the simple one-file model experience that GGUF is good at.

Thanks again for working on GGUF support. I’m really excited about it. I plan to refactor the code to further decouple asset loading from the model impl (after merging your PR), and future models will be GGUF-first.

0xShug0:
@mirek190 I’ll merge your PR into `dev` first, maybe in about 2 hours, because I really need to use your current PR as the base for a refactor. I’d like to include these new changes in tomorrow’s model release. If you’re interested in further improvements, feel free to submit follow-up PRs targeting the main branch (tag will be release 0.3).

mirek190:
Pushed the packed-GGUF and rank-0 tensor follow-up in commit 4b68fd4. For Higgs Audio v3 STT, whisper-large-v3/preprocessor_config.json is now needed only as an input while creating the GGUF via --sidecar. Once embedded, the completed GGUF can be renamed or moved and loaded directly without the external Whisper file or directory. Existing Higgs GGUFs created without that embedded sidecar still require the old layout or reconversion.... that why I updated the GGUF for higgs-stt on my huggingface repo.

For creation it looks like that now 
```
.\audiocpp_gguf.exe `
    --input models\higgs-audio-v3-stt\model.safetensors.index.json `
    --root models\higgs-audio-v3-stt `
    --sidecar models\whisper-large-v3\preprocessor_config.json=preprocessor_config.json `
    --output models\higgs-audio-v3-stt-Q8_0\model.gguf `
    --type q8_0
```

0xShug0:
@mirek190 Thank you! Merged into dev and I will highlight this feature in the 0.3 release.

mirek190:
The best is that Q8 has nearly identical quality as fp16 but are 2x faster - at least for ASR / STT

0xShug0:
According to this independent benchmark https://github.com/Martossien/transcria/blob/main/docs/STT_BENCHMARK_REAL_MEETINGS.md , our implementation gets the same WER in about 1/4 of the time. Can’t imagine another 2x speedup on top of that!

mirek190:
> According to this independent benchmark https://github.com/Martossien/transcria/blob/main/docs/STT_BENCHMARK_REAL_MEETINGS.md , our implementation gets the same WER in about 1/4 of the time. Can’t imagine another 2x speedup on top of that!

Sonic the Hedgehog speed ;)

PR #46, opened 2026-07-14
Add standalone GGUF support for Hviske v5.3 and Citrinet-256

mirek190:
## What changed

- sync the change onto `main` at `66d233b`
- add `model_specs/hviske_asr.json` and `model_specs/citrinet_asr.json`
- describe native safetensors and standalone GGUF layouts, including named configuration, tokenizer, vocabulary, and tensor resources, in those specs
- refactor both model loaders to consume the framework-resolved named resources instead of selecting model filenames in C++
- accept direct, renamed GGUF files while retaining native safetensors support
- decode quantized source matrices to F32 when a model reshapes them into an incompatible Conv1D layout
- document Q8_0 conversion commands and update the package-spec/GGUF test matrix in `docs/gguf.md`
- add regression coverage for native/GGUF package resolution and the quantized matrix-to-Conv1D reshape path

## Why

The shared GGUF core could package these checkpoints, but these families did not have package specs. Citrinet Q8_0 also exposed a GGML row-layout mismatch when a quantized matrix was reshaped into a convolution kernel whose final dimension was not block-sized.

This change lets a completed `model.gguf` be moved, renamed, and passed directly to the CLI or server without the original model directory.

## Validation

- focused CPU targets (`asr_standalone_gguf_test`, CLI, and server) build successfully on `66d233b`
- `asr_standalone_gguf_test` passes through the package-spec resolver
- CUDA CLI and server load both real Q8_0 files successfully through their package specs
- CUDA CLI loads both native safetensors layouts successfully through their package specs
- Hviske native and Q8_0 produced the same transcript on the test WAV
- Citrinet native and Q8_0 produced the same transcript on the test WAV

The complete default CPU build on `66d233b` is currently blocked by the existing MOSS codec parity targets, which still pass filesystem paths to constructors that now require `TensorSource` objects. Those unrelated targets are not changed by this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -822,6 +822,17 @@ if (ENGINE_BUILD_TESTS)
         COMMAND scaled_dot_product_attention_test
     )
 
+    add_engine_unittest(asr_standalone_gguf_test tests/unittests/test_asr_standalone_gguf.cpp)
+    target_include_directories(asr_standalone_gguf_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/tests/unittests)
+    target_compile_definitions(asr_standalone_gguf_test PRIVATE
+        ENGINE_TEST_ASSET_ROOT="${ENGINE_UNITTEST_ASSET_ROOT}"
+    )
+
+    add_test(
+        NAME asr_standalone_gguf_test
+        COMMAND asr_standalone_gguf_test
+    )
+
     add_executable(server_multipart_test
         tests/unittests/test_server_multipart.cpp
         app/server/multipart.cpp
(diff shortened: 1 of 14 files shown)


0xShug0:
@mirek190 you moved fast :+1:  I pushed the package-spec refactor, along with the other pending changes, to main. Could you please sync with the latest main (66d233bbed1a60d7138704b704d25cbaf8e3c003), check the current package-spec loading flow, and update your code against that version? 

Package-spec loading means model file layout is described in model_specs/*.json, not hardcoded in each model.
The framework resolves configs, tokenizers, sidecars, safetensors, or GGUF sources from that spec, then the model code just consumes named resources. 

There is a new doc to track Package-spec refactor and GGUF testing https://github.com/0xShug0/audio.cpp/blob/main/docs/gguf.md. Please also update this table in the PR. 

Thanks again for your contribution!

mirek190:
Sure

All conflicts resolved and updated.

- Refactored both loaders to consume named package resources
- Removed hardcoded model-layout selection from C++
- Updated the docs/gguf.md testing table

Synced exactly with the main monster commit 66d233b

0xShug0:
@mirek190 Merged! Thanks a lot!

PR #30, opened 2026-07-05
Add Dockerfiles and Docker examples

xashr:
### Summary

Docker support for audio.cpp is currently limited and focused on development (see `Docker.md`).

This PR adds Docker support to help users get started quickly. It provides:

- **Multi-stage Dockerfiles** (CPU and CUDA) with separate build and runtime stages
  - Inspired by llama.cpp (hence the `.devops` folder naming)
  - "Full" images with all binaries and the multiplexer (similar to llama.cpp's full images)
- **Ready-to-run examples for Pocket TTS (CPU & CUDA)**
  - **CLI:** run Pocket TTS via `docker run`
  - **Server:** run Pocket TTS via Docker Compose

Tested on Ubuntu Linux with an RTX 5090 and an Intel CPU.

### Quick Start

**CLI:**
1. Download Pocket TTS models into the `examples/docker/models/` folder
2. Run from repo root: `docker build -f .devops/cuda.Dockerfile -t local/audiocpp:full-cuda .`
3. Run `./cuda-tts.sh` to generate speech

**Server:**
1. Download Pocket TTS models into the `examples/docker/models/` folder
2. Run `docker compose -f cuda-server.yml up`
3. Run `./tts.sh` to generate speech via a curl request

### Known Issues

- ~~The audio.cpp server does not terminate gracefully. The container cannot be stopped cleanly and must be killed (e.g. press `Ctrl+C` twice with Docker Compose). Proper signal handling is needed in the server.~~

### Todo

Some open tasks before merging:

- ~~Extend main documentation regarding Docker build and Docker examples~~
- Decide on naming convention for docker images (e.g. audiocpp vs audio.cpp)
- (Support Vulkan)

PR #51, opened 2026-07-14
Add Docker GitHub Workflow

xashr:
## Summary

#30 added Dockerfiles for CPU and CUDA including examples. This PR adds a full GitHub workflow for building, caching and publishing multi-architecture Docker images to GHCR. Inspired by llama.cpp, adapted to audio.cpp.

This allows the user to simply reference a public docker image without having to build it locally (which is still supported).

Example (if merged): `docker pull ghcr.io/0xShug0/audio.cpp:full-cpu`

See the [fork workflow](https://github.com/xashr/audio.cpp/actions/runs/29353985240)  and [fork image tags](https://github.com/xashr/audio.cpp/pkgs/container/audio.cpp/versions?filters%5Bversion_type%5D=tagged) in action.

## Workflow

### Trigger

Workflow gets triggered daily at 3:21 UTC or manually. The workflow checks for new commits. If there is no new commit since the last build, the workflow jobs get cancelled to avoid wasting resources.

Can be adjusted as needed, e.g. build weekly to reduce actions usage.

### Stages

 1. **`check-commits`** — Compares HEAD to `last-docker-build` tag. Skips the entire run if no new commits since the last daily build. `workflow_dispatch` with `force_build=true` overrides the skip.                                        
                                                                                                                                                                                                                                              
 2. **`metadata`** — Computes `build_date` and `date_tag` once so all jobs use the same timestamp.                                                                                                                                            
                                                                                                                                                                                                                                              
 3. **`build`** — Runs in parallel across architectures (`amd64`, `arm64`) and variants (`cpu`, `cuda12`, `cuda13`). Each job:                                                                                                                
    - Builds a single-arch image targeting the `full` stage                                                                                                                                                                                   
    - Pushes by digest to `ghcr.io`                                                                                                                                                                                                           
    - Uploads the digest as an artifact for the merge stage                                                                                                                                                                                   
                                                                                                                                                                                                                                              
 4. **`merge`** — Downloads per-arch digests and uses `docker buildx imagetools create` to assemble multi-arch manifests with OCI annotations. Produces:                                                                                      
    - `full-cpu` — mutable "latest" tag                                                                                                                                                                                                       
    - `full-cpu-YYYYMMDD-HHHHHHH` — immutable pinned tag                                                                                                                                                                                      
    - `full-cuda12` — mutable "latest" tag                                                                                                                                                                                                       
    - `full-cuda12-YYYYMMDD-HHHHHHH` — immutable pinned tag                                                                                                                                                                                      
    - `full-cuda13` — mutable "latest" tag                                                                                                                                                                                                       
    - `full-cuda13-YYYYMMDD-HHHHHHH` — immutable pinned tag                                                                                                                                                                                      
                                                                                                                                                                                                                                              
 5. **`update-tag`** — Force-pushes the `last-docker-build` git tag to HEAD, enabling the skip-detection in step 1.   

## Tested variants

**Tested:**
amd64: cpu, cuda12, cuda13 on Intel CPU + RTX 5090
arm64: cpu on Macbook Pro M5

**Not tested (lacking hardware):**
arm64: cuda12, cuda13

## What is not included in this PR

This PR provides the workflow. If the PR gets accepted and the workflow works reliably, I can extend the documentation and examples.

## Open issues
- `ENGINE_ENABLE_NATIVE_CPU=OFF`: For local Docker builds this can be set to ON for improved performance. For published Docker builds it needs to be OFF to ensure portability, resulting in reduced performance. llama.cpp offers the option to build with `GGML_CPU_ALL_VARIANTS=ON` which builds multiple separate CPU backend shared libraries, each optimized for a different CPU generation and loaded dynamically. It would be great to have that option for audio.cpp as well. Otherwise I could try to enable native cpu for local and disable it for published images for now.
=> **Will create separate issue. For now: OFF for published builds, ON for local builds**
- ~~arm64 build warning~~
=> **Removed "free disk space" step, currently not needed**


0xShug0:
@xashr Thanks! One runtime concern:  The image currently copies binaries only. Current audio.cpp needs package specs unless the build uses embedded specs, so the Docker build should either enable `AUDIOCPP_DEPLOYMENT_BUILD=ON` or include `model_specs/` in the runtime image.

Comments from ChatGPT (not sure if they are relevent):
Add permissions: contents: read
move the commet outside the build-args block: `ENGINE_ENABLE_NATIVE_CPU=OFF # disable for cross-arch portability`

xashr:
@0xShug0 : Valuable feedback! Look's like I was testing with the only model that does not require a model spec (pocket tts)... Fixed that. I will add another example which requires a model with spec in the follow up PR (docs).

Applied the other two suggestions as well. (permissions setting not needed if repo default, but makes it more explicit)

0xShug0:
@xashr Thank you for this great improvement! Merged.

Issue #55, opened 2026-07-15
pocket_tts: failed CUDA alloc in MimiDecoder::decode aborts the whole server (GGML_ASSERT) — same class as the voice-prep issue fixed in 1627694


5uck1ess:
Follow-up to #50 (thanks for the fast fix in 1627694 — verified working for voice prep). The same abort-on-OOM pattern still exists in the **synthesis decode path**.

## What happened

Build: upstream main @ b2e0bb5, Linux + CUDA (RTX 3090). A long TTS request arrived while other processes held most of the card (~2.5 GiB free). The Mimi decode graph tried to allocate ~3.9 GiB, the cudaMalloc failed, and instead of failing the request the process aborted:

```
ggml_backend_cuda_buffer_type_alloc_buffer: allocating 3889.51 MiB on device 0: cudaMalloc failed: out of memory
alloc_tensor_range: failed to allocate CUDA0 buffer of size 4078444288
external/ggml/src/ggml-backend.cpp:327: GGML_ASSERT(buf != NULL && "tensor buffer not set") failed
```

Symbolized (addr2line, debug info in the release build):

```
engine::models::pocket_tts::MimiDecoder::decode(...)   <- allocation happens here
minitts::server::ServerState::run_model(...)           app/server/runtime.cpp:788
```

## Why it matters

Same operational problem as #50: one over-budget request (here: a long utterance under VRAM pressure) kills every loaded model and any concurrent session, instead of returning an error for that request. A supervisor can revive the process, but in-flight work is lost.

## Suggestion

The treatment from 1627694 seems to apply directly: allocate the decode graph through `ggml_gallocr` reserve/alloc (or otherwise check the buffer) and throw a `runtime_error` on failure so the server maps it to a failed request rather than marching into the GGML_ASSERT.

Happy to test a fix on the same 3090 setup — the condition is easy to reproduce here by loading the card and requesting a long synthesis.

5uck1ess:
Happy to send a PR for this mirroring the 1627694 approach (gallocr reserve/alloc + runtime_error on failure for the decode graph) — I have the 3090 repro rig to verify under real VRAM pressure. Say the word if you'd rather take it yourself, either works.

0xShug0:
@5uck1ess Can you try this branch https://github.com/0xShug0/audio.cpp/tree/fix/pocket-tts-mimi-decoder-oom. If it looks good I will merge the fix into main.  

[...]

PR #59, opened 2026-07-15
Harden PocketTTS FlowLM step allocation against OOM (follow-up to #55 decoder fix)

5uck1ess:
**Draft** — follow-up to the Mimi decoder OOM hardening on `fix/pocket-tts-mimi-decoder-oom` (this PR is based on that branch; retarget to `main` once it merges).

While verifying that branch on a 3090 under real VRAM pressure (#55), I found the SIGABRT still reproduces at tighter memory — just in a sibling `pocket_tts` runtime the branch doesn't touch: `FlowLMStepRuntime`, reached via `PocketTTSSession::prepare`. Its `ggml_backend_alloc_ctx_tensors` (and two `ggml_gallocr_alloc_graph` calls) are unguarded, so a failed CUDA buffer alloc returns null and the following `write_tensor`/graph use trips `GGML_ASSERT(buf != NULL)` at ggml-backend.cpp:327.

This applies the same treatment you used in the decoder to all three FlowLM allocation sites:
- **params buffer** (`alloc_ctx_tensors`) — null-check → `runtime_error`
- **prompt graph** — `ggml_gallocr_reserve` + `alloc_graph`-or-throw
- **lazy step graph** (`ensure_step_graph_allocated`) — same, freeing the partial gallocr so a retry doesn't leak

Because a throw from the constructor skips the destructor, the destructor cleanup is factored into an idempotent `release_runtime()` that the failure paths call, so the ggml context and any partial buffers are freed rather than leaked.

**Verified on an RTX 3090** (A/B, squeezing free VRAM with a filler then synthesizing a cloned voice):
- Original bug: `main` SIGABRTs on the ~3 GB Mimi reserve at ~1.1 GB free.
- Your branch: fixes that, and lowers peak VRAM (same synth that aborted `main` at 1143 MiB free succeeds at 997 MiB free) — but still SIGABRTs in FlowLM at ~250 MiB free.
- With this change: at **227 MiB free** the same synthesis returns a clean `500` (`FlowLM step parameter buffer allocation failed (out of memory)`) and the server stays responsive — no abort.

Draft because you may prefer to fold it into the decoder branch or adjust the messages/idiom. Happy to adapt.

### DIFF
--- a/src/models/pocket_tts/flow_lm.cpp
+++ b/src/models/pocket_tts/flow_lm.cpp
@@ -381,6 +381,13 @@ public:
         combined_scratch_.resize(static_cast<size_t>(config_.latent_size + 1));
 
         params_buffer_ = ggml_backend_alloc_ctx_tensors(ggml_ctx_, backend_);
+        if (params_buffer_ == nullptr) {
+            // Under VRAM pressure the CUDA buffer alloc fails and returns null;
+            // the write_tensor calls below would then trip GGML_ASSERT and
+            // SIGABRT the whole server. Fail the request instead (see #55).
+            release_runtime();
+            throw std::runtime_error("FlowLM step parameter buffer allocation failed (out of memory)");
+        }
         if (prompt_steps_ > 0) {
             core::write_tensor_f32(prompt_embeddings_, prompt_embeddings_buffer_);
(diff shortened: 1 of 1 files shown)


0xShug0:
Thank you @5uck1ess! Merged!

PR #61, opened 2026-07-15
Remove temporary file writes for multipart upload

Cr4xy:
Wanting to run audio.cpp as an ASR server, it used to hurt thinking about the disk getting hit every time. It no longer hurts with this patch :)

### DIFF
--- a/app/cli/request.cpp
+++ b/app/cli/request.cpp
@@ -53,6 +53,15 @@ engine::runtime::AudioBuffer read_audio_buffer(std::istream & input) {
     };
 }
 
+engine::runtime::AudioBuffer read_audio_buffer(std::string_view input) {
+    const auto wav = engine::audio::read_wav_f32(input);
+    return engine::runtime::AudioBuffer{
+        wav.sample_rate,
+        wav.channels,
+        wav.samples,
+    };
+}
+
 std::string json_option_string(const engine::io::json::Value & value) {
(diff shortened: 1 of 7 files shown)


0xShug0:
@Cr4xy Thank you for the PR!. One concern is in your PR the current path is `memory upload bytes -> JSON audio_data string -> istringstream -> AudioBuffer` which creates an extra memory copy. A better path may be `memory upload bytes -> non-owning string_view stream -> AudioBuffer`. I just did a quick test. 


Ver. | Peak RSS | RSS Rise | Wall
-- | -- | -- | --
main temp-file path | 941480 KB | 768992 KB | 1.788s
PR #61 original | 1221116 KB | 1048444 KB | 1.851s
Prototype direct buffer | 956500 KB | 783856 KB | 1.615s

The implementation in the attached patch is just a POC and is not clean, but it should give you the idea.

[pr61-direct-multipart-buffer.patch](https://github.com/user-attachments/files/30061302/pr61-direct-multipart-buffer.patch)

Cr4xy:
I think the prototype is good. Since I don't really know what's not clean about it, my best guess is that maybe its the  MemoryStreamBuffer that is a bit of boilerplate?
Please feel free to change whatever else you need in this PR.

0xShug0:
@Cr4xy Merged. Thank you!

PR #62, opened 2026-07-15
platform-neutral binary audiocpp_gguf moved to others binaries src under app/

mirek190:
## Summary

- move the platform-neutral `audiocpp_gguf` entry point from `tools/gguf_convert.cpp` to `app/gguf/main.cpp`
- update the CMake target to compile the converter from its new application location
- explicitly compile `audiocpp_gguf` in the Windows, Linux, and macOS build workflows

## Why

`audiocpp_gguf` is a distributed application alongside `audiocpp_cli` and `audiocpp_server`, rather than a development-only helper. Keeping its entry point under `app/` makes that ownership clear while leaving the shared GGUF reader, writer, packing, and quantization implementation in the framework layer.

## Impact

The executable name and command-line interface are unchanged. Windows still produces `audiocpp_gguf.exe`; Linux and macOS produce `audiocpp_gguf`. The source move is byte-for-byte identical and does not change converter behavior.

## Validation

- Windows CPU Release `audiocpp_gguf` build passed from a clean worktree
- existing standalone Citrinet GGUF inspected successfully with the rebuilt binary
- source relocation detected as a 100% Git rename
- no references to `tools/gguf_convert.cpp` remain
- `git diff --check` passed


0xShug0:
@mirek190 Merged, thank you!

PR #64, opened 2026-07-15
Return voices for single-model servers when model param is omitted

kevin-ho:
## Problem

When the server has a single model configured, `GET /v1/audio/voices` without a `?model=` query parameter returns `{"voices":[]}`. Clients that call this endpoint to populate a voice picker — without prior knowledge of model IDs — receive an empty list and cannot discover available voices.

## Fix

When no `?model=` param is provided and exactly one model is configured, fall back to that model's voices. When multiple models are configured, the existing behavior is unchanged (empty response, since voices cannot be attributed to a specific model without the param).

```cpp
size_t model_idx = SIZE_MAX;
if (!model_id.empty()) {
    const auto it = model_index_.find(model_id);
    if (it != model_index_.end()) {
        model_idx = it->second;
    }
} else if (models_.size() == 1) {
    model_idx = 0;
}
```

## Rationale

The change is purely additive — it only affects the `models_.size() == 1` path that previously returned empty. The multi-model path and the explicit `?model=` path are unchanged.

Voice presets are static config metadata parsed at startup, so this works regardless of `lazy_load` setting. No runtime state or model session is involved — `handle_voices` only reads the in-memory `voice_presets` map and the `embeddings/` directory listing.

## Tested

- `GET /v1/audio/voices` (no param) → voices returned (was `[]`)
- `GET /v1/audio/voices?model=<id>` → unchanged
- Backend: CUDA

### DIFF
--- a/app/server/README.md
+++ b/app/server/README.md
@@ -237,7 +237,7 @@ The stream emits `transcript.text.delta` events, one final `transcript.text.done
 
 ### `GET /v1/audio/voices?model=<id>`
 
-Lists the cached voice ids and configured server voice preset names available for a TTS model, so a client can populate a voice picker instead of guessing generic names. For families that keep voice presets under `model_root/embeddings/*.safetensors` (`pocket_tts` today), this returns those ids too; for other families with no configured presets, or an unknown/missing `model` parameter, it returns an empty list.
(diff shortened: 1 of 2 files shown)


0xShug0:
@kevin-ho Thank you! PR merged.

PR #72, opened 2026-07-17
build: require CUDA 12.0+ and document the Linux CUDA toolchain

patrickjchen:
A native Linux CUDA build had no documented toolkit version: docs/build/linux.md asked only for "a supported backend toolchain", CMake set no minimum, and Linux CI builds with ENGINE_ENABLE_CUDA=OFF, so nothing caught an unusable toolkit.

That combination fails badly on a stock Ubuntu 22.04 box. CMake takes the first nvcc on PATH, which is the distro nvidia-cuda-toolkit (CUDA 11.5); nvcc < 11.6 cannot parse libstdc++ >= 11.3 headers, so configure succeeds and the build dies at the first .cu file with "parameter packs not expanded with '...'" inside <functional> -- a toolchain fault that reads as a code fault.

find_package(CUDAToolkit 12.0 REQUIRED) turns that into a clear configure-time error. 12.0 is a floor, not a target: any 12.x or 13.x works (the Dockerfile pins 12.9.0 and a CUDA 13 package is documented for Windows). The upper bound is the host compiler the chosen toolkit accepts, not the CUDA version, so it is left to the toolkit.

The docs also cover the trap that CMAKE_CUDA_COMPILER alone does not close: it selects the compiler but not the libraries, so with a distro CUDA also installed the link silently picks up libcudart/libcublas 11 from /usr/lib/x86_64-linux-gnu while compiling with the newer nvcc. CUDAToolkit_ROOT is what pins the libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,7 +89,16 @@ set(AUDIOCPP_GGML_SOURCE_DIR
 
 if (ENGINE_ENABLE_CUDA)
     enable_language(CUDA)
-    find_package(CUDAToolkit REQUIRED)
+    # 12.0 is a floor, not a target: any 12.x or 13.x works (the Dockerfile pins
+    # 12.9.0, and docs/build/windows.md ships a CUDA 13 package). The minimum is
+    # enforced because an older toolkit configures happily and only fails at the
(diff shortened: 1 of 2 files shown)


0xShug0:
@patrickjchen Merged! Thank you!

PR #75, opened 2026-07-18
Fix/tts hang chatterbox

patrickjchen:
1) timed_mutex is used, so that one failed/stuck operation on GPU will not block other threads.
2) Allow busy timeout to be configured per model and per request

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -903,6 +903,21 @@ if (ENGINE_BUILD_TESTS)
         COMMAND server_multipart_test
     )
 
+    add_executable(server_busy_guard_test
+        tests/unittests/test_server_busy_guard.cpp
+    )
+    target_include_directories(server_busy_guard_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/app/server)
+    find_package(Threads REQUIRED)
+    target_link_libraries(server_busy_guard_test PRIVATE Threads::Threads)
+
+    add_test(
+        NAME server_busy_guard_test
+        COMMAND server_busy_guard_test
+    )
+    # The suite deliberately drives threads against a held lock; bound it so a
+    # regression that reintroduces unbounded queuing fails instead of hanging CI.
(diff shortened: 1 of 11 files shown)


0xShug0:
@patrickjchen Thanks for the PR! Merged.

I validated the main behavior locally and the direction looks useful.

For future PRs, could you also include a short validation summary in the description? A few concrete results really help review. That makes it much easier to review quickly and gives us more confidence that the change is safe across the server paths.

### Build and Unit Tests

- Built `audiocpp_server` in debug mode.
- Built and ran:
  - `server_busy_guard_test`: passed
  - `server_config_test`: passed

### Normal Server Request

- Started server with PocketTTS on CUDA.
- Sent one `/v1/audio/speech` request.
- Result: HTTP 200, valid WAV.

### Same-Model Busy Guard

- Started PocketTTS server with model `busy_timeout_ms=1`.
- Sent two overlapping requests to the same model.
- Result:
  - First request: HTTP 200, valid WAV
  - Second request: HTTP 503, `type=server_busy`

Confirms same-model fail-fast behavior works.

### Disabled Guard Behavior

- Started server with `busy_timeout_ms=0`.
- Sent two overlapping PocketTTS requests.
- Result: both requests returned HTTP 200.

Confirms old queueing behavior is preserved when the guard is disabled.

### Request-Level Override

- Started server with default guard disabled via `busy_timeout_ms=0`.
- Sent a second overlapping request with request body `"busy_timeout_ms": 1`.
- Result: second request returned HTTP 503, `type=server_busy`.

Confirms request-level timeout works.

### Cross-Model Concurrency

- Started one server hosting both Qwen3-TTS Base and PocketTTS.
- Started a long Qwen3-TTS request.
- While Qwen3-TTS was running, sent 5 short PocketTTS requests.
- Tested both current `main` and PR #75.
- Result: all PocketTTS requests returned HTTP 200 quickly in both versions.

Confirms PR #75 does not break existing cross-model concurrency.

patrickjchen:
Thanks @0xShug0, noted.

PR #77, opened 2026-07-19
Follow upstream PocketTTS default temperature (0.3 for English models)

adambenhassen:
## Motivation

Upstream pocket-tts changed the English models' default temperature from 0.7 to 0.3 in [kyutai-labs/pocket-tts@d108410](https://github.com/kyutai-labs/pocket-tts/commit/d108410d23eef7e01db282f9442891162dbc3db6) ("Human evaluations consistently prefer the English model at temperature 0.3 over the current default 0.7"). audio.cpp hardcodes 0.7 for all languages and ignores the new `default_temperature` yaml key, so English output at default settings no longer matches reference quality.

## Changes

- `PocketTTSModelConfig` gains `default_temperature` (fallback 0.7, matching upstream's `Config.default_temperature` schema default for custom yamls)
- Config yaml path reads upstream's `default_temperature` key when present
- Builtin defaults use 0.3 for `english` and `english_2026-04`, mirroring the two configs upstream changed; other languages stay at 0.7
- Session request builders seed the request from the model default before option parsing, so an explicit `temperature=` option still overrides
- Generation options now also accept `pocket_tts.`-prefixed spellings (following the existing `omnivoice.mem_saver`/`mem_saver` alias pattern): the warm bench passes namespaced keys the session never read, so its generation options (including parity seed and noise file) were silently ignored - masked until now because the harness default equaled the hardcoded 0.7

## Verification

```sh
scripts/build_metal.sh --target audiocpp_cli   # compiles clean
```

With `language=english` and no temperature option, generation runs at 0.3; `temperature=0.7` (either spelling) reproduces the previous output. Non-English languages unchanged.

### DIFF
--- a/include/engine/models/pocket_tts/assets.h
+++ b/include/engine/models/pocket_tts/assets.h
@@ -25,6 +25,7 @@ struct TransformerKVState;
 namespace engine::models::pocket_tts {
 
 struct PocketTTSModelConfig {
+    float default_temperature = 0.7F;
     int sample_rate = 24000;
     float frame_rate = 12.5F;
     float mimi_frame_rate = 12.5F;
--- a/src/models/pocket_tts/assets.cpp
+++ b/src/models/pocket_tts/assets.cpp
@@ -22,6 +22,7 @@ namespace {
 namespace binding = engine::modules::binding;
 
 struct PocketTTSDescriptor {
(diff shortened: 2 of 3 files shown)


0xShug0:
@adambenhassen Merged! Thank you!

PR #81, opened 2026-07-20
Fix Docker GitHub Workflow: Use lowercase image repository for registry references 

xashr:
## Summary

#51 added a workflow for docker builds. The build step currently fails due to uppercase letters in the registry name (0x**S**hug0).

## Fix

Replace the top-level IMAGE_REPO env var with a lowercase image_repo output from the metadata job. Docker/OCI registry references must be lowercase, which caused failures for repos with uppercase owner names (e.g. '0xShug0/audio.cpp').

- Remove env.IMAGE_REPO
- Add image_repo step to metadata job using tr '[:upper:]' '[:lower:]'
- Update build job outputs, cache-from, cache-to
- Update merge job repo variable

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -30,9 +30,6 @@ permissions:
   contents: read
   packages: write
 
-env:
-  IMAGE_REPO: ${{ github.repository }}
-
 jobs:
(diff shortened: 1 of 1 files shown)


0xShug0:
Thank you @xashr, merged!

PR #82, opened 2026-07-20
build: add cross-platform Nix flake and documentation

fedeizzo:
### Summary
This PR adds a `flake.nix` to provide deterministic, cross-platform builds via the Nix package manager, addressing the "cross-platform build and packaging polish" priority area in `CONTRIBUTING.md`.

It allows users to build and run `audio.cpp` entirely isolated from their host system dependencies, while natively supporting hardware acceleration.

### Changes Made
- **Added `flake.nix`:** Exposes dynamic packages for different backends (`vulkan`, `cuda`, `metal`, `cpu`), ensuring all CMake flags and dependencies (e.g., `cudatoolkit`, `darwin.apple_sdk`) are automatically handled.
- **Python Integration:** Bundles `tools/model_manager.py` directly into the build output (as `audiocpp_model_manager`) wrapped with its own isolated Python environment (containing `torch`, `safetensors`, and `pyyaml`).
- **Added `docs/nixos.md`:** Included user-facing documentation mimicking the style of `Docker.md`, covering usage, development shells, and how to consume this flake as an input in a NixOS configuration.

### Impact & Validation
- **Affects:** Only docs and build wiring. No changes were made to internal model routing, behavior, or runtime memory/performance.
- **Tested Backends:** Built and verified against Vulkan (Linux/Framework desktop) and CPU configurations.
- **Commands Checked:**
  - `nix build .#vulkan`
  - `nix shell .#vulkan -c audiocpp_model_manager install supertonic_3`
  - `nix develop` (verifying the dev environment assembles CMake/Ninja correctly)

### DIFF
--- a/docs/nixos.md
+++ b/docs/nixos.md
@@ -1,0 +1,111 @@
+# Running audio.cpp on NixOS (and Nix)
+
+## Prerequisites
+
+- The [Nix package manager](https://nixos.org/download) must be installed.
+- [Flakes](https://nixos.wiki/wiki/Flakes) must be enabled in your Nix configuration.
+
+## Packages
+
+The `flake.nix` provides multiple platform and hardware-specific builds. 
+The following backends are available:
+- **`vulkan`**: Hardware-accelerated build using Vulkan. (Default on Linux)
+- **`cuda`**: Hardware-accelerated build using NVIDIA CUDA.
+- **`metal`**: Hardware-accelerated build using Apple Metal. (Default on macOS)
+- **`cpu`**: Portable CPU-only build.
+
+All packages include the main tools: ***audiocpp_cli***, ***audiocpp_server***, ***audiocpp_gguf***, and the Python-based ***audiocpp_model_manager***.
+
+## Build the package
+
+### Default (Auto-detected OS)
+
+Build the default package for your operating system:
+
+```bash
+nix build .
+```
+
+### Specific Backends
+
+Build explicitly for a desired backend:
+
+```bash
+nix build .#vulkan
+nix build .#cuda
+nix build .#metal
+nix build .#cpu
+```
+
+## Usage
+
+After building, the compiled binaries will be available in the `result/bin/` directory.
+
+### Running the CLI
+
+You can execute the CLI directly via `nix run` (which uses `audiocpp_cli` as the main program):
+
+```bash
+nix run .#vulkan -- <arguments...>
+```
+
+### Running the Server
+
+To run the server, use `nix shell` or execute from the build result:
+
+```bash
+nix shell .#vulkan -c audiocpp_server <arguments...>
+# or
+./result/bin/audiocpp_server <arguments...>
+```
+
+### Downloading Models (Model Manager)
+
+The flake seamlessly bundles the `model_manager.py` Python script along with all its required dependencies (PyTorch, Safetensors, etc.) as `audiocpp_model_manager`.
+
+```bash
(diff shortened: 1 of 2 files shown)


docs/nixos.md
```diff
@@ -0,0 +1,111 @@
...
+You can easily include `audio.cpp` as a flake input in your own NixOS configuration or other Nix projects.
+
+In your `flake.nix`, add it to your `inputs`, making sure to use `follows` to avoid duplicating `nixpkgs` versions:
+
+```nix
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    
+    audiocpp.url = "github:fedeizzo/audio.cpp";
```
francescobozzo:
```suggestion
    audiocpp.url = "github:0xShug0/audio.cpp";
```

0xShug0:
@fedeizzo  Thank you! Nix support is a great addition. 

Is it possible to avoid hardcoding the package version as 0.3.0, since it may become outdated quickly?

fedeizzo:
> @fedeizzo Thank you! Nix support is a great addition.
> 
> Is it possible to avoid hardcoding the package version as 0.3.0, since it may become outdated quickly?

i just forced pushed a change where the version is the git commit revision, as far as i know this is the only way to have a dynamic version. The nix package is built in isolation so it cannot access the version from any other file. What do you think of this solution? Keep in mind that the nix version is used only to make an unique binary in the nix store, the version exposed by the binary will be the one you set in cpp

0xShug0:
@fedeizzo Merged! I'm not familiar with Nix, so I'll trust your judgment on the implementation :).

flake.nix
```diff
@@ -0,0 +1,120 @@
...
+              pname = "audio.cpp";
+              version = self.shortRev or self.dirtyShortRev or "dirty";
+
+              src = ./.;
+
+              nativeBuildInputs = with pkgs; [
+                cmake
+                ninja
+                pkg-config
+                makeWrapper
```
francescobozzo, line 41: Is this used?

flake.nix
```diff
@@ -0,0 +1,120 @@
...
+  };
+
+  outputs = { self, nixpkgs }:
+    let
+      supportedSystems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
+      forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
+      nixpkgsFor = forAllSystems (system: import nixpkgs { 
+        inherit system; 
+        config.allowUnfree = true;
+        config.cudaSupport = true;
```
francescobozzo, line 15: This should be applied to CUDA build only

docs/nixos.md
francescobozzo, line 89: It looks like this fix has been removed by a force push

flake.nix
```diff
@@ -0,0 +1,120 @@
...
+                ++ pkgs.lib.optional cudaSupport "-DENGINE_ENABLE_CUDA=ON"
+                ++ pkgs.lib.optional metalSupport "-DENGINE_ENABLE_METAL=ON";
+
+              installPhase = ''
+                runHook preInstall
+
+                mkdir -p $out/bin
+                
+                # Find and copy the built C++ executables
+                find . -type f -executable \( -name "audiocpp_cli" -o -name "audiocpp_server" -o -name "audiocpp_gguf" \) -exec cp {} $out/bin/ \;
```
francescobozzo, line 76: Is the find needed here? Can't we point directly to the output folder?

PR #85, opened 2026-07-21
Handle CORS requests

mosujiba:
This PR adds preliminary support for cross-origin HTTP requests from web applications to the API server.


0xShug0:
@mosujiba The security policy is too permissive and CORS is always enabled and cannot be disabled or restricted to trusted origins.

Safer direction: make CORS opt-in, configured with explicit allowed origins, only answer preflight for known endpoints/methods/headers, and only add CORS headers when the request Origin matches the allowlist.

mosujiba:
I wrote an opt-in implementation. Additional security features can be added later on.

0xShug0:
@mosujiba  Thanks! I will test it later.

0xShug0:
@mosujiba merged!

PR #74, opened 2026-07-18
Add machine-readable loader and package catalog exports

lapy:
## Summary

Integrators currently scrape `--list-loaders` text and infer package identity from `model_manager` metadata. This PR adds a small, versioned machine-readable discovery contract so tools can learn loaders and installable packages from the build itself.

### 1. `audiocpp_cli --list-loaders --json`

```json
{
  "schema_version": 1,
  "loaders": {
    "omnivoice": {
      "tasks": {"tts": ["offline"]},
      "api_endpoints": ["/v1/audio/speech"],
      "instructions_policy": "soft_tags"
    }
  }
}
```

- `IVoiceModelLoader::advertise*()` / `ModelRegistry::advertise_loaders()` provide a path-free catalog
- **Each loader overrides** `advertised_capabilities()` with the same task/mode contract it already uses in inspect/load (no central family-name heuristic table)
- Instruction policy defaults from advertised speech tasks (`openai_instruct` / `none`); loaders with a different contract override (`omnivoice`, `voxcpm2`, `irodori_tts`)
- API endpoints are derived from the advertised capability set
- Plain `--list-loaders` remains for humans, with task/mode summaries

### 2. Richer `model_manager.py list --json`

Each package payload includes:

- `family`, `tasks`, `modes`
- `standalone`, `parent_package_id`
- `gated`

`ModelPackage` accepts optional explicit overrides. Defaults come only from local package metadata:

- family: strip size/version suffixes from `id` (plus rare explicit overrides)
- standalone: `Subcomponent only.` / `Utility only.` / utility install kind / tokenizer|audiovae ids
- gated: known gated HF repo markers

### 3. Lazy Torch import for catalog commands

`list` / `info` no longer import Torch at module load. Install/convert paths import heavy deps on first use.

## Motivation

When a new loader or package is added, integrators should not need a parallel hardcoded catalog. The engine already knows its registered loaders and install packages — this surfaces that knowledge as a stable JSON contract.

## Test plan

- [x] `audiocpp_cli --list-loaders` prints a human-readable listing
- [x] `audiocpp_cli --list-loaders --json` returns `schema_version: 1` with non-empty tasks for all 30 registered families
- [x] Spot-check: omnivoice=`soft_tags`, voxcpm2=`text_prefix`, irodori=`caption_option`, silero_vad=`none`
- [x] `python tools/model_manager.py list --json` succeeds without Torch
[...]

PR #86, opened 2026-07-21
Keep package catalog in sync with registered loaders

lapy:
## Summary

After the machine-readable catalog contract (#74), integrators treat `model_manager list --json` `family` and `audiocpp_cli --list-loaders` as authoritative. Several packages were still advertised as installable even though their loaders are **not in this release tree**.

This PR aims to bridge the gap but is really intended as a **suggestion** to improve developer experience and software contracts. It would be nice to have a conversation about this.

### Mismatch examples (this tree)

| Family / stub | In `registry.cpp` | Loader sources under `src/` / `include/` | Catalog |
|---|---|---|---|
| `kokoro_tts` | commented stub only | **absent** (only `tests/kokoro_tts/*`) | was live HF snapshot → now `UnsupportedSource` |
| `higgs_tts` / catalog `higgs_audio_tts` | commented stub (`higgs_tts`) | **absent** | was live → now `UnsupportedSource` |
| `parakeet_tdt` | commented stub only | **absent** | was live → now `UnsupportedSource` |
| `vietneu_tts` | registered | present | package id inferred wrong → set `family="vietneu_tts"` |
| `silero_vad` | registered | present | no package by design (bundled) |

### Changes

- Mark the parked packages above as `UnsupportedSource`
- Set explicit `family` where id inference does not match the loader id
- Add maintainer docs + `tools/check_loader_catalog_sync.py` (registry ↔ catalog ↔ README; CI on Linux/macOS/Windows)
- Update README package table / CONTRIBUTING / usage docs

In-repo docs and the sync script stay generic; concrete mismatch cases are listed here only.

## Maintainer rule (suggestion)

**Never** leave a live `SnapshotSource` for a loader that is commented out of `registry.cpp`. Either ship the loader sources + register them, or mark the package `UnsupportedSource` and README **Unavailable**.

When adding a family: one family string everywhere (loader, `make_*_loader`, `model_specs/`, `ModelPackage.family`, README).

## Test plan

- [x] Confirmed no in-tree loader sources for the parked stubs above
- [x] `python3 tools/check_loader_catalog_sync.py --self-test`
- [x] `python3 tools/check_loader_catalog_sync.py`
- [x] Parked packages report `installable: false` via `package_payload`
- [ ] CI green on this PR

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -38,7 +38,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Check loader/catalog sync
-        run: python3 tools/check_loader_catalog_sync.py
+        run: |
+          python3 tools/check_loader_catalog_sync.py --self-test
+          python3 tools/check_loader_catalog_sync.py
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -37,6 +37,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Check loader/catalog sync
+        run: python3 tools/check_loader_catalog_sync.py
+
       - name: Install dependencies
         run: |
           sudo apt-get update
--- a/.github/workflows/mac-build.yml
+++ b/.github/workflows/mac-build.yml
@@ -27,7 +27,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Check loader/catalog sync
-        run: python3 tools/check_loader_catalog_sync.py
+        run: |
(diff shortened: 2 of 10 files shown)


0xShug0:
Thank you @lapy! This is a thoughtful improvement. Somehow, the current sync feels brittle, but it’s my fault because of the framework’s current design. In the long term, I do want to get rid of Python and use json to define downloadable model (and even model registration) similar to the current model spec system (https://github.com/0xShug0/audio.cpp/pull/53#issuecomment-4974295116). Further, there could be a single source of truth for model metadata, which would make sync easier.

As we move toward GGUF, model downloading and management can become much simpler, and the Python manager will eventually become a legacy tool. 

Happy to hear your thoughts on this direction!

PR #87, opened 2026-07-21
Add WebUi (Adopted from https://github.com/kigner/audio.cpp-webui, add support for Linux)

patrickjchen:
i18n:
    - Adopt webui/ui_i18n.py:English/Simplified/Traditional all now come from one source.
    - Default to English (was Chinese) and reorder the picker to match.
    - AUDIOCPP_LANG supplies the default only when nothing has been picked in the
      UI yet; a saved pick wins, so a stale env var can't defeat the picker.
      Accepts the spellings people actually write (zh_TW, zh-CN, english, ...).

Linux:
    - Re-apply EXE_SUFFIX / SERVER_EXE_NAME / SERVER_LAUNCHER and the from-source
      build discovery (_cmake_cache_backend, _discover_dev_bin_dirs), which also
      fixes _find_gguf_exe: upstream hardcoded build/windows-*-release, so it
      could never find a Linux converter build.
    - ffmpeg lookup and the "gguf converter not found" messages no longer name
      .exe unconditionally.
    - requirements.txt: mark pywin32 and the pythonnet/pywebview stack
      sys_platform == "win32" so the file installs on Linux.


0xShug0:
Thank you @patrickjchen. The changes to model manager in this PR may break existing CI/Nix flows. Could you tentatively create a dedicated model_manager.py for the UI so the existing one stays stays lightweight and focused on the framework installer/catalog? The manager for UI can be in the webui dir, which will make the UI self-contained.

patrickjchen:
> Thank you @patrickjchen. The changes to model manager in this PR may break existing CI/Nix flows. Could you tentatively create a dedicated model_manager.py for the UI so the existing one stays stays lightweight and focused on the framework installer/catalog? The manager for UI can be in the webui dir, which will make the UI self-contained.

done, a new file model_manager_webui.py is created, and the original one is restored

0xShug0:
@patrickjchen Merged! This is a big step forward for the project and thank you for the great work!

0xShug0:
@patrickjchen I played around with it, and the UI is pretty cool!

I organized it a bit to make the webui fully self-contained.

patrickjchen:
@0xShug0,

That's great!  Glad it's merged.

About the next project: have you thought about making it a MCP server? I
think it's necessary to have that functionality, and you are the best
person to define the schema. I can help implement the functionality.

Best regards,

Patrick

0xShug0:
@patrickjchen MCP sounds interesting, but I’m currently considering a redesign of the public interface to make UI development, external integrations, and community contributions easier. I'd prefer to stabilize that interface first so MCP can be implemented as a thin adapter instead of becoming coupled to the current framework internals.

BTW, would you like to add an English README for the WebUI? The current one is only available in Chinese.

0xShug0:
@patrickjchen Would you like to update the Web UI based on the release 0.4?

PR #90, opened 2026-07-22
Added English version README.md under webui

patrickjchen:
and the original Chinese version is renamed README.zh.md


0xShug0:
Merged. Thanks!

PR #84, opened 2026-07-20
Support all CPU variants with portable Docker builds

xashr:
## Summary

Introduces `ENGINE_ENABLE_CPU_ALL_VARIANTS`, a new CMake option for building **portable Docker images** that support all CPU instruction sets at runtime. The same binary runs optimally on any x86_64 or ARM64 CPU — no recompilation needed.

To make this work, the backend layer was modified to use only generic GGML APIs (`ggml-backend.h`), removing all compile-time `#ifdef` guards and backend-specific headers. This also aligns audio.cpp's architecture with llama.cpp's.

## The Problem

Before: producing a portable CPU (or GPU) Docker image required `-DENGINE_ENABLE_NATIVE_CPU=OFF`, which disabled all CPU optimizations (AVX2, AVX-512, etc.). The tradeoff was **portable or fast, not both**.

After: `-DENGINE_ENABLE_CPU_ALL_VARIANTS=ON` builds multiple `.so` variants (one per ISA) and selects the best at runtime. Portable **and** fast.

### 1. New CMake Option (`CMakeLists.txt`)

`ENGINE_ENABLE_CPU_ALL_VARIANTS` (default `OFF`, fully backward-compatible):
- Enables `BUILD_SHARED_LIBS` + `GGML_BACKEND_DL` + `GGML_CPU_ALL_VARIANTS`
- Disables `GGML_NATIVE`
- Sets RPATH to `$ORIGIN` so runtime finds `.so` files alongside the binary

### 2. Docker Build Updates

Both `.devops/cpu.Dockerfile` and `.devops/cuda.Dockerfile`:
- Pass `-DENGINE_ENABLE_CPU_ALL_VARIANTS=ON` (replaces the old `ENGINE_ENABLE_NATIVE_CPU` build arg)
- Collect `.so*` files into `/app/lib` and copy to the runtime stage
- CPU Dockerfile validates `TARGETARCH` (amd64/arm64 only)
- CUDA Dockerfile adds `-Wl,--allow-shlib-undefined` for dynamic backends

`.github/workflows/docker.yml` — removes the now-unnecessary `ENGINE_ENABLE_NATIVE_CPU=OFF` env var.

### 3. Backend API Modifications (`backend.cpp`, `backend.h`)

[...]

PR #102, opened 2026-07-25
[Backend selection] Fix metal backend

patrickvonplaten:
## Problem

`--backend metal` has never worked. Backend selection rejects the Metal device immediately
after ggml has brought it up successfully.

## Repro (before this PR)

```
audiocpp_cli --task asr --family voxtral_realtime \
  --model voxtral-mini-4b-realtime-2602-q8_0.gguf \
  --backend metal --audio speech.wav --text-out transcript.txt
```

```
ggml_metal_device_init: GPU name:   MTL0 (Apple M3)
ggml_metal_device_init: GPU family: MTLGPUFamilyApple9  (1009)
ggml_metal_device_init: has unified memory    = true
...
audiocpp_cli failed: Metal backend requested but no Metal device found
```

Any model and any task reproduces it — the failure is in `init_backend()`, before the model
is touched.

## Root cause

`find_device_by_backend_type()` identified a backend by matching both the ggml **registry
name** and a hardcoded **`ggml_backend_dev_type`**. Device type is not a stable per-backend
property, so that second condition silently excluded valid devices:

| `--backend` | Code expected | ggml reports | Result |
| --- | --- | --- | --- |
| `metal` | `ACCEL` | `GPU`, unconditionally (`ggml-metal.cpp`) | never worked |
| `vulkan` | `GPU` | `IGPU` on integrated GPUs (`ggml-vulkan.cpp`) | broken on APUs / iGPUs |
| `cuda` | reg name `"CUDA"` | `GGML_CUDA_NAME` is `"ROCm"` under HIP, `"MUSA"` under MUSA | broken on those builds |

`ACCEL` is what **BLAS** registers as, which is the giveaway.

The CUDA alias gap also reached `backend_type()`: `is_cuda_backend_handle()` compared against
`"CUDA"` only, so a ROCm/MUSA handle classified as `BackendType::BestAvailable`. That is
load-bearing — ~20 sites across `src/models/**`, `src/framework/modules/**` and
`release_backend_graph_resources()` branch on it, so those builds silently lost every CUDA
fast path and the CUDA graph cleanup.

## Fix

- Backend identity now comes from the owning ggml registry alone; the device-type filter is
  gone. Device lookup is `ggml_backend_reg_get` → `ggml_backend_reg_dev_get`.
- One alias table (`k_backend_reg_names`) is shared by the device lookup and the
  `is_*_backend_handle()` classifiers, so selection and classification cannot drift apart.
  CUDA accepts `CUDA` / `ROCm` / `MUSA`.
- The three near-identical CUDA/Vulkan/Metal branches in `init_backend()` collapse into one
  helper, which also adds a missing null check on `ggml_backend_dev_init()` — previously only
  the CPU branch checked, so the GPU branches could hand back a null backend.
- Errors now enumerate what ggml actually found:
  ```
  Metal backend requested but registry 'MTL' has no device 99
    (available: MTL:0 "MTL0" [GPU], BLAS:0 "BLAS" [ACCEL], CPU:0 "CPU" [CPU])
  ```

`query_backend_memory(const BackendConfig &)` shares the same lookup and is fixed with it.

**Behaviour note:** `--device <n>` now indexes devices within the selected backend's registry,
matching ggml's own numbering. Previously, on a Vulkan host with a discrete *and* an
integrated GPU, the iGPU was unreachable and indices silently skipped it.

## Test

`tests/unittests/test_backend_device_resolution.cpp` derives its expectations from ggml's
registry list rather than hardcoding a device class, so it passes on CPU-only CI and
exercises the real accelerator wherever one exists. It asserts that:

- every enumerated device of an accelerator registry is selectable by index;
- `backend_type()` round-trips the requested backend (the ROCm/MUSA guard);
- out-of-range and negative device indices are rejected;
- an unregistered backend never picks up a stray device from another registry.

## Verification

On Apple M3, `scripts/build_metal.sh --with-tests`:

- the repro command above now transcribes correctly on `--backend metal`; `cpu` and `best`
  produce the same transcript;
- `--device 99 --backend metal` fails with the enumerated message;
- `ctest`: 24/24 pass.

Left alone: four targets already broken on `main` and unrelated to backend selection
(`test_asr_standalone_gguf.cpp` and `tests/moss_tts_local/codec_*_parity.cpp` reference
renamed APIs and don't compile; `scaled_dot_product_attention_test` hardcodes CUDA with no
availability guard; `gguf_tensor_source_test` fails a model-spec message assertion). All
confirmed failing identically with this change stashed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -939,6 +939,14 @@ if (ENGINE_BUILD_TESTS)
         COMMAND torch_random_test
     )
 
+    add_engine_unittest(backend_device_resolution_test tests/unittests/test_backend_device_resolution.cpp)
+    target_include_directories(backend_device_resolution_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/tests/unittests)
+
+    add_test(
+        NAME backend_device_resolution_test
+        COMMAND backend_device_resolution_test
+    )
+
     add_engine_unittest(hf_sampler_test tests/unittests/test_hf_sampler.cpp)
 
     add_test(
--- a/src/framework/core/backend.cpp
+++ b/src/framework/core/backend.cpp
@@ -17,65 +17,158 @@ void ensure_backends_loaded() {
     }
 }
 
-static bool backend_has_reg_name(ggml_backend_t backend, const char * name) {
+// A backend is identified by the name of the ggml registry that owns it. The device type
+// (GPU/IGPU/ACCEL) deliberately plays no part in that: Metal reports GPU rather than ACCEL,
+// Vulkan reports IGPU on integrated GPUs, and those values are free to change upstream.
+//
+// The CUDA registry name depends on how ggml was built - GGML_CUDA_NAME is "ROCm" under HIP
+// and "MUSA" under MUSA - so every alias has to be accepted. The names are spelled out here
+// instead of pulled from ggml-cuda.h/ggml-vulkan.h because those headers resolve against the
+// backend's own build flags, which are not visible from this translation unit.
+struct BackendRegNames {
+    BackendType  type;
+    const char * names[3];  // unused entries are nullptr
+};
+
+constexpr BackendRegNames k_backend_reg_names[] = {
+    {BackendType::Cuda,   {"CUDA", "ROCm", "MUSA"}},
+    {BackendType::Vulkan, {"Vulkan", nullptr, nullptr}},
+    {BackendType::Metal,  {"MTL", nullptr, nullptr}},
+};
+
+bool reg_name_matches(BackendType type, const char * reg_name) {
+    if (reg_name == nullptr) {
+        return false;
+    }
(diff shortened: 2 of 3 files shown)


patrickvonplaten:
This PR should fix:

```
audiocpp_cli --task asr --family voxtral_realtime \
  --model voxtral-mini-4b-realtime-2602-q8_0.gguf \
  --backend metal --audio speech.wav --text-out transcript.txt
```

which now works.

patrickvonplaten:
Also replacing the command `--backend metal` with `--backend best` works but I believe that `--backend metal` should also work correctly. Curious to hear your thoughts.

0xShug0:
Thanks! I think the regression may be introduced by the recent PR https://github.com/0xShug0/audio.cpp/pull/84 merged just after release 0.4. I'm going to merge this PR as a hotfix and retest tomorrow.

PR #108, opened 2026-07-25
Release 0.4 webui

patrickjchen:
webui: resync the model catalog with release 0.4

    The WebUI catalog is a hand-maintained copy of facts that live in
    tools/model_manager.py, model_specs/ and registry.cpp, and 0.4 moved all
    three under it. Nothing failed loudly: a renamed install directory just
    makes an installed model read as "not installed", and a dropped package
    makes the Download button run a package id that no longer exists.

    - marblenet_vad is no longer an installable package; the weights ship in
      assets/framework/models/marblenet_vad. Point at them and drop the
      download_id, as silero_vad already does. Verified: VAD runs from the
      bundled path (2 segments on a 2.2s clip).
    - voxtral_realtime is now a standalone Q8_0 GGUF package installed into
      Voxtral-Mini-4B-Realtime-2602-GGUF. Follow the rename and drop the
      weight_type=q8_0 session option, which was quantizing an already-Q8
      package.
    - required_files.json: regenerated. kokoro/parakeet/marblenet are gone,
      voxtral and higgs_audio_v3_tts_4b are single .gguf files now, and
      fish_audio_s2_pro / vietneu_tts_v3_turbo / outetts_1_0_1b were missing
      entirely, so their installs were never integrity-checked.
    - Add the four families 0.4 registered that the UI never offered:
      higgs_audio_tts, fish_audio, outetts, vietneu_tts. VRAM estimates come
      from the Q8 peaks in docs/reports/gguf_q8_performance.md. Untested
      locally — the weights are not on this machine.
    - GGUF_NATIVE_FAMILIES: read model_specs/*.json instead of restating it.
      The literal listed 15 of the 33 families that now have specs, so the
      UI told users that e.g. Chatterbox "does not support native GGUF".
      The old literal stays as the fallback for bundles without model_specs/.

    test_catalog_sync.py pins all of it: package ids exist, install
    directories match, no-download_id entries are really bundled, the
    required-files map mirrors model_manager, and every registered family is
    reachable from the UI. Four of its six tests fail on the 0.4 tree.

### DIFF
--- a/webui/README.md
+++ b/webui/README.md
@@ -101,8 +101,10 @@ Starts the Gradio web interface (`webui.py`); open **http://127.0.0.1:7860** in 
   (one model in VRAM at a time; switching models restarts it).
 - The UI lets you upload a reference voice, download uninstalled models, enter an HF token / proxy, and so on.
 - Backend is auto-detected (as above: GPU if CUDA is present, otherwise CPU); `AUDIOCPP_BACKEND=gpu|cpu` forces it.
-  In CPU mode, the ggml thread count is set automatically to cores-1 (override with `AUDIOCPP_THREADS=N`), and the
-  VRAM warning is no longer shown.
+  In CPU mode, the ggml thread count is set automatically from the **physical** core count — SMT/Hyper-Threading
+  siblings are not counted, and one core is left free above 4 — so a long run leaves the rest of the machine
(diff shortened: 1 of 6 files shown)


0xShug0:
@patrickjchen  Just in time! Really appreciate it! 🚀

PR #107, opened 2026-07-25
Update Docker Docs and Examples (Published Images, GGUF, Qwen3 TTS)

xashr:
## Summary

This PR now updates the Docker documentation and examples to reflect the current state:
- Reference published images from ghcr.io/0xshug0/audio.cpp as primary source for images, local builds still supported
- Switch to GGUF models from https://huggingface.co/audio-cpp/audio.cpp-gguf
- Add Qwen3 TTS as second example to demonstrate voice cloning
- Add ToC, clean up, etc.

PR #109, opened 2026-07-25
Fix: Remove #ifdef GGML_USE_CUDA from CUDA probe in torch_random.cpp

xashr:
## Summary 
When ENGINE_ENABLE_CPU_ALL_VARIANTS=ON (Docker builds), ggml's CMake sets GGML_BACKEND_DL=ON, which prevents GGML_USE_CUDA from being defined as a compile-time macro. The #ifdef GGML_USE_CUDA in resolve_torch_cuda_sampling_policy() then took the #else branch and threw, even though CUDA is fully available at runtime.

The probe code already uses dynamic library loading (open_dynamic_library / dynamic_library_symbol) to query libcuda.so at runtime, so there is no compile-time dependency on CUDA. Removing the #ifdef guards allows the probe to work correctly in both static-link and GGML_BACKEND_DL builds.

Fixes Qwen3 TTS failing with 'requires a build with CUDA support' in CUDA Docker images.

See #84 where ENGINE_ENABLE_CPU_ALL_VARIANTS was introduced
See [#107](https://github.com/0xShug0/audio.cpp/pull/107#issuecomment-5078272077) where this issue was noticed

### DIFF
--- a/src/framework/sampling/torch_random.cpp
+++ b/src/framework/sampling/torch_random.cpp
@@ -312,7 +312,7 @@ TorchCudaSamplingPolicy resolve_torch_cuda_sampling_policy(
         log_default_policy(log_category, "backend is not CUDA");
         return policy;
     }
-#ifdef GGML_USE_CUDA
+    // CUDA runtime probe via dynamic library (works with both static and GGML_BACKEND_DL builds)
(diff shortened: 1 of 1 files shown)


0xShug0:
@xashr Just in time! Thank you!

PR #111, opened 2026-07-25
Add Parakeet-TDT 0.6B v3 (FastConformer-TDT ASR)

dleiferives:
Native port of [`nvidia/parakeet-tdt-0.6b-v3`](https://huggingface.co/nvidia/parakeet-tdt-0.6b-v3),
loaded directly from the Transformers-compatible `model.safetensors` checkpoint
rather than the `.nemo` archive. Offline (full-context) transcription, CPU and
CUDA.

```
Frontend: 16kHz -> 128 mel bins, preemphasis=0.97, NeMo per-feature normalization
Encoder:  3-stage Conv2D subsampling (8x) -> 24-layer FastConformer -> 1024-dim
Decoder:  2-layer LSTM predictor + joint network (TDT: 5 duration classes [0..4])
```

### Tier 1 — golden transcription (CI)

`ctest -R parakeet_golden_transcription_test`, wired into `ENGINE_BUILD_TESTS`,
skips cleanly when the model isn't installed. Asserts the decoded text of a
checked-in LibriSpeech test-clean clip matches NeMo's exactly.

### Tier 2 — numerical parity against NeMo (manual)

`tests/parakeet_tdt/parity/` compares mel features, an isolated encoder layer,
and full encoder output against real NeMo activations captured via forward
hooks (torch 2.13.0+cu130, nemo 2.7.3). Current results, on both backends:

| stage | CPU | CUDA | threshold |
|---|---|---|---|
| `mel_features` | 1.000000 | 1.000000 | >= 0.999 |
| `layer_0` | 1.000000 | 1.000000 | >= 0.999 |
| `enc_out` | 0.972258 | 0.972258 | >= 0.97 |

`enc_out`'s looser bound is float32 summation order across a ~2M-node graph,
not a correctness gap — the README explains why, and it was verified by
chaining 24 isolated single-layer builds, which holds cosine >= 0.999999 the
whole way. That 0.972258 is **identical to the value measured before the
optimization pass**, which is the strongest available statement that none of
the performance work changed the math: it is checked against NeMo directly
rather than against our own earlier output.

**This tier found true bugs that the golden tests missed**

### Tier 3 — multilingual accuracy (new)

`tests/parakeet_tdt/multilingual/` — 120 FLEURS clips across 24 of the 25
supported languages, 21.8 min of audio. Raw transcripts for every clip and
weight type are checked in under `multilingual/results/` so the numbers can be
re-derived without re-running or re-downloading.

This exists because one clip in one language is not enough to clear a
quantization: `q8_0` is byte-identical on the golden clip, and across 24
languages it changes **8.3% of transcripts**.

### Backends

Both, on the same machine, agreeing numerically:

| | CPU (12 threads) | CUDA (GTX 1650 Max-Q) |
|---|---|---|
| encoder compute, 7.4s clip | ~1140 ms | ~124 ms |
| peak host RSS (f32 / q8_0) | 3843 / 1078 MiB | 1818 MiB |
| peak VRAM | — | 2642 of 4096 MiB |

`enc_out` agrees between CPU and CUDA to cosine **0.99999988** (rel RMS
3.5e-06, mel bit-identical) — float32 accumulation noise between different
kernels. Checked with the parity dump's `--backend` flag.

Worth noting: `nemo_asr` **cannot load this model on this card at all** —
PyTorch's overhead on top of the weights exhausts the 4 GB budget. The native
path runs in 2642 MiB.

## Performance

Reference clip (7.435s), Intel i7-9750H (6C/12T) + GTX 1650 Max-Q:

| implementation | settings | CPU | CUDA |
|---|---|---|---|
| NeMo / PyTorch (the Python reference) | as shipped | 857.7 ms (8.7x real-time) | **OOM** — won't load in 4 GB |
| this port | untuned: 8 threads, f32 | ~1245 ms (6.0x) | ~169 ms (44.0x) |
| this port | tuned: 12 threads, `q8_0` | **~715 ms (10.4x)** | **~131 ms (56.8x)** |

Untuned, this port is *slower* than PyTorch on CPU (~1245 ms vs 857.7 ms).
Tuned it wins by ~1.2x, and on this GPU there is no comparison to make: the
PyTorch reference cannot load the model in 4 GB, while the native path runs it
in 2642 MiB. "Untuned" is just the two defaults — `--threads` 8, and
`matmul_weight_type` `native`, which is f32 for this checkpoint.

Measured with a drift-cancelling harness (ABBA ordering, 4 passes, median of 5
iterations per run, machine idle) — this laptop swings 20-30% between cold and
thermally saturated, so sequential before/after timing is not trustworthy at
this scale. Tuned is 1.72x default on CPU, 1.31x on CUDA.

The largest single win was **not** a kernel or precision change — it was a
correctness bug. The encoder graph runs at whatever capacity it was built for,
and `encode()` zero-pads up to that capacity, so a session prepared with a long
audio contract paid the long price on every short clip:

| 7.4s clip, graph built for | encoder compute |
|---|---|
| 7.4s (matched) | 1018 ms |
| 60s (oversized) | 10928 ms |
| 60s contract, fixed | **1140 ms — 9.6x** |

And it was simultaneously wrong. Zero input doesn't stay zero (the subsampling
convs have biases), and the attention mask was unconditionally all-zeros, so
real frames attended to padded garbage. On the reference clip that silently
dropped an entire sentence:

```
matched    "...turning away her eyes. It is certainly very like the old portrait."
oversized  "...turning away her eyes."
```

Fixed in two commits (mask padded keys; rebuild rather than run oversized), and
the same fix carries to CUDA at ~7.5x.

Three further CPU optimizations are **bit-exact** — verified identical across
all 95,232 `enc_out` floats: attention operands read as strided views instead
of copies, constant scales folded into weights (exact, both are powers of two),
and a redundant transpose pair removed.

## Weight precision

`parakeet_tdt.matmul_weight_type` accepts `native|f32|f16|bf16|q8_0`.

| | speed | identical to f32 | absolute WER |
|---|---|---|---|
| `native` (f32) | 1.00x | 100% | 0.1338 |
| `f16` | ~1.00x | 99.2% | 0.1336 |
| `bf16` | 1.14x | 99.2% | 0.1334 |
| `q8_0` | **1.79x** | 91.7% | 0.1331 |

`q8_0` changes 8.3% of transcripts but absolute WER does not move. The churn is
mostly formatting (`T Rex` → `T-Rex`), and where words change it goes both
directions — it is sometimes *right* where f32 was wrong. It costs
reproducibility against an f32 baseline, not quality. It also cuts peak RSS
3.6x, which may matter more than the speed on constrained machines.

`native` stays the default as the only setting that reproduces exactly.

## Build and run

```bash
python3 tools/model_manager.py install parakeet_tdt --models-root models

cmake --build build/<preset> --target parakeet_warm_bench
build/<preset>/bin/parakeet_warm_bench \
    --model models/parakeet-tdt-0.6b-v3 \
    --audio tests/parakeet_tdt/assets/2086-149220-0033.wav \
    --backend cpu --threads 12
```

Package id `parakeet_tdt`, model dir `models/parakeet-tdt-0.6b-v3`.
`tools/check_loader_catalog_sync.py` reports catalog and loaders in sync.

## Known limitations

- **Offline only.** This checkpoint ships `att_context_style="regular"`,
  `att_context_size=[-1,-1]` — unlimited bidirectional attention, not NeMo's
  `chunked_limited` streaming style. `setup_streaming_params()` doesn't error
  but produces a degenerate config (~5.8s chunks, ~10000-frame cache) with none
  of the real benefit. NVIDIA's maintainers point at a different architecture
  for streaming. Real support needs a purpose-trained checkpoint; this repo
  already has `nemotron_asr` for that. 
- Per-language WER in the study is 5 clips each — enough to show no language
  falls off a cliff under quantization, not publishable per-language figures.
- CUDA numbers come from a 35 W GTX 1650 Max-Q, compute capability 7.5, which
  is below the 8.0 threshold where ggml enables CUDA graph capture. Several
  optimizations measured as *no change* on CUDA; that is this card, not the
  optimization. Worth re-measuring on Ampere or newer.

## Docs

- `docs/community_models/parakeet_tdt.md` — user-facing (162 lines)
- `docs/reports/parakeet_tdt_performance.md` — full methodology, profiling, and
  the optimizations tried and rejected with numbers, so they don't get
  re-attempted
- `tests/parakeet_tdt/{parity,multilingual,bench}/README.md` — harnesses


0xShug0:
@dleiferives Thank you for your contribution! I'd appreciate it if you could update the spec to follow the current model_specs_v1/ format, but without adding a schema version number. The existing code will simply ignore the new v1 fields, so this won’t break anything. This is just to make the eventual migration easier.

dleiferives:
> @dleiferives Thank you for your contribution! I'd appreciate it if you could update the spec to follow the current model_specs_v1/ format, but without adding a schema version number. The existing code will simply ignore the new v1 fields, so this won’t break anything. This is just to make the eventual migration easier.

Yeah, of course. Just implemented & updated pr body to denote work and results more accurately.

0xShug0:
@dleiferives You can check my old impl of https://github.com/0xShug0/audio.cpp/tree/preview/parakeet_tdt_old. That implementation was done before the 0.1 release and isn't very polished, but it may still be useful.

0xShug0:
@dleiferives Please check https://github.com/0xShug0/audio.cpp/issues/128. The example is Confucius4-TTS.

dleiferives:
Awesome, that is super helpful thank you.

dleiferives:
Implemented the requested Parakeet follow-up and aligned it with the schema-v1 flow described in #128, using Confucius4-TTS as the reference shape.

What changed:

- `model_specs/parakeet_tdt.json` is now the single source of truth and declares `schema_version: 1`, metadata, tasks/modes, package sources, and the complete option contract.
- Removed the duplicate `model_specs_v1` entry and the repetitive Parakeet `loader.cpp` / `loader.h` path. The loader factory now lives beside the session and uses the generic spec-backed loader.
- Normalized the old chunk setting to `audio_chunk_threshold_sec`, matching the naming rules in #128, and enforce the declared duration minima at runtime.
- All declared enum options are now validated centrally across both offline and streaming session creation, so an invalid value cannot be silently ignored merely because it belongs to the other mode.
- Added timestamp-aware TDT decoding, word-span assembly, bounded-window long-form processing, and buffered left/center/right streaming with incremental partial/final transcript delivery.
- The streaming behavior is described accurately as buffered/pseudo-streaming: this checkpoint uses unlimited bidirectional attention and is not a native cache-aware streaming model.
- After the latest upstream merge, Parakeet is also registered as a first-class selectable model in the new composite build system; the generated registry includes its spec-backed loader correctly.

I reviewed `preview/parakeet_tdt_old` as suggested and adapted the useful pieces—emission-frame timing, persistent predictor state, bounded window scheduling, center-only incremental decoding, and transcript merging—to the current framework APIs. The relevant commits explicitly cite that branch where implementation ideas were carried forward.

Validation on the merged tree:

- `model_spec_system_test` passed
- `tdt_decoder_duration_loop_test` passed
- `parakeet_golden_transcription_test` passed
- `parakeet_streaming_transcription_test` passed
- Focused CPU build passed with `AUDIOCPP_MODEL_SET=custom` and `AUDIOCPP_MODELS=parakeet_tdt`
- Full composite configuration includes `parakeet_tdt`, and the loader/catalog synchronization check passes

The main schema/streaming work is in `5ad45e3` through `0a93eb4`; `f1e1333` resolves the latest upstream composite-build merge while preserving the Parakeet registration.

0xShug0:
@dleiferives Merged! Thank you for contributing the new model!

PR #98, opened 2026-07-24
GLM-TTS model

mirek190:
# GLM-TTS

This  is a community GLM-TTS implementation, its prepared
safetensors package, and a standalone mixed Q8/F16 GGUF. Measurements were
captured per request, with deterministic Python parity completed before the
normal sampled path was measured.

## Implementation scope

- Adds the community `glm_tts` loader and offline `tts` / `clon` session.
- Runs the ChatGLM tokenizer, Llama speech-token generator, Whisper-VQ
  reference encoder, CAMPPlus speaker encoder, Flow/DiT mel generator, and
  HiFT vocoder natively through the shared framework.
- Adds the `glm_tts` model-manager package, package spec, CLI/server registry
  entry, user documentation, converter, and validation probes.
- Supports prepared safetensors packages and standalone multi-component GGUF
  files with embedded configs, tokenizer files, and package metadata.
- Keeps reusable component runtimes resident, caches prepared references,
  reuses shape-compatible graphs, and restricts the Llama output projection to
  GLM-TTS audio-token rows.
- Extends the shared OuteTTS Llama runtime, CAMPPlus encoder, and Whisper
  embedding surface only where GLM-TTS needs reusable framework behavior.

## Test environment

Tests were run on 2026-07-24 with:

- Windows 11 and MSVC 2022 14.43
- NVIDIA RTX 3090 24 GB
- CUDA toolkit 12.4.131
- PyTorch 2.5.1+cu121 and Transformers 5.13.0
- GLM-TTS branch based on audio.cpp main `f8fb0c1`

The official Python reference was checked out at:

```text
..\_reference\GLM-TTS
```

The prepared package id and paths were:

```text
model-manager package: glm_tts
safetensors package:   ..\models\GLM-TTS
standalone GGUF:       ..\models\GLM-TTS-Q8\GLM-TTS_Q8.gguf
reference WAV:         ..\SAMPLES\EN_2.wav
reference transcript:  If you actually care about security
```

Install the package with:

```powershell
python tools\model_manager.py install glm_tts --models-dir ..\models
```

The installer downloads `zai-org/GLM-TTS`, converts Flow and HiFT from the
official PyTorch checkpoints to safetensors, exports the tokenizer resources,
and installs matching CAMPPlus safetensors weights.

## Exact build commands

CPU:

```powershell
.\scripts\build_windows.ps1 `
  -Preset windows-cpu-release -ConfigureOnly

cmake -S . -B build\windows-cpu-release `
[...]

PR #123, opened 2026-07-27
Correct GLM-TTS model_specs_v1 metadata

mirek190:
## Summary

Follow-up to #98 that aligns GLM-TTS with the normalized `model_specs_v1`
option contract while preserving its existing request keys as compatibility
aliases.

## Changes

- keeps the normalized v1 request names:
  `num_inference_steps`, `flow_guidance_scale`, `flow_noise_path`,
  `hift_source_random_path`, and `hift_prior_noise_count`
- advertises those normalized names through the loader
- maps normalized names to the existing GLM-TTS implementation
- retains `flow_steps`, `cfg_rate`, `flow_noise_file`,
  `hift_source_random_file`, and `hift_prior_noise_values` as backward-compatible
  aliases
- keeps language metadata scoped to Chinese and English
- removes claims for streaming and explicit phoneme/emotion controls that the
  current native runtime does not expose
- makes Llama context-size minimums match runtime validation
- describes the standalone model accurately as mixed Q8_0/F16
- installs the package into the documented and tested `GLM-TTS-Q8` directory
- preserves the existing SafeTensors/GGUF source layouts

## Why

`model_specs_v1` is the normalized public contract used by framework validation,
catalog clients, and generated UIs. It must not mirror model-specific legacy
implementation names. Compatibility belongs at the model boundary, so new
clients get one consistent vocabulary without breaking existing GLM-TTS
commands.

## Impact

Normalized v1 requests now work directly. Existing requests using the old
GLM-TTS keys continue to work. The option resolver also rejects conflicting
values when both a normalized name and its legacy alias are supplied.

Inference math, weights, package loading, and generated output are unchanged.

## Validation

- parsed `model_specs_v1/glm_tts.json`
- confirmed all normalized request names against the framework contracts
- confirmed v1 `sources` remain identical to `model_specs/glm_tts.json`
- resolved `glm_tts_q8_0` with `model_manager_v2.py`
- verified the remote GGUF at `5,143,813,728` bytes
- built the complete `engine_runtime` CPU target
- built `glm_tts_warm_bench` with normalized option ingestion
- `git diff --check`


mirek190:
This is a metadata-only follow-up to #98, based on current main `41246aa`.

## Why this update was needed

The migrated GLM-TTS v1 entry contained five descriptive option aliases that are not the public keys consumed by `GlmTTSSession`. A catalog/UI generated from that entry would therefore send ignored options.

| Previous v1 name | Runtime key |
|---|---|
| `num_inference_steps` | `flow_steps` |
| `flow_guidance_scale` | `cfg_rate` |
| `flow_noise_path` | `flow_noise_file` |
| `hift_source_random_path` | `hift_source_random_file` |
| `hift_prior_noise_count` | `hift_prior_noise_values` |

The description also mixed upstream GLM-TTS capabilities with what audio.cpp currently exposes. Upstream supports streaming and optional phoneme input, but the native audio.cpp contribution currently advertises offline `tts`/`clone` and accepts ordinary text plus reference audio/transcript. The spec now describes that implemented surface only. Chinese and English remain the correct language metadata.

## Package metadata corrections

The published standalone file is the quality-tested mixed model: the autoregressive Llama tensors remain F16 while speech-tokenizer, Flow, HiFT, and CAMPPlus tensors are Q8_0. The display name and package description now say this explicitly. Its install target is corrected from `GLM-TTS-GGUF` to `GLM-TTS-Q8`, matching the existing documentation and validation paths.

The SafeTensors and GGUF `sources` arrays are byte-for-byte structurally identical to `model_specs/glm_tts.json`, and no `schema_version` field was added.

## Validation

- JSON parse: pass
- v1/runtime-bridge source-layout equality: pass
- tasks: `tts`, `clone`
- languages: `zh`, `en`
- `model_manager_v2.py info glm_tts`: pass
- remote package check: pass (`5,143,813,728` bytes)
- `git diff --check`: pass

No C++ inference code changed, so audio generation behavior and existing GGUF compatibility are unaffected.

0xShug0:
@mirek190 Sorry, I didn't explain this clearly. The v1 spec uses normalized option keys so that different models don’t use different names for the same concept. It also allows the framework to validate certain options centrally (topk etc), instead of duplicating validation logic in every model.

When we migrate existing models, we can add a compatibility layer for each model to map the old keys in the requests to the new normalized keys, or just directly align model names with normalized v1 keys. New models should use the normalized keys directly, or at least follow the same naming conventions for any model-specific option keys.  

I’m currently working on releasing a model early to serve as a reference implementation for future v1-native new models.

mirek190:
Thanks for the clarification. Commit `81a2c28` revises this PR to follow the intended v1 direction.

- Restored the normalized spec names: `num_inference_steps`, `flow_guidance_scale`, `flow_noise_path`, `hift_source_random_path`, and `hift_prior_noise_count`.
- Updated loader metadata and GLM-TTS documentation to advertise those normalized names.
- Added an additive model-boundary compatibility mapping so the existing `flow_steps`, `cfg_rate`, `flow_noise_file`, `hift_source_random_file`, and `hift_prior_noise_values` keys continue to work.
- Normalized keys are checked first; if both forms are supplied with different values, the shared resolver reports a conflict rather than choosing one silently.
- Updated the GLM warm-bench input path to exercise the normalized names.

Validation: JSON/model-manager checks pass, v1/runtime source layouts remain equal, the complete CPU `engine_runtime` target builds, `glm_tts_warm_bench` builds, and `git diff --check` passes. Inference math and generated outputs are unchanged.

PR #127, opened 2026-07-27
Emit streaming transcript deltas instead of restating the transcript

patrickvonplaten:
Last remaining piece of #112, which is down to this one change after #116 and #118 landed. That PR can be closed once this merges.

## Problem

`voxtral_realtime` set `partial_text` to a decode of **every token so far**, so each streaming update restated the whole transcript:

```
partial_text= Some
partial_text= Some call
partial_text= Some call me
partial_text= Some call me nature
```

Three things follow from that:

1. **It is the odd one out.** `vibevoice_asr`, `higgs_audio_stt` and `nemotron_asr` all emit only the newly decoded text. `voxtral_realtime` was the only session restating.
2. **The server is wrong for it.** `app/server/runtime.cpp:1101` sends `partial_text` as `transcript.text.delta`. For the other three that is a delta; for `voxtral_realtime` a client concatenating deltas rebuilt the pyramid.
3. **It is quadratic** in the transcript length. Modelled on a one-hour session at 150 wpm — 54,000 chars, ~13,500 partials — that is **364 MB** written against **54 KB** of actual transcript.

## Change

`record_stream_token` emits the new token's text rather than re-decoding the whole list, which fixes all three at the source. `finalize()` no longer repeats the transcript as a partial either — every token has already been delivered, and the whole transcript stays available as `result.text_output`.

The CLI printed each partial on its own labelled line. Now that partials concatenate, a terminal gets them appended unlabelled and the transcript scrolls like ordinary output:

```
 Some call me nature. Others call me Mother Nature. I've been here for over 4.5 billion years.
```

Redirected output keeps the `partial_text=` line-per-update form so pipes and logs stay parseable.

**Rejected approach.** An earlier revision of this PR diffed successive partials in the CLI instead. That is not sound: the CLI cannot tell which convention a session follows, and diffing the three delta-emitting models destroyed their output — feeding ` Some`, ` call`, ` me`, ` nature` through it left only ` nature` on screen, because each delta was mistaken for a revision of the last.

**Behaviour change worth calling out:** for `voxtral_realtime`, a redirected `partial_text=` line now carries a delta rather than the running transcript. Consumers concatenate. This aligns it with the three other streaming ASR models, where `partial_text=` already meant a delta. `text_output=` at the end is unchanged.

## Why?

Voxtral Realtime was trained to transcribe hour-long sessions which seems to work well now. However when transcribing such long audios live, it becomes prohibitely expensive to output the whole transcript at every timestamp as this results essentially to O(N2) amount of work. This PR adapts the output rendering so that we only render the partials.

Command to test:

```sh
ffmpeg -f avfoundation -i ":0" -ar 16000 -ac 1 -f s16le - 2>/dev/null \
  | ./build/macos-metal-release/bin/audiocpp_cli \
      --model /models/Voxtral-Mini-4B-Realtime-2602-GGUF/voxtral-mini-4b-realtime-2602-q8_0.gguf \
      --backend metal --mode streaming --audio -
```

should also work when wanting to transcribe to a file directly as explained in the asr.md doc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1581,6 +1581,13 @@ if (ENGINE_BUILD_TESTS)
     # block forever rather than fail. Bound it so CI reports a failure instead.
     set_tests_properties(http_live_body_test PROPERTIES TIMEOUT 120)
 
+    add_engine_unittest(partial_text_render_test tests/unittests/test_partial_text_render.cpp)
+
+    add_test(
+        NAME partial_text_render_test
+        COMMAND partial_text_render_test
+    )
+
     add_engine_unittest(hf_sampler_test tests/unittests/test_hf_sampler.cpp)
 
     add_test(
--- a/app/cli/main.cpp
+++ b/app/cli/main.cpp
@@ -1,5 +1,6 @@
 #include "args.h"
 #include "batch.h"
+#include "partial_render.h"
 #include "request.h"
 #include "../streaming/pcm_source.h"
 #include "../streaming/streaming.h"
@@ -18,6 +19,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstddef>
+#include <cstdio>
 #include <fstream>
 #include <filesystem>
 #include <iostream>
@@ -32,7 +34,10 @@
 #ifndef NOMINMAX
 #define NOMINMAX
 #endif
+#include <io.h>
 #include <windows.h>
+#else
+#include <unistd.h>
 #endif
 
 #ifdef _OPENMP
@@ -461,6 +466,16 @@ bool stream_audio_from_stdin(int argc, char ** argv) {
     return audio.has_value() && minitts::cli::is_stdin_audio_source(*audio);
 }
 
+// Only a terminal can take partials as running text; a redirected stream has to keep the
+// line-per-update format so pipes and logs stay parseable.
+bool stdout_is_terminal() {
+#ifdef _WIN32
+    return _isatty(_fileno(stdout)) != 0;
+#else
+    return isatty(fileno(stdout)) != 0;
+#endif
+}
+
 // Feeds raw PCM from stdin into the session chunk by chunk, so nothing has to be buffered up
 // front and transcription tracks the input as it arrives.
 engine::runtime::TaskResult run_streaming_from_stdin(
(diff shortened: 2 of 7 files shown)


patrickvonplaten:
This should be the last PR for Voxtral Realtime! Will try to advertise it a bit once in - think we have a cool & fast use case now for running it locally.

0xShug0:
@patrickvonplaten Thanks! I'll test it soon.

0xShug0:
@patrickvonplaten Tested using local Q4 Voxtral GGUF and assets/resources/sample_16k.wav

```
default: partial_lines=33 partial_len=114 text_output_len=114 match=True
batch4:  partial_lines=19 partial_len=76  text_output_len=114 match=False
```

For stream_batch_tokens=4, final text_output was complete, but concatenated partial_text= deltas were missing words:
`partial=" Some me. Others me Mother Nature.'ve been for over5 billion. 5 times longer"`

Root cause:The steady-state decoder loops over multiple token rows and calls `record_stream_token()` for each token, but all calls write into the same StreamEvent. Only one event is emitted per chunk in session.cpp, so `record_stream_token()`  overwrites earlier partial_text deltas. This was safe when partials restated the whole transcript, but not after switching to per-token deltas.

0xShug0:
@patrickvonplaten Could you take a look at this?

patrickvonplaten:
> @patrickvonplaten Could you take a look at this?

Ah yeah nice catch, will adapt!

patrickvonplaten:
Ok think should be fixed - also added a couple more tests

0xShug0:
@patrickvonplaten Merged! The Voxtral streaming path is in good shape.

```
batch=1 partial_lines=33 partial_len=114 text_output_len=114 match=True
batch=2 partial_lines=26 partial_len=114 text_output_len=114 match=True
batch=4 partial_lines=19 partial_len=114 text_output_len=114 match=True
batch=7 partial_lines=16 partial_len=114 text_output_len=114 match=True
```

patrickvonplaten:
Adapted the model card of the quant checkpoints a bit and did some ads in Mistral Reddit: https://www.reddit.com/r/MistralAI/comments/1v9k9s7/voxtral_realtime_at_13x_realtime_on_an_m3_air/ 

Once next Voxtral Streaming drops we should probably try to also advertise in llama.cpp reddit among others

0xShug0:
Definitely! Hopefully, audio.cpp can eventually become the default way to release new audio models and reduce setup pain for users 😄

Issue #113, opened 2026-07-25
whisper-medium


kardinalisimo:
I am using a custom windows build and the webgui. Trying to run the Vevo2 model. 
First, I thought I will need just a guff file. It turned I needed a bunch of config and model.safetensors files as well. I finally got it. Now it askes for Whisper-medium. Downloaded the guff as well, few different variants. Nothing works. Kepp getting :  audiocpp_server failed: missing required json key: preprocess  . I even tried to inject it in the gguf and did not work. 
Any help will be appreciated.

0xShug0:
@kardinalisimo Have you tried the model manager tool? It will download all the dependencies and create the model package for you. 

For GGUFs: Looks like HF is still hosting an old version of the GGUFs with missing metadata. I don't know why. I will reupload.

kardinalisimo:
I even don't have active model manager tool in the gui. Am I missing something?  I was downloading everything manually
So, does all guff models need extra files for them to work? As to Whisper, I tried even with Qwen ASR but same error.
Also, the gui does not have the new models from the drop down menues.

0xShug0:
Can you update your repo and use the latest webui (updated a few hours ago). It may solve the issue.  The latest UI does show the new models in the release 0.4.

Also model manager tool is another tool in the repo.

kardinalisimo:
I will do it later and let you know how it behave. I will definitely benefit from automated downloads.

0xShug0:
@patrickjchen Looks like there is a gap in model download in the WebUI. It still prefers safetensor over GGUFs from https://huggingface.co/audio-cpp/audio.cpp-gguf and only looks for "model.gguf". Can you take a look?

kardinalisimo:
I tricked it by creating fake files but should be better way.

0xShug0:
@kardinalisimo Please check the latest WebUI in the repo.

kardinalisimo:
I have the latest webui update and server build and when try to use the download tool for vevo2, it downloads the model.safetensors.  When I do convert I am getting this -❌ This model backend does not currently support native GGUF..  Inspect says ⚠️ No GGUF yet.
I downloaded the fish studio and came clean in gguf. 
So, seems like some models are good and some still not pulling gguf. Not sure which one because I have no intentions to download them.

0xShug0:
[...]

PR #130, opened 2026-07-28
Fix for #113, and related model downloading issues

patrickjchen:
Fix GGUF package loading and downloading (#113)

A model directory only counted as a GGUF package when it held a file named
exactly model.gguf, so published packages under their release name fell
through to the safetensors source and demanded config files the GGUF never
ships — the "missing required json key: preprocess" in #113. Separately, the
WebUI downloader ignored strip_prefix, installing every audio-cpp/audio.cpp-gguf
package one directory too deep, so those downloads never validated.

Adds the self-contained vevo2_gguf package, and reworks the WebUI download
flow to state its cost (size, disk utilization, VRAM or system RAM share),
alarm past 75% disk / 80% device memory, and require an explicit confirmation
before fetching anything.


0xShug0:
@patrickjchen Thanks! I'll take a look after merging the spec v1 branch, which should make both UI integration and model development easier. If your PR has any conflicts with the updated main, I'll resolve them directly.

0xShug0:
Merged! Great extension for both the framework and UI.

PR #135, opened 2026-07-28
Fix Windows build under MSYS2 / MinGW GCC 16

noctrex:
Build the project with MSYS2 (UCRT64, MinGW-w64 GCC 16.1). Two issues blocked it; both are Windows-only, no behavior change elsewhere.

1. CLI entry point failed to link (app/cli/main.cpp) MinGW links with the plain-C CRT startup, which calls `main`. The `wmain` entry is only selected with `-municode`, which the project does not pass, so the link failed with "undefined reference" / ld error 5. Replace `wmain` with `main` that re-derives UTF-8 argv via CommandLineToArgvW(GetCommandLineW()). This needs no extra link flag, works on both MSVC and MinGW, and preserves the original UTF-8 argument handling. Aligns the CLI with the server, which already uses plain `main`.

2. NOMINMAX redefinition warning (app/server/http.cpp) GCC 16's MinGW C++ runtime (bits/os_defines.h) now auto-defines `NOMINMAX 1` whenever a standard header is included. http.cpp includes std headers before its unguarded `#define NOMINMAX`, so the empty-body redefine collides and trips -Wmacro-redefined. Wrap the define in `#ifndef NOMINMAX`, matching the pattern already used in the four other NOMINMAX sites in the repo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1144,6 +1144,14 @@ add_executable(audiocpp_cli
 )
 
 target_link_libraries(audiocpp_cli PRIVATE engine_runtime ggml)
+# MinGW selects the wmain() entry point only when linked with -municode;
+# without it the CLI fails to link (undefined reference / ld error 5).
+# MSVC detects wmain natively, so this is MinGW-only. Link option only:
+# -municode also defines UNICODE at compile time, so passing it link-only
+# keeps the source compiled with no macro side effects.
+if (MINGW)
+    target_link_options(audiocpp_cli PRIVATE -municode)
+endif()
 if (ENGINE_ENABLE_OPENMP)
     target_link_libraries(audiocpp_cli PRIVATE OpenMP::OpenMP_CXX)
 endif()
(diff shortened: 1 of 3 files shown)


0xShug0:
@noctrex Thank you for the PR. I can't validate it on Windows right now, but adding shellapi.h looks potentially risky to me. Do you need to link Shell32.lib for MSVC builds?

noctrex:
> @noctrex Thank you for the PR. I can't validate it on Windows right now, but adding shellapi.h looks potentially risky to me. Do you need to link Shell32.lib for MSVC builds?

OK so I tried it without shellapi and it's still compiled, so I guess it was not needed after all, I just added it out of habit.
And after a bit more digging I realized I didn't even need to touch main.cpp,  MinGW just needs -municode at link time to pick up wmain(). So I reverted that file and pass the flag instead.
The NOMINMAX guard in http.cpp is still needed as it won't compile in MSYS2 without it.
Sorry for the rushed PR!

0xShug0:
@noctrex Thanks! PR merged.

PR #138, opened 2026-07-28
fix(safetensors): validate data_offsets at parse time

chikosan:
## Problem

`data_offsets` is read straight from the safetensors header and cast to `size_t` with no validation:

```cpp
info.data_begin = static_cast<size_t>(offsets[0]);
info.data_end   = static_cast<size_t>(offsets[1]);
```

`parse_int_array` returns `int64_t`, so two shapes get through:

- a **negative** value becomes `~SIZE_MAX` on the cast
- **`data_end < data_begin`** makes the `data_end - data_begin` that every consumer computes underflow to the same

Consumers then do:

```cpp
const size_t data_offset = index_.header_bytes + info.data_begin;
const size_t byte_size   = info.data_end - info.data_begin;
if (data_offset + byte_size > bytes_.size()) { throw; }
return {bytes_.data() + data_offset, byte_size};
```

With an underflowed `byte_size`, `data_offset + byte_size` wraps and the comparison passes, so the guard does not fire and the function returns a span far larger than the mapped file — an out-of-bounds read driven by a model file's header.

Reachable wherever a `.safetensors` file is loaded from a source the operator doesn't control (a downloaded or user-supplied model). Affects both `framework/assets/tensor_source.cpp` and `models/pocket_tts/voice_state_assets.cpp`, which contain the same computation.

## Fix

**Validate at parse time**, in `safetensors.cpp`, which is the single point every consumer routes through — rather than patching each call site:

- reject negative `data_offsets`
- reject `data_end < data_begin`

**Overflow-safe bounds checks** in both consumers as defence in depth: the comparisons are rewritten as subtractions against the total (`byte_size > total - header_bytes`, etc.) so no term can overflow. Parse-time validation makes that unreachable now, but these are the last gate before a raw pointer escapes.

Zero-length tensors (`data_begin == data_end`) remain valid — covered by a test.

## Test

`tests/unittests/test_safetensors_offsets.cpp`, registered in CMake alongside the other unit tests. Covers inverted offsets, negative offsets, and that valid headers still parse.

Verified by compiling both ways: the test **fails** against the unpatched parser (`inverted data_offsets must be rejected`) and **passes** with the fix.

⚠️ I compiled and ran this test directly rather than through the full CMake build, so the change hasn't been through the project's complete build matrix — worth a CI run before merging.

## Provenance

Found by an automated security scan while evaluating that tooling; the underflow was then traced and confirmed by hand. Happy to adjust the approach if you'd rather validate elsewhere.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1391,6 +1391,13 @@ if (ENGINE_BUILD_TESTS)
         COMMAND wav_reader_test
     )
 
+    add_engine_unittest(safetensors_offsets_test tests/unittests/test_safetensors_offsets.cpp)
+
+    add_test(
+        NAME safetensors_offsets_test
+        COMMAND safetensors_offsets_test
+    )
+
     add_engine_unittest(audio_chunking_test tests/unittests/test_audio_chunking.cpp)
 
     add_test(
--- a/src/framework/assets/tensor_source.cpp
+++ b/src/framework/assets/tensor_source.cpp
@@ -638,11 +638,19 @@ private:
         if (bytes_.empty()) {
             bytes_ = engine::io::read_binary_blob(index_.source_path);
         }
-        const size_t data_offset = index_.header_bytes + info.data_begin;
         const size_t byte_size = info.data_end - info.data_begin;
-        if (data_offset + byte_size > bytes_.size()) {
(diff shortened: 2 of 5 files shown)


0xShug0:
Thank you @chikosan! PR merged.

PR #143, opened 2026-07-28
fix: bound allocations sized by attacker-controlled file and header fields

chikosan:
Follow-up to #138, same family: a value read out of an untrusted file or header sizes an allocation directly.

## 1. WAV `data` chunk — `framework/audio/wav_reader.cpp`

```cpp
const uint32_t chunk_size = read_scalar<uint32_t>(input);
...
data.resize(chunk_size);   // commits up to 4 GiB before reading a byte
input.read(data.data(), chunk_size);
if (!input) { throw; }     // only fails *after* the allocation
```

A 50-byte WAV declaring `chunk_size = 0xFFFFFFFF` allocates 4 GiB, then throws. Now grows in 1 MiB blocks as data actually arrives, so an unbacked claim fails after one block.

**Measured on that input:**

| | peak RSS |
|---|---|
| before | 4,296,540,160 B (4.0 GiB) |
| after | 2,605,056 B (2.5 MiB) |

## 2. Tokenizer vocab — `framework/tokenizers/hf_tokenizer_json.cpp`

```cpp
max_id = std::max(max_id, static_cast<size_t>(value.as_i64()));
std::vector<std::string> id_to_token(max_id + 1);
```

`as_i64()` is unbounded. A negative id becomes `~SIZE_MAX` on the cast; a large positive one asks for terabytes of `std::string`.

A `tokenizer.json` declaring id `10^12` **does not throw — the process is OOM-killed (exit 137)**. Ids are now range-checked against `2^24`, roughly two orders of magnitude above the largest published vocabulary.

## 3. HTTP request body — `app/server/http.cpp`

```cpp
content_length = static_cast<size_t>(std::stoull(it->second));
...
while (request.body.size() < content_length) { /* append recv() data */ }
```

No bound, so a single request with an absurd `Content-Length` drives unbounded in-memory accumulation. `std::stoull` also throws on a malformed header, which surfaces as a 500 rather than a 400.

Parsed with `from_chars` and bounded by `kMaxRequestBody`.

> ⚠️ **`kMaxRequestBody` is a policy choice I can't make for you.** I set it to 2 GiB — generous, since long audio uploads are the point of this server, but finite. Please set it to whatever you actually intend to accept; it's a single named constant.

## Tests

`tests/unittests/test_wav_reader_chunk_bounds.cpp` and `test_hf_tokenizer_vocab_bounds.cpp`, both registered in CMake. Cover oversized and truncated WAV chunks, out-of-range and negative token ids, and that ordinary files still load.

Verified against unpatched code: the tokenizer test is OOM-killed, and the WAV case allocates 4 GiB (numbers above).

⚠️ No test for the HTTP change — it needs a socket, and I didn't want to add a test harness shape the project doesn't already have. I compiled all three files and ran the two unit tests directly rather than through the full CMake build, so a CI run is worth having before merge.

## Provenance

Same automated scan as #138; each finding was then traced and confirmed by hand.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1418,6 +1418,20 @@ if (ENGINE_BUILD_TESTS)
         COMMAND wav_reader_test
     )
 
+    add_engine_unittest(wav_reader_chunk_bounds_test tests/unittests/test_wav_reader_chunk_bounds.cpp)
+
+    add_test(
+        NAME wav_reader_chunk_bounds_test
+        COMMAND wav_reader_chunk_bounds_test
+    )
+
+    add_engine_unittest(hf_tokenizer_vocab_bounds_test tests/unittests/test_hf_tokenizer_vocab_bounds.cpp)
+
+    add_test(
+        NAME hf_tokenizer_vocab_bounds_test
+        COMMAND hf_tokenizer_vocab_bounds_test
+    )
+
     add_engine_unittest(safetensors_offsets_test tests/unittests/test_safetensors_offsets.cpp)
 
     add_test(
--- a/app/server/README.md
+++ b/app/server/README.md
@@ -22,6 +22,7 @@ cat > server.json <<'JSON'
   "device": 0,
   "threads": 1,
   "lazy_load": true,
+  "max_request_body_bytes": 2147483648,
   "models": [
     {
       "id": "pocket-tts",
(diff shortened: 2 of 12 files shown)


0xShug0:
@chikosan Merged! Updated the PR to make kMaxRequestBody configurable. I like this type of PR. As a (former) security researcher who finds vulnerabilities in other people’s systems... shame on me 😂

PR #144, opened 2026-07-28
server: add POST /v1/audio/transcriptions/live for live PCM ingest

5uck1ess:
## What this is

The engine already does incremental transcription. `audiocpp_cli --audio -` reads PCM from a pipe and emits partial transcripts as speech arrives — `AudioChunkStream`, `run_streaming_task`, `StreamingInputKind::AudioChunks` are all already there and already tested.

The HTTP server can't reach any of it. Every ASR endpoint buffers the entire request body before the model is touched, so a client whose only interface to audio.cpp is HTTP has no way to get a partial transcript, regardless of which model it loads. This PR is the missing transport, not a new capability.

## The endpoint

```
POST /v1/audio/transcriptions/live?model=<id>&sample_rate=16000&channels=1&sample_format=s16le
Transfer-Encoding: chunked

<raw PCM, written continuously>
```

SSE deltas stream back on the same connection while the upload is still in progress:

```
data: {"type":"transcript.text.delta","delta":"so I was"}
data: {"type":"transcript.text.delta","delta":" thinking"}
...
data: {"type":"transcript.text.done","text":"..."}
data: [DONE]
```

| param | default | notes |
|---|---|---|
| `model` | — | required, must be `mode: "streaming"` |
| `sample_rate` | `16000` | 1000–384000 |
| `channels` | `1` | 1–16 |
| `sample_format` | `s16le` | or `f32le` |
| `language` | — | passed through to the model |

`sample_rate` and `channels` are range-checked rather than merely validated as positive: they are multiplied into the per-chunk sample count that sizes a buffer allocated *after* the model lock is taken, so unbounded values are an allocation request made while holding the lock. Every malformed parameter returns `400` with the specific reason.

Working example:

```bash
ffmpeg -f alsa -i default -ar 16000 -ac 1 -f s16le - 2>/dev/null \
  | curl -N -X POST -H 'Expect:' -T - \
    'http://localhost:8080/v1/audio/transcriptions/live?model=voxtral-realtime'
```

`-T -` is load-bearing and not interchangeable with `--data-binary @-`: the latter drains stdin to EOF before opening the connection, which turns a live capture back into a file upload. `-H 'Expect:'` suppresses `Expect: 100-continue`, which this server does not answer.

## Why query params instead of a JSON envelope

The body is the audio — there's nowhere to put a JSON envelope without either framing the audio inside it or spending a second request on setup. Deepgram's `/v1/listen` takes raw audio as the body with `encoding`/`sample_rate`/`channels` as query params for the same reason. This is not an OpenAI-compatible endpoint and doesn't claim to be; `/v1/audio/transcriptions` keeps its existing shape untouched.

## Why not a WebSocket

A WebSocket would mean a framing layer, a handshake, and a second server loop, for a stream that is one-directional in and one-directional out. HTTP/1.1 already permits a request body and a response body to be in flight simultaneously on one connection, and that's exactly the shape of the problem. If a WebSocket surface is wanted later it can sit on top of the same `AudioChunkStream` plumbing this PR adds — nothing here forecloses it.

## Measured on a 3090

Whether you get partials *during* speech is a property of the model, not of this transport. Two separate runs, stated with their conditions rather than merged into one table, because they used different recordings.

**The reproducible one**, through the exact `curl` command above, feeding a 20 s recording at real-time pace — re-run after every change in this PR, with identical results each time:

| | voxtral-realtime |
|---|---|
| first delta | 1.50 s |
| deltas arriving while audio was still uploading | 63 / 63 |
| `transcript.text.done` | at the moment the audio ended |

**A model comparison** on a shorter clip, included because the difference matters when picking a model:

| model | first delta | deltas before speech ended | tail after last audio |
|---|---|---|---|
| voxtral-realtime | 1.44 s | 27 / 28 | 12–15 ms |
| nemotron_asr | 9.7 s | 0 | ~600 ms |
| faster-whisper (batch, for reference) | n/a | never | ~187 ms |

nemotron buffers internally, so over this endpoint it behaves like a batch model with extra steps. It works; it just doesn't feel live. The docs say so plainly rather than implying the endpoint makes any model live.

## Implementation

- `HttpRequest::body_stream` (`std::istream *`) is non-null **only** when the client declared `Transfer-Encoding: chunked` **and** targeted this exact path. Every other endpoint keeps the existing buffered `request.body` path, byte for byte.
- `ChunkedSocketStreambuf` de-frames chunked encoding incrementally, bounded on four axes: 30 s idle, 600 s absolute, 512 MiB total, 8 MiB per chunk.
- Chunk sizes are parsed with an overflow-safe hand-rolled hex loop (a declared `ffffffffffffffff` is rejected, not wrapped). Chunk data is verified CRLF-terminated so a desynchronised sender's payload can't be reinterpreted as framing. Trailers are capped.
- **A peer that disconnects without the terminating chunk is an error, never a short body.** For audio, a silently truncated body is indistinguishable from a speaker who stopped talking, so it must not be reported as success. The stream is configured `exceptions(badbit)` so this can't be swallowed.
- `run_streaming_model` and `run_streaming_model_from` now share `run_streaming_model_impl`, which selects between `request.audio_input` and a live `AudioChunkStream`. No behaviour change for existing callers.
- `app/streaming/pcm_source.cpp` was CLI-only; it's now also linked into `audiocpp_server`.

## Things you should know before merging

Stated up front rather than left to be found:

1. **HTTP/1.1 duplex is not universally proxied.** Behind nginx it needs `proxy_http_version 1.1` as well as `proxy_request_buffering off` — the latter alone still buffers a chunked body. Browser `fetch()` cannot drive it at all: request streaming is half-duplex by specification. This endpoint targets server-side and native clients, and `app/server/README.md` says so.
1. **`Expect: 100-continue` is not implemented.** A client that sends it gets no interim response and waits out its own continue timeout before uploading. curl adds it for some POSTs, so the documented example passes `-H 'Expect:'`. Worth implementing properly if you'd like — I left it out to keep this change to one concern.
2. **`SO_SNDTIMEO` is per-call, not an overall response deadline.** A client that reads one byte per timeout period can hold a connection longer than the nominal bound. Bounded in practice by the existing model busy guard. (Configurable since the review round below, as `live_ingest.send_timeout_ms`.)
3. **Pre-existing, not introduced here, but now more reachable:** the listener uses `select()`/`FD_SET`, which is UB above `FD_SETSIZE`, and each client runs on a detached thread with no cap. The new code uses `poll()`/`WSAPoll` for its own waits. Happy to convert the listener in a separate PR if you'd like it, but I didn't want to bundle an unrelated change.
4. ~~**The four limits are hardcoded.**~~ **Fixed in review** — they are now a `live_ingest` config block, server-wide with per-model overrides. See below.
5. **`ENGINE_BUILD_TESTS=ON` is already broken on `main`.** Three `tests/moss_tts_local/` targets fail on constructor signature mismatches at `c0d1c9f` — unrelated to this change (those sources are untouched here, and the CMakeLists diff is purely additive). Flagging it because the new test lives under the same guard, so enabling tests in order to run it will surface those pre-existing failures.

## What was already found and fixed

I put this through several independent adversarial review rounds before opening it, at escalating effort, and they found real defects. Listing them so you can see what has already been looked at rather than spending your review rediscovering it:

- **Unbounded `sample_rate`/`channels`** fed `resolve_chunk_samples()`, whose product sizes a `vector::resize` reached *after* the model lock. Two `INT_MAX` values overflowed the multiply; smaller ones requested hundreds of GB while holding the lock. `std::stoi` also stopped at the first non-digit, so `16000junk` silently became `16000`.
- **`Transfer-Encoding` matched by substring**, so `notchunked` selected the incremental path. Fixing that to token matching then revealed a second route to the same place: repeated field lines overwrote each other, so `gzip` on one line and `chunked` on the next read as bare `chunked`.
- **The 600 s deadline was only checked before a socket read**, so a buffered tail — up to a whole 8 MiB chunk, many model iterations — could be consumed past it. Hence the 64 KiB windowing.
- **The 8 KiB chunk-header bound admitted ~16 KiB**; the length test ran only while still searching for CRLF.
- **Validation failures returned 500** instead of 400.
- **The documented `curl` command did not stream at all.** `--data-binary @-` drains stdin before connecting, turning a live capture into a file upload — the exact thing this endpoint exists to avoid. Measured against a timestamping listener, corrected to `-T -`, re-verified end to end.
- **Two test assertions passed for the wrong reason.** They asserted only that *an* error occurred, but these requests end by half-closing, which is itself an error — so they held even with the guard removed. Each now asserts its specific diagnostic.
- **No test crossed the 64 KiB compaction threshold**, so that branch never executed. The added case is mutation-verified: making compaction drop one unconsumed byte fails it, and it passes again on revert.

The chunked de-framing state machine was independently verified clean twice over — no byte skipped or repeated, exact-window-multiple chunks correct, compaction unable to invalidate the get area, bulk `read()` unable to bypass the deadline windows.

## Tests

`tests/unittests/test_http_live_body.cpp` drives real loopback sockets against `serve_http`, 13 cases:

- multi-chunk body arrives byte-identical and complete
- one 300 KB chunk, assembled across many `recv()` calls
- a body spanning several buffer compactions (the 64 KiB branch)
- 40 small chunks arriving several to a `recv()`
- chunk extensions ignored rather than parsed as size
- trailers accepted
- premature close → error, **not** a short body
- `ffffffffffffffff` overflow → rejected
- chunk over the per-chunk cap → rejected
- chunk data not CRLF-terminated → rejected
- non-hex size → rejected
- isolation: a chunked `POST /v1/audio/speech` still takes the buffered path, publishes no body stream, and keeps its existing (empty) buffered body

The compaction case is verified non-vacuous by mutation: making compaction drop one unconsumed byte fails it, and it passes again on revert.

Scope, stated plainly: this test links the transport, not the server runtime, so it covers framing/bounds/isolation and does **not** exercise the route itself or the SSE error conversion. Its header comment says so.

Existing `streaming_audio_input_test`, `citrinet_tokenizer_decode_test`, `sentencepiece_tokenizer1_test` pass unchanged. CUDA build clean.

---

## Review round (commit 2)

Both changes @0xShug0 asked for, in this PR rather than as follow-ups.

### 1. `live_ingest` is configurable, server → model

```json
{
  "live_ingest": { "total_timeout_ms": 600000, "max_chunk_bytes": 8388608 },
  "models": [
    { "id": "voxtral-realtime", "live_ingest": { "total_timeout_ms": 1800000 } }
  ]
}
```

A model entry overrides only the keys it names and inherits the rest, so raising one bound for a long-capture model doesn't mean restating the whole policy and letting it drift from the fleet default.

`send_timeout_ms` joined the set. It is the same class of bound — a client that stops reading holds the model open — and it was equally hardcoded, so leaving it out would have made the block look complete while one axis stayed fixed.

Values use `0` to mean *disabled*, matching `busy_timeout_ms` rather than inventing a second convention. **`max_chunk_bytes` is the exception and must stay positive**: a chunk is materialized in memory, so unbounded is not implementable, and the same value backs the chunk-size overflow check (see below). A negative value is rejected at startup naming the offending scope; silently reading it as "no bound" would remove a guard the operator believed they had set.

One design note worth flagging, since it touches the transport interface: the streambuf takes its bounds at construction, before `handle()` runs, and the transport has no notion of models. So `IHttpHandler` gains a virtual `live_ingest_limits(request)`, consulted once and only for a request that opts into an incremental body; `ServerState` resolves it from config. It deliberately does **not** reject an unknown or non-streaming model — that would turn a client's mistake into a dropped connection instead of the `400` `handle_transcription_live` already produces.

### 2. `busy_timeout_ms` query parameter

```
/v1/audio/transcriptions/live?model=voxtral-realtime&busy_timeout_ms=30000
```

A query parameter for the same reason the other parameters are: the body is audio and has nowhere to hold JSON. Semantics are unchanged from the other routes — clamped by the configured ceiling by the existing `resolve_busy_timeout_ms`, so a request can shorten its own wait but never weaken the guard. Malformed or out-of-range values return `400` with the specific reason.

### Three defects this change surfaced

Making the bounds settable made them reachable at values the compiled defaults never took, and an adversarial review pass found three real bugs there. Each is fixed with a mutation-verified test:

1. **The chunk-size guard underflowed for small caps.** `size > (cap - value) / 16` wraps whenever `cap < value`. At `max_chunk_bytes: 1` a declared `f` was accepted outright, and at `0` the guard never fired — re-admitting the `ffffffffffffffff` declaration whose `size + 2` wraps to 1, which is the exact bug this hand-rolled parse exists to prevent. Split into two checks, each on operands already in range.
2. **`max_body_bytes` missed a fully-prefetched body.** It was enforced only after a `recv()`, but a body short enough to arrive alongside the headers never re-enters that path — so the cap was skipped by exactly the request most likely to be probing it. Now also checked before the first byte is served.
3. **The config helpers accepted wrong-typed and truncating values.** `optional_i32`/`optional_i64` return the supplied fallback for a present-but-wrong-typed field, so in a model override `"max_body_bytes": "oops"` recorded the compiled 512 MiB default as a deliberate override and widened a stricter server policy. `optional_i32` also narrows before any range check, so `4294967296` became `0` (= disabled) on a 32-bit int. Both now parse and validate explicitly.

### Verification

`test_http_live_body.cpp` is now 18 cases. The new ones: a per-request cap takes effect where the default would allow it (chunk and body, separately); a cap below one hex digit still rejects; a zero cap falls back rather than disabling; a fully-prefetched body is still measured against the cap; and the limits hook is not consulted for a non-live route. Every one is mutation-verified — reverting the corresponding fix fails it, restoring passes.

The branch was rebased onto current `main`; `read_http_request` had gained `max_request_body_bytes` from #143 in the same place this branch splits off the incremental path. Resolved so the two bounds are complementary: the chunked early return precedes Content-Length handling, so a live body is bounded by `live_ingest.max_body_bytes` and a buffered one by `max_request_body_bytes`, with neither path escaping both.

Against a running server: a server-level `max_chunk_bytes` rejects a 64 KiB chunk the compiled default would accept, adding a per-model override to 1 MiB makes the identical request succeed, and negative values in either scope fail at startup with a message naming the scope. `busy_timeout_ms=abc`, `=-1` and a 20-digit value each return `400`; a real value streams normally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1220,6 +1220,7 @@ add_executable(audiocpp_server
     app/server/runtime.cpp
     app/cli/args.cpp
     app/cli/request.cpp
+    app/streaming/pcm_source.cpp
     app/streaming/streaming.cpp
 )
 
@@ -1561,6 +1562,24 @@ if (ENGINE_BUILD_TESTS)
         NAME streaming_audio_input_test
         COMMAND streaming_audio_input_test
     )
+
+    add_executable(http_live_body_test
+        tests/unittests/test_http_live_body.cpp
+        app/server/http.cpp
+    )
+    target_link_libraries(http_live_body_test PRIVATE engine_runtime ggml)
+    if(WIN32)
+        target_link_libraries(http_live_body_test PRIVATE ws2_32)
+    endif()
+    target_include_directories(http_live_body_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/tests/unittests)
+
+    add_test(
+        NAME http_live_body_test
+        COMMAND http_live_body_test
+    )
+    # This test talks to a real loopback socket, so a framing regression could
+    # block forever rather than fail. Bound it so CI reports a failure instead.
+    set_tests_properties(http_live_body_test PROPERTIES TIMEOUT 120)
 
     add_engine_unittest(hf_sampler_test tests/unittests/test_hf_sampler.cpp)
 
--- a/app/server/README.md
+++ b/app/server/README.md
@@ -102,7 +102,7 @@ The bound is resolved in three layers, since model runtimes differ by orders of 
 
 1. **Server** — top-level `"busy_timeout_ms"` (or `--busy-timeout-ms`) sets the fleet default.
 2. **Model** — `"busy_timeout_ms"` on an entry in `"models"` overrides that default for one model, and becomes the ceiling for requests to it.
-3. **Request** — `"busy_timeout_ms"` in the request body (or as a `busy_timeout_ms` form field on multipart transcription) lets a caller bound its own wait.
+3. **Request** — `"busy_timeout_ms"` in the request body (or as a `busy_timeout_ms` form field on multipart transcription, or a `busy_timeout_ms` query parameter on the live-ingest route, whose body is audio and has nowhere to put JSON) lets a caller bound its own wait.
 
 A request may ask for a **shorter** bound than the model's ceiling but never a longer one — `effective = min(request, ceiling)` — so a client cannot weaken the guard and reintroduce the hang it prevents. Because `0` means "unbounded", it compares as infinity on both sides: a request asking for `0` is still capped by the model ceiling, while under a ceiling of `0` a request's own bound is honored.
 
@@ -303,6 +303,7 @@ Because the body carries audio rather than JSON, parameters are query parameters
 | `channels` | `1` | interleaved channel count |
 | `sample_format` | `s16le` | `s16le` or `f32le` |
 | `language` | unset | passed through to the model |
+| `busy_timeout_ms` | model policy | how long to wait for the model lock, as elsewhere; clamped by the configured ceiling, so a request can shorten its own wait but never weaken the guard |
 
 ```bash
 # Microphone straight into transcription (macOS; -f alsa on Linux, -f dshow on Windows)
@@ -319,16 +320,35 @@ Whether partial text actually appears *during* capture is a property of the mode
 
 The request ends when the client sends the terminating chunk. Closing the connection without one is an error, not an end of speech — a truncated transcript that arrives as a normal `transcript.text.done` would be indistinguishable from the speaker stopping, so the endpoint refuses to produce one. The same applies to a stall past the idle timeout, an oversized chunk, or a malformed frame: each surfaces as an SSE `error` event.
 
-Because the model is held for the length of the request, the body is bounded on four axes, all compile-time constants today:
+Because the model is held for the length of the request, the body is bounded on several axes:
 
-| bound | value | meaning |
+| key | default | meaning |
 | --- | --- | --- |
-| idle | 30 s | longest wait for more data once the reader asks for it |
-| total | 600 s | checked at every point the body advances, so it caps the whole request |
-| size | 512 MiB | received body bytes, framing included |
-| chunk | 8 MiB | largest single declared chunk |
+| `idle_timeout_ms` | 30 s | longest wait for more data once the reader asks for it |
+| `total_timeout_ms` | 600 s | checked at every point the body advances, so it caps the whole request |
+| `max_body_bytes` | 512 MiB | received body bytes, framing included |
+| `max_chunk_bytes` | 8 MiB | largest single declared chunk |
+| `send_timeout_ms` | 30 s | `SO_SNDTIMEO` on the connection, so a client that stops reading the SSE response cannot hold the model open |
 
-The 600 s figure is an upper bound on one dictation; continuous captioning would need it raised. `sample_rate` and `channels` are range-checked too (1000–384000 and 1–16): they size the model's per-chunk buffer, so an absurd value would otherwise be an allocation request made while the model lock is held.
+Those defaults suit one dictation at a time. Continuous captioning needs a longer deadline, and a trusted deployment may want a different trade entirely, so they are configurable — server-wide under `live_ingest`, with any subset overridden per model:
+
+```json
+{
+  "live_ingest": { "total_timeout_ms": 600000, "max_chunk_bytes": 8388608 },
+  "models": [
+    {
+      "id": "voxtral-realtime",
+      "family": "voxtral_realtime",
+      "mode": "streaming",
+      "live_ingest": { "total_timeout_ms": 1800000 }
+    }
+  ]
+}
+```
+
+A model entry sets only the values it needs and inherits the rest, so raising one bound for a long-capture model does not mean restating the whole policy and letting it drift. `0` means *disabled*, the same convention as `busy_timeout_ms`; a negative value is rejected at startup rather than silently treated as "no bound". The exception is `max_chunk_bytes`, which must stay positive and is rejected at `0`: a chunk is materialized in memory before it is served, so an unbounded one is not implementable, and the same value backs the overflow check that rejects a declared chunk size of `SIZE_MAX`. The keys are only consulted for this route, since no other one delivers its body incrementally.
+
+`sample_rate` and `channels` are range-checked too (1000–384000 and 1–16), and are not configurable: they size the model's per-chunk buffer, so an absurd value would be an allocation request made while the model lock is held.
 
 **Deployment.** This endpoint streams the request body and the response on one connection at the same time. That is legal HTTP/1.1, but it needs a direct connection or a proxy that does not buffer requests. Behind nginx both of these are required, because `proxy_request_buffering off` still buffers a chunked body unless the upstream connection is HTTP/1.1:
 
--- a/app/server/README.md
+++ b/app/server/README.md
@@ -286,6 +286,60 @@ curl -N http://127.0.0.1:8080/v1/audio/transcriptions \
 
 The stream emits `transcript.text.delta` events, one final `transcript.text.done` event containing the full transcript, then `data: [DONE]`.
 
+Note that `stream=true` streams the *output* of an already-uploaded file: the whole recording is sent first, and the deltas describe decoding it. It shortens time-to-first-token on long audio, but nothing can appear while the speaker is still talking. For that, use the live endpoint below.
+
+### `POST /v1/audio/transcriptions/live`
+
+Streams raw PCM **as it is captured** and returns transcript deltas on the same connection, so partial text can appear while the user is still speaking.
+
+The request body is raw interleaved PCM sent with `Transfer-Encoding: chunked`; the response is the same SSE event shape as `stream=true` above, so a client can share one reader. There is no multipart form and no file — the audio never has to exist on disk, and the transport hands each chunk to the model as it arrives rather than assembling the recording first. Whether the *model* then keeps the whole utterance in memory is its own business: `nemotron_asr`, for instance, accumulates internally regardless of how the audio reaches it.
+
+Because the body carries audio rather than JSON, parameters are query parameters:
+
+| parameter | default | meaning |
+| --- | --- | --- |
+| `model` | required | id of a model configured with `mode: "streaming"` |
+| `sample_rate` | `16000` | samples per second of the PCM being sent |
+| `channels` | `1` | interleaved channel count |
+| `sample_format` | `s16le` | `s16le` or `f32le` |
+| `language` | unset | passed through to the model |
+
+```bash
+# Microphone straight into transcription (macOS; -f alsa on Linux, -f dshow on Windows)
+ffmpeg -f avfoundation -i ":0" -ar 16000 -ac 1 -f s16le - \
+  | curl -N -X POST -H 'Expect:' -T - \
+      'http://127.0.0.1:8080/v1/audio/transcriptions/live?model=voxtral-realtime&sample_rate=16000&channels=1&sample_format=s16le'
+```
+
+`-T -` is what makes this live, and it is not interchangeable with `--data-binary @-`: the latter drains stdin to completion before opening the connection, which turns a live capture back into a file upload and defeats the endpoint. `-H 'Expect:'` suppresses curl's `Expect: 100-continue`, which the server does not answer — without it curl waits out its one-second continue timeout before sending any audio. (`-T .` reads stdin non-blocking; measured against this endpoint it behaves the same, so either works.)
+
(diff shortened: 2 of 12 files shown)


5uck1ess:
small pr 😭

mirek190:
> small pr 😭

My had 7k   ;)
https://github.com/0xShug0/audio.cpp/pull/98

0xShug0:
@5uck1ess 

Sorry for the inconvenience😅. I think we can include these two improvements in this PR instead of doing follow-ups.:

1. Make the live-ingest limits optional configurable, with server-level defaults in code and optional per-model overrides
```
{
  "live_ingest": {
    "idle_timeout_ms": 30000,
    "total_timeout_ms": 600000,
    "max_body_bytes": 536870912,
    "max_chunk_bytes": 8388608,
    "send_timeout_ms": 30000
  }
}
```

A model entry should be able to override only the values it needs:

```
{
  "id": "voxtral-realtime",
  "family": "voxtral_realtime",
  "mode": "streaming",
  "live_ingest": {
    "total_timeout_ms": 1800000
  }
}
```

2. Add request-level busy_timeout_ms as a query parameter for /v1/audio/transcriptions/live, consistent with the other server paths that already support per-request busy-timeout overrides.

`/v1/audio/transcriptions/live?model=voxtral-realtime&busy_timeout_ms=30000`

5uck1ess:
No inconvenience at all — both are in, pushed as a second commit (`fdf39c4`).

**1. `live_ingest`, server → model.** Exactly the shape you specified, and a model entry overrides only the keys it names:

```json
{
  "live_ingest": { "total_timeout_ms": 600000, "max_chunk_bytes": 8388608 },
  "models": [
    { "id": "voxtral-realtime", "live_ingest": { "total_timeout_ms": 1800000 } }
  ]
}
```

I added `send_timeout_ms` to the block as you listed it. It's the same class of bound — a client that stops reading holds the model open — and it was equally hardcoded, so leaving it out would have made the block look complete while one axis stayed fixed.

Two decisions I'd rather surface than have you find:

- **`0` means disabled**, on every value, matching `busy_timeout_ms` rather than inventing a second convention. A negative value is rejected at startup naming the offending scope — silently reading it as "no bound" would remove a guard the operator believed they'd set.
- **`IHttpHandler` gains a virtual `live_ingest_limits(request)`.** The streambuf takes its bounds at construction, before `handle()` runs, and the transport has no notion of models — so the transport asks and `ServerState` resolves. It's consulted once, only for a request that opts into an incremental body, and it deliberately does *not* reject an unknown or non-streaming model: doing so there would turn a client's mistake into a dropped connection instead of the `400` the handler already returns. If you'd rather the transport not grow an interface point, the alternative is resolving limits from the path+query inside `http.cpp`, which means teaching the transport about model ids — happy to switch if you prefer that trade.

**2. `busy_timeout_ms` query parameter**, as you asked:

```
/v1/audio/transcriptions/live?model=voxtral-realtime&busy_timeout_ms=30000
```

Same semantics as the other routes — it goes through the existing `resolve_busy_timeout_ms`, so it's clamped by the configured ceiling and a request can shorten its own wait but never weaken the guard. Malformed and out-of-range values return `400` with the reason.

**Verification.** Three new cases in `test_http_live_body.cpp` prove the bounds are per-request rather than compiled in — the same 2000-byte chunk is accepted under the defaults and rejected once the handler tightens `max_chunk_bytes`, the same for the whole-body cap, and the limits aren't resolved at all for a non-live route. Mutation-verified: pinning the transport back to default limits fails them, reverting passes. Against a running server, a server-level `max_chunk_bytes` rejects a 64 KiB chunk the default accepts, a per-model override to 1 MiB makes the identical request succeed, and bad values in either scope fail at startup naming the scope.

README and `streaming_example.json` updated; the "limits are hardcoded" caveat in the description is struck through.

5uck1ess:
Rebased onto current `main` — the PR is mergeable again. `read_http_request` had gained `max_request_body_bytes` from #143 in the same place this branch splits off the incremental path; resolved so the two bounds are complementary rather than one shadowing the other. The chunked early return happens *before* Content-Length handling, so a live body is bounded by `live_ingest.max_body_bytes` and a buffered one by `max_request_body_bytes`, with neither path escaping both.

I then put the result through an adversarial review pass, and it found three real defects — all of them reachable only *because* the limits became configurable, i.e. introduced by the change you asked for. Fixed here, each with a mutation-verified test:

1. **Chunk-size guard underflowed for small caps.** `size > (cap - value) / 16` wraps whenever `cap < value`. At `max_chunk_bytes: 1` a declared `f` was accepted outright — a 15-byte chunk against a 1-byte bound — and at `0` the guard never fired at all, re-admitting the `ffffffffffffffff` declaration whose `size + 2` wraps to 1. That is precisely the bug this parse is hand-rolled to prevent, so the configurable cap had quietly reopened it. Now two checks, each on operands already known to be in range.

2. **`max_body_bytes` missed fully-prefetched bodies.** It was only enforced after a `recv()`, but a body short enough to arrive alongside the headers never re-enters that path — so the cap was skipped by exactly the request most likely to be probing it. Now also checked before the first byte is served.

3. **Config helpers accepted wrong-typed and truncating values.** `optional_i32`/`optional_i64` return the supplied fallback for a present-but-wrong-typed field, which in a *model override* meant `"max_body_bytes": "oops"` recorded the compiled 512 MiB default as a deliberate override — silently widening a stricter server policy. `optional_i32` also narrows before any range check, so `4294967296` became `0` (= disabled) on a 32-bit int. Both now parse and validate explicitly, with `2^53 - 1` and platform-`size_t` ceilings.

Consequence for the spec you gave: **`max_chunk_bytes` cannot be `0`.** A chunk is materialized in memory, so unbounded is not implementable, and the same value backs the overflow check above. It is rejected at startup with that reason rather than quietly substituted. Every other bound keeps `0` = disabled.

Also fixed a rebase artifact the same pass caught: the first commit's test called the pre-#143 five-argument `serve_http`, so that commit alone would not build with `ENGINE_BUILD_TESTS=ON`. Folded into the commit that introduces the test, so both commits build standalone.

Test file is now 18 cases. New ones: a per-request cap actually takes effect where the default would allow it (chunk and body, separately), a cap below one hex digit still rejects, a zero cap falls back rather than disabling, a fully-prefetched body is still measured, and the limits hook is not consulted for non-live routes. Each mutation-verified — reverting the corresponding fix fails it.

5uck1ess:
@0xShug0 should be good to go.

0xShug0:
@5uck1ess Tested and it looks good to me. Merged and thank you!

PR #149, opened 2026-07-29
metal: dispatch conv_transpose_1d with 256-thread threadgroups (~5.6x AudioVAE decode)

liuzl:
## What

`kernel_conv_transpose_1d` was dispatched as `OL x OC` threadgroups of a **single thread** each, so 31/32 SIMD lanes of every warp were idle. This PR keeps the kernel's math and loop order unchanged and only regroups the launch: 256 output positions per threadgroup, with a bounds check against a new `args.OL`; the kernel now reads `OC` from kargs instead of `threadgroups_per_grid`.

Diff is +24/-8 across the three Metal backend files (kargs struct, host dispatch, kernel + template instantiations).

## Why it matters

`conv_transpose_1d` dominates AudioVAE-style waveform decoders (VoxCPM2 among them). On Apple Silicon the one-thread-per-threadgroup launch makes waveform decode the single largest cost of a TTS request.

## Measurements

M3 Max (Metal build, `GGML_METAL_EMBED_LIBRARY=OFF`), VoxCPM2-2B `q8_0` from `audio-cpp/audio.cpp-gguf`, voice clone, fixed `--seed 1234`, 15.5 s of 48 kHz output, warm session (`--log` timing):

| metric | before | after |
|---|---:|---:|
| `voxcpm2.audiovae_decoder_ms` | 10 234 | 1 806–1 871 (**~5.6x**) |
| `session.wall_ms` | 22 579 | 14 144–14 611 |

## Correctness

- Same seed, same command: output WAV is **bit-identical** to the previous kernel (`max |diff| = 0` over 744 960 samples). Per-output-element summation order is unchanged, so this is expected, not tolerance-based.
- Streaming mode (`--mode streaming`, chunked AudioVAE decode) runs through the same kernel: regression-tested OK.
- `qwen3_asr` transcription regression-tested OK (unchanged output).

Commands used are reproducible from the table above; happy to add more shapes or an f16-weight run if useful.

## Disclosure

Developed with AI assistance (Claude Code); the same fix has been validated for weeks in a downstream VoxCPM.cpp fork where it was originally written against upstream ggml.

### DIFF
--- a/external/ggml/src/ggml-metal/ggml-metal-impl.h
+++ b/external/ggml/src/ggml-metal/ggml-metal-impl.h
@@ -599,6 +599,8 @@ typedef struct {
     int32_t  IL;
     int32_t  K;
     int32_t  s0;
+    int32_t  OL;
+    int32_t  OC;
     uint64_t nb0;
     uint64_t nb1;
 } ggml_metal_kargs_conv_transpose_1d;
--- a/external/ggml/src/ggml-metal/ggml-metal-ops.cpp
+++ b/external/ggml/src/ggml-metal/ggml-metal-ops.cpp
@@ -4053,6 +4053,8 @@ int ggml_metal_op_conv_transpose_1d(ggml_metal_op_t ctx, int idx) {
         /*.IL  =*/ IL,
         /*.K   =*/ K,
         /*.s0  =*/ s0,
+        /*.OL  =*/ OL,
+        /*.OC  =*/ OC,
         /*.nb0 =*/ nb0,
         /*.nb1 =*/ nb1,
     };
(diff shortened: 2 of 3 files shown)


0xShug0:
@liuzl Thank you! The performance gain is huge, and the change will benefit many models! Looking foward to more Metal optimizations!

case | WAV parity | baseline | PR | delta
-- | -- | -- | -- | --
voxcpm2_streaming_voice_design | exact match | 16426.344 ms | 8320.115 ms | -49.35%
voxcpm2_tts_voice_design | exact match | 15840.184 ms | 6192.202 ms | -60.91%
voxcpm2_voice_clone | exact match | 18545.740 ms | 8781.549 ms | -52.65%

PR #154, opened 2026-07-29
Fuse compatible Q8 CUDA residual graph patterns

mirek190:
## Summary

This adds a reusable CUDA optimization for compatible `Q8_0` decoder graph patterns:

```text
MUL_MAT(Q8_0) -> RESHAPE (metadata only) -> ADD (residual)
```

The optimized path writes the final residual result directly from a dedicated Q8 MMVQ specialization, skipping the standalone residual-add CUDA kernel.

## Root cause

The CUDA backend already supports fused matvec bias/add work, but the generic fusion matcher expects adjacent compute nodes. Decoder graphs can place a metadata-only `RESHAPE` between `MUL_MAT` and the residual `ADD`, so the existing matcher does not recognize the otherwise compatible chain.

## Implementation

- Add a strict Q8-only matcher for `MUL_MAT -> RESHAPE -> ADD`.
- Require the reshape to consume the matvec directly.
- Require single-use intermediates, matching element counts, non-output intermediates, and safe memory ranges.
- Add a residual-only Q8 MMVQ template specialization.
- Remove unused GLU/gate calculation and shared-memory state from that specialization.
- Preserve the existing generic fusion paths for all other patterns and types.

The matcher is model-independent, so any compatible Q8 graph can benefit.

## Validation

Tested on:

- NVIDIA GeForce RTX 3090, 24 GiB
- CUDA 12.4
- compute capability 8.6 / `CMAKE_CUDA_ARCHITECTURES=86-real`
- Windows, MSVC 14.43, Ninja, Release
- CUDA graphs enabled
- Higgs Audio v3 TTS Q8 GGUF as the representative compatible decoder

Build command:

```powershell
cmake --build build/windows-cuda-higgs-perf-v124 --config Release --target higgs_audio_tts_warm_bench -j 8
```

The complete CUDA backend rebuilt successfully. Existing compiler warnings were unchanged and non-fatal.

### Parity

Five fixed request cases were compared with equal iteration counts:

- identical output frame counts for every request
- WAV cosine: `1.00000000` for every request
- log-mel cosine: `1.00000000` for every request

A final post-commit parity run produced the same results.

### Performance

Three iterations per request, comparing the unfused and specialized paths in the same binary:

| Request | Unfused | Specialized | Speedup |
|---|---:|---:|---:|
| 0 | 972.87 ms | 954.82 ms | 1.019x |
| 1 | 922.51 ms | 905.93 ms | 1.018x |
| 2 | 1408.51 ms | 1384.46 ms | 1.017x |
| 3 | 512.97 ms | 501.44 ms | 1.023x |
| 4 | 936.02 ms | 917.65 ms | 1.020x |

Mean request wall time:

```text
unfused:     950.57 ms
specialized: 932.86 ms
reduction:     1.86%
speedup:       1.019x
```

The residual-only specialization is also about `0.56%` faster in aggregate than routing the same graph pattern through the generic fused MMVQ kernel.

## Scope and limitations

- Only `Q8_0` matrix-vector decode patterns matching the validated chain activate this specialization.
- Prefill/multi-column paths and incompatible layouts keep their existing execution paths.
- No model-specific source code is changed.

### DIFF
--- a/external/ggml/src/ggml-cuda/common.cuh
+++ b/external/ggml/src/ggml-cuda/common.cuh
@@ -1518,10 +1518,12 @@ struct ggml_cuda_mm_fusion_args_host {
     const ggml_tensor * gate = nullptr;
     const ggml_tensor * gate_bias = nullptr;
     ggml_glu_op glu_op;
+    bool residual_only = false;
 };
 struct ggml_cuda_mm_fusion_args_device {
     const void * x_bias = nullptr;
     const void * gate = nullptr;
     const void * gate_bias = nullptr;
     ggml_glu_op glu_op;
+    bool residual_only = false;
 };
--- a/external/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/external/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -3771,6 +3771,43 @@ static bool ggml_cuda_check_fusion_memory_ranges(const ggml_cgraph * cgraph,
     return is_ok;
 }
 
+// Some model graphs reshape a matvec result before adding the residual. RESHAPE
+// is metadata-only and therefore cannot pass the generic compute-node fusion
+// validator. Validate this exact chain explicitly so the residual-only Q8_0
+// specialization can write the final result directly.
+static bool ggml_cuda_can_fuse_q8_0_mul_mat_reshape_add(
+        const struct ggml_cgraph * cgraph, int node_idx) {
+    if (node_idx + 2 >= cgraph->n_nodes) {
+        return false;
+    }
+
+    const ggml_tensor * mul_mat = cgraph->nodes[node_idx + 0];
+    const ggml_tensor * reshape = cgraph->nodes[node_idx + 1];
+    const ggml_tensor * add     = cgraph->nodes[node_idx + 2];
+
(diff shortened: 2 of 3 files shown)


mirek190:
[0xShug0](https://github.com/0xShug0)

What is with vulkan for windows / linux ? 
Are you working on that?

0xShug0:
@mirek190 afk..back in ~6 hours and will check. I'm not working on Vulkan optimization, but the WIP PR may be relevent.

0xShug0:
@mirek190 The changes are scoped and safe. It's good to keep pushing the performance boundaries!

Graph shape is like decoder token with hidden size 4096. Q8_0 matmul -> reshape -> add residual 

Main:
samples: 100
iterations_per_sample: 80
max_abs_diff: 0.000002
mean_abs_diff: 0.000000
mean_ms: 0.013063
median_ms: 0.011655
min_ms: 0.011183
max_ms: 0.023231

PR:
samples: 100
iterations_per_sample: 80
max_abs_diff: 0.000002
mean_abs_diff: 0.000000
mean_ms: 0.011035
median_ms: 0.010399
min_ms: 0.010085
max_ms: 0.028383

PR #156, opened 2026-07-30
fix: make Linux build script executable

LauraGPT:
## Summary

Make scripts/build_linux.sh executable so the direct commands documented in the README work after a normal clone.

The script contents are unchanged; this only changes its Git mode from 100644 to 100755.

## Verification

- Reproduced the current failure before the change: the documented direct invocation exits with Permission denied.
- Confirmed the script blob is unchanged: 88e5ca0717f734e24b4e5c236646b84c59b876b6 before and after.
- Ran the documented direct Fun-ASR-Nano CPU build successfully:
  scripts/build_linux.sh --backend cpu --model-set custom --models fun_asr_nano --target audiocpp_cli --target audiocpp_server
- Confirmed both audiocpp_cli and audiocpp_server were produced.
- git diff --check passes.


0xShug0:
Good catch. PR merged!

PR #163, opened 2026-07-30
feat(nix): support selecting model targets in Nix builds

francescobozzo:
## Changes

 - package.nix: Added models parameter: non-empty list auto-sets `AUDIOCPP_MODEL_SET=custom`, empty defaults to full
 - CMakeLists.txt: `EXCLUDE_FROM_ALL` on all model OBJECT libraries so unselected models aren't compiled; `vibevoice_finetune_overlay_check` now gated on vibevoice being linked. Note: this change influences all other build pipelines, but was necessary to exclude unwanted models not just from linking but also from compile time.
 - docs/build/nixos.md: New Model Selection section with CLI and NixOS examples

## Tests

Validated model list reduces the number of building steps in cmake with the following configuration:

```
vulkan = base.override {
  vulkanSupport = true;
  models = [
    "vibevoice"
    "miotts"
  ];
};
```

### DIFF
--- a/.devops/nix/package.nix
+++ b/.devops/nix/package.nix
@@ -11,7 +11,6 @@
   vulkan-tools,
   glslang,
   shaderc,
-  darwin,
   python-scripts,
   config,
   version,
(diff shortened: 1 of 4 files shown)


0xShug0:
@francescobozzo Merged! Thanks!

Issue #105, opened 2026-07-25
Qwen3-TTS-1.7B Base voice cloning becomes distorted mid-generation


cinetube:
I'm seeing an issue similar to #67 when using **Qwen3-TTS-1.7B Base** for voice cloning.

The cloned voice starts out sounding correct, but somewhere in the middle of the generated audio it becomes noticeably distorted. The beginning of the output is fine, but the quality degrades partway through the generation.

What makes me think this is **not** a model issue is that I have another voice cloning setup using the **same Qwen3-TTS-1.7B Base model**, and voice cloning works correctly there without any distortion. Because of that, I suspect the issue is somewhere in the `audio.cpp` inference path rather than with the model itself.

### Expected behavior

The cloned voice should remain consistent throughout the entire generated audio.

### Actual behavior

* Voice cloning starts normally.
* Midway through the generated audio, the voice becomes distorted.
* The distortion persists for the remainder of the output.

### Additional information

* Model: Qwen3-TTS-1.7B Base
* Task: Voice cloning
* I can reproduce the issue consistently.
* I have verified that the same model works correctly in another voice cloning implementation, so this does not appear to be an issue with the model weights themselves.

If there's any additional logging, debug output, or test case I can provide, I'd be happy to help investigate further.

0xShug0:
@cinetube Please share the reference voice and the exact cmd you used so I can reproduce the issue. 

Update --- also the model you used: official safetensor or GGUFs from the audio.cpp HF repo.

0xShug0:
Just a quick update: I can reproduce the issue and reducing the text chunk size makes the distortion go away. 

The current default parameters for Qwen3-TTS are bad: both `max_tokens` and the `text chunk size `are set to 8192, ~~which forces generation onto a streaming path after roughly 20 seconds~~. One gap I noticed is that the UI does not appear to expose a text chunk size option.

Still digging. 

**Update 1**: the streaming path is not the root cause. Text chunking is just a workaround.

The drift appears in longer text:

Windowed WAV cosine between cpp and python wavs:
0-5s: 0.999802
5-10s: 0.999836
10-15s: 0.999773
15-20s: 0.735259
20-25s: 0.152537
25-28.7s: 0.037134

**Update 2**: Interesting, Q8 + perf on mode has no distortion issue.

0xShug0:
[...]

PR #170, opened 2026-07-31
Optimize Qwen3-TTS ROCm inference on Strix Halo

fedeizzo:
## What does this PR do?

- `.devops/nix/package.nix`, `flake.nix`, `CMakeLists.txt`: add an opt-in `rocm-gfx1151` build. It enables the Strix Halo path only for HIP with exactly `gfx1151`, and keeps rocBLAS because hipBLASLt regressed latency on this hardware.
- `include/engine/framework/core/module.h`, `src/framework/modules/{conv_modules.cpp,optimizations/fast_projection_modules.cpp}`: allow HIP to use the verified GGML CUDA-family ConvTranspose1d and packed-projection paths used by Qwen3-TTS. Other backends keep their existing behavior.
- `src/models/qwen3_tts/assets.cpp`: validate transformer dimensions, model configuration, and numeric parameters before loading invalid graphs.
- `src/models/qwen3_tts/session.cpp`, `src/models/qwen3_tts/talker.cpp`: validate generation options and voice-clone inputs, support speaker-embedding-only mode, make `mem_saver` cleanup exception-safe, and remove redundant backend synchronizations.
- `src/models/qwen3_tts/tokenizer_speech_decoder.cpp`: retain the common 300- and 105-frame decoder graphs, restore every graph input before reuse, safely replace fallback shapes, and validate reference-audio trimming.
- `tests/unittests/*qwen*`, `tests/unittests/test_conv_transpose_fast_path.cpp`: cover configuration validation and HIP fast-path selection while confirming unsupported backends are rejected.

All Strix-specific behavior is disabled by default. Generic CPU, CUDA, Vulkan, Metal, and HIP builds retain their current defaults.

## Motivation

Qwen3-TTS repeatedly rebuilt stable decoder graphs on Strix Halo and showed a 1-second/2-second decoder compute mode. Reusing those graphs removes graph construction and stabilizes decoder execution, but reuse is only correct when codes, positions, and attention masks are restored before every invocation.

The implementation is restricted to the tested `gfx1151` HIP configuration so hardware-specific tuning does not affect generic builds. hipBLASLt and HIP graph capture are not enabled because they were slower in testing.

## Gains

Correctness-matched Ryzen AI Max / Strix Halo results with HIP graphs disabled:

| Metric | Before | After | Gain |
| --- | ---: | ---: | ---: |
| Mean synthesis latency | 10.400 s | 9.454 s | **9.1%** |
| Median synthesis latency | 10.499 s | 9.354 s | **10.9%** |
| Decoder latency | 2.369 s | 1.304 s | **45.0%** |
| Decoder graph build | 140 ms | 0 ms | **100%** |

All 29 cold, warmup, and measured calls produced the same audio digest and sample metadata.

## How to use it

Use the specialized Nix development shell:

```bash
nix develop .#rocm-gfx1151
```

Or configure manually:

```bash
cmake -S . -B build/rocm-gfx1151 -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DENGINE_ENABLE_HIP=ON \
  -DGPU_TARGETS=gfx1151 \
  -DENGINE_HIP_STRIX_HALO_OPTIMIZATIONS=ON
cmake --build build/rocm-gfx1151 --parallel
```

Run `audiocpp_cli` normally with the HIP backend. The decoder cache activates automatically for the specialized build when the runtime backend is HIP.

For generic ROCm behavior, use `nix develop .#rocm` or leave `ENGINE_HIP_STRIX_HALO_OPTIMIZATIONS=OFF`.

## AI usage disclosure

AI was used for the following purposes:

- understand project structure and validate hypothesis about code organization
- help defining benchmarking scripts (not included in the PR)
- clean-up code after manual definition

AI was NOT used for the following purposes:

- optimization ideas
- basic implementation
- initial exploration of similar projects like llama-cpp and [CrispASR](https://github.com/CrispStrobe/CrispASR)

### DIFF
--- a/.devops/nix/package.nix
+++ b/.devops/nix/package.nix
@@ -96,11 +96,11 @@ stdenv.mkDerivation (finalAttrs: {
     # Copy the built C++ executables directly from the bin directory
     cp bin/audiocpp_cli bin/audiocpp_server bin/audiocpp_gguf $out/bin/
 
-    # Install the python model manager script
-    cp $src/tools/model_manager.py $out/bin/audiocpp_model_manager
-    chmod +x $out/bin/audiocpp_model_manager
+    # Install the supported spec-backed model manager and the catalog it reads
+    # relative to its installed location.
+    install -Dm755 $src/tools/model_manager_v2.py $out/bin/audiocpp_model_manager
+    cp -R $src/model_specs $out/model_specs
 
-    # Patch the shebang to use our python environment with torch/safetensors/pyyaml
     patchShebangs $out/bin/audiocpp_model_manager
 
     runHook postInstall
--- a/.devops/nix/package.nix
+++ b/.devops/nix/package.nix
@@ -21,10 +21,13 @@
   metalSupport ? stdenv.isDarwin,
   rocmSupport ? config.rocmSupport or false,
   rocmGpuTargets ? (lib.optionals rocmSupport rocmPackages.clr.gpuTargets),
+  strixHaloOptimizations ? false,
   # Model selection: if non-empty, only these model targets are built.
   # See CMakeLists.txt AUDIOCPP_MODEL_SET / AUDIOCPP_MODELS.
   models ? [ ],
 }:
+
+assert !strixHaloOptimizations || rocmSupport;
 
 stdenv.mkDerivation (finalAttrs: {
   pname = "audio.cpp";
@@ -78,7 +81,8 @@ stdenv.mkDerivation (finalAttrs: {
   ++ lib.optional metalSupport "-DENGINE_ENABLE_METAL=ON"
   ++ lib.optional rocmSupport "-DENGINE_ENABLE_HIP=ON"
   ++ lib.optional rocmSupport "-DCMAKE_HIP_COMPILER=${rocmPackages.llvm.clang}/bin/clang"
-  ++ lib.optional rocmSupport "-DGPU_TARGETS=${lib.concatStringsSep ";" rocmGpuTargets}";
+  ++ lib.optional rocmSupport "-DGPU_TARGETS=${lib.concatStringsSep ";" rocmGpuTargets}"
+  ++ lib.optional strixHaloOptimizations "-DENGINE_HIP_STRIX_HALO_OPTIMIZATIONS=ON";
 
   env = lib.optionalAttrs rocmSupport {
     ROCM_PATH = "${rocmPackages.clr}";
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,9 @@ option(ENGINE_ENABLE_VULKAN "Build ggml with Vulkan backend support" ${ENGINE_DE
 option(ENGINE_ENABLE_METAL "Build ggml with Metal backend support" ${ENGINE_DEFAULT_ENABLE_METAL})
 option(ENGINE_ENABLE_LLAMAFILE "Build ggml with llamafile SGEMM support" ${ENGINE_DEFAULT_ENABLE_LLAMAFILE})
 option(ENGINE_ENABLE_CUDA_GRAPHS "Enable ggml CUDA graphs support" ${ENGINE_DEFAULT_ENABLE_CUDA_GRAPHS})
+option(ENGINE_HIP_STRIX_HALO_OPTIMIZATIONS
+    "Enable opt-in HIP optimizations validated only for Strix Halo (gfx1151)"
+    OFF)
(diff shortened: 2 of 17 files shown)


.devops/nix/package.nix
```diff
@@ -21,11 +21,14 @@
   metalSupport ? stdenv.isDarwin,
   rocmSupport ? config.rocmSupport or false,
   rocmGpuTargets ? (lib.optionals rocmSupport rocmPackages.clr.gpuTargets),
+  strixHaloOptimizations ? false,
```
francescobozzo, line 24: Can we infer this directly if `gfx1151` is set as target?

.devops/nix/package.nix
```diff
@@ -92,11 +96,11 @@ stdenv.mkDerivation (finalAttrs: {
     # Copy the built C++ executables directly from the bin directory
     cp bin/audiocpp_cli bin/audiocpp_server bin/audiocpp_gguf $out/bin/
 
-    # Install the python model manager script
-    cp $src/tools/model_manager.py $out/bin/audiocpp_model_manager
-    chmod +x $out/bin/audiocpp_model_manager
+    # Install the supported spec-backed model manager and the catalog it reads
+    # relative to its installed location.
+    install -Dm755 $src/tools/model_manager_v2.py $out/bin/audiocpp_model_manager
```
francescobozzo, line 101: I think we should start considering the idea of adding at least a basic nix build in the CI step. This file rename caused the nix build to fail. We can have a separate PR for this. Any thoughts @0xShug0 ?

0xShug0: @francescobozzo  Sounds good!

.devops/nix/package.nix
fedeizzo: good point, can you only confirm that `gfx1151` points only to strix halo?

0xShug0:
@fedeizzo Thank you for the PR! Just a quick skim with a few comments:

Could you please rebase and split this into three PRs?

1. Nix packaging fix. This is small and should be quick to merge.
2. ROCm/Strix Halo optimization.
3. Qwen3-TTS behavior/validation changes.

The ROCm optimization, Qwen3-TTS validation/behavior changes, and Nix packaging fix have different review surfaces, so splitting them will make review and validation much easier.

For 2: The framework HIP fast-path expansion looks additive, but it is broader than Strix Halo: `FastPackedProjection4Module` and ~~ConvTranspose1d~~ now accept HIP as a CUDA-family backend. Could you validate a few more models that use these modules on HIP? I have seen cases where layout, view, or materialization behavior differs across backends.  A module may work in one model but still fail in another path on a non-CUDA backend purely because of materialization/layout differences. CUDA tends to be more tolerant here because these ggml fast paths are usually developed and exercised on CUDA first.

For 3: Qwen3-TTS originally reused this decoder graph, but not in a clean way. It was later changed to rebuild per request because of #105. Please make sure this PR does not reintroduce the same class of issue. This can be hard to catch with short requests, so it would be useful to validate longer or mixed-shape requests, not only short samples.

We are currently moving models toward spec v1; Confucius4-TTS is a good example of the intended shape. In that design, common request/session options such as `top_k` should be validated by the framework/spec layer rather than duplicated inside each model.

I’d suggest leaving asset and session option parsing as-is in this PR. Some of this defensive code may become obsolete after the spec-v1 migration, and keeping it here makes the PR harder to review.

.devops/nix/package.nix
francescobozzo: Yes: https://en.wikipedia.org/wiki/RDNA_(microarchitecture)#Comparison_of_RDNA_chips

Perhaps other architectures might benefit from these optimizations, but I think these should be tested and eventually enabled automatically

Actually, I think we should moving this to some macros, like llama-cpp is doing here: https://github.com/ggml-org/llama.cpp/blob/876a4321163249c43ca4e986818fab5ab081f282/ggml/src/ggml-cuda/vendors/hip.h#L219-L221

0xShug0:
@fedeizzo Are you still planning to continue working on this PR? Please update the PR when you get a chance...

fedeizzo:
> @fedeizzo Are you still planning to continue working on this PR? Please update the PR when you get a chance...

@0xShug0 Yes, sorry. I didn't have time to look into that in the past few days, my bad. I'll try to wrap up separated PRs in the next few days, at most this weekend 🙇

fedeizzo:
Closed in favor of 

https://github.com/0xShug0/audio.cpp/pull/202

and 

https://github.com/0xShug0/audio.cpp/pull/201

PR #172, opened 2026-08-01
Add nix CI workflow and fix package.nix build

francescobozzo:
## Changes

- New .github/workflows/nix-build.yml builds .#cpu and .#vulkan via cachix/install-nix-action@v31
 - package.nix: install spec-backed model_manager_v2.py + model_specs dir (fixes broken build). Courtesy of @fedeizzo from #170 

Note: I didn't include intentionally as the ROCm dependencies are huge to pull within Nix. We might have to leverage some caching for that. Anyway cpu and vulkan builds only reflect the matrixes of other workflows and is a good starting point.

## Tests

Verified locally with [act](https://github.com/nektos/act) the new Nix CI is working as expected.

```
nix run nixpkgs#act -- -W .github/workflows/nix-build.yml
```

### DIFF
--- a/.devops/nix/package.nix
+++ b/.devops/nix/package.nix
@@ -92,9 +92,10 @@ stdenv.mkDerivation (finalAttrs: {
     # Copy the built C++ executables directly from the bin directory
     cp bin/audiocpp_cli bin/audiocpp_server bin/audiocpp_gguf $out/bin/
 
-    # Install the python model manager script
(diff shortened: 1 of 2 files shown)


0xShug0:
Thank you @francescobozzo, merged!

0xShug0:
@francescobozzo The Nix Vulkan workflow frequently fails and has to be restarted. It looks like there may be a race condition. Is it possible to fix this issue?

PR #185, opened 2026-08-04
WebUI - Initial native Svelte WebUI embedded inside audiocpp_server.exe — no Python, Node.js, or Gradio needed

mirek190:
## Summary
Screenshot 2026-08-04 082608

This introduces the initial native SvelteKit/TypeScript WebUI and embeds its complete static output inside `audiocpp_server.exe`.

Normal browser inference now needs only the server executable plus the selected CPU/GPU runtime libraries. Python, Node.js, and Gradio are not runtime dependencies:

```text
Browser
   |
   | HTTP on localhost
   v
audiocpp_server.exe
   |-- compiled-in Svelte HTML/CSS/JavaScript
   |-- native /v1 server APIs
   `-- audio.cpp model sessions
          |-- CPU
          `-- CUDA / other configured backends
```

Node.js is used only when a developer rebuilds the frontend artifact. Python is used only when the user explicitly asks the optional model installer to download or convert a package; loading a prepared model and running inference remain native. The existing Gradio UI is retained as a compatibility option but is not used by this UI.

## What is included

- Build `webui/native` as a single static document and convert it to a generated C++ byte array during CMake configuration.
- Serve the compiled asset at `/` and `/index.html`, with cache, CSP, and content-type safety headers.
- Add server switches `--ui`, `--no-ui`, and `--ui-management`; `--ui` can start a model-management host without an external server JSON file.
- Add guarded native endpoints for dynamic model loading/unloading, path checks, temporary uploads, and asynchronous package-install status.
- Cover the primary audio.cpp workflows:
  - text to speech and voice cloning
  - ASR/transcription and near-live ASR
  - music generation
  - voice conversion
  - source separation, including HTDemucs, BS-RoFormer, and Mel-Band RoFormer presets
  - audio analysis
  - voice design
- Add long-text splitting/merge, microphone capture, multiple-output playback, timing, model catalog/settings, and a browser-local IndexedDB voice library.
- Route package-spec installs through `model_manager_v2.py`; keep converter-input packages on the deprecated helper until those packages migrate.
- Saved voices feature: lets the user name and keep a reference voice recording (audio + transcript) for reuse across voice-clone sessions. Voices are persisted ONLY in this browser's IndexedDB.
- Extend portable Windows packaging with the optional model-management resources.
- Document native and legacy WebUI setup and add focused configuration/model-installer tests.

## Build and test commands

Frontend:

```powershell
cd webui/native
npm ci
npm run check
npm run build
```

CPU server and tests:

```powershell
cmake --build build/native-ui-cpu --config Release --target audiocpp_server server_config_test server_model_installer_test --parallel 8
ctest --test-dir build/native-ui-cpu -C Release --output-on-failure -R "server_(config|model_installer)_test"
```

CUDA server (RTX 3090 / compute capability 8.6, CUDA 12.4):

```powershell
.\scripts\build_windows.ps1 -Preset windows-cuda-release -Target audiocpp_server -CudaArchitectures 86 -Jobs 8
```

Run the native UI:

```powershell
.\build\windows-cuda-release\bin\audiocpp_server.exe --ui --backend cuda --host 127.0.0.1 --port 18991
```

Then open `http://127.0.0.1:18991/`.

## Results

- `svelte-check`: 0 errors, 0 warnings.
- Frontend static build: passed.
- CPU `audiocpp_server`: built successfully against current `main` (`238ab6a`).
- `server_config_test`: passed.
- `server_model_installer_test`: passed.
- CUDA `audiocpp_server`: built successfully with CUDA 12.4 for SM 8.6.
- CUDA smoke test:
  - `GET /health`: `status=ok`, `backend=cuda`, `ui=true`, `ui_management=true`
  - `GET /`: HTTP 200, 167,097 bytes
  - embedded page contained the current BS-RoFormer catalog entry

Generated artifacts:

- `webui/native/dist/index.html`
- `build/native-ui-cpu/bin/Release/audiocpp_server.exe`
- `build/windows-cuda-release/bin/audiocpp_server.exe`

## Known limitations

- Model download/conversion from the WebUI is optional and still invokes the repository Python model-manager helper. Users with already prepared models do not need Python.
- The legacy Gradio interface remains in the repository for advanced Python-assisted or compatibility workflows; it is separate from the embedded native UI.
- Near-live ASR uses repeated server requests rather than a browser-side model runtime.


0xShug0:
The UI looks super cool!

mirek190:
Pushed commit `684832b` with the current native WebUI model-library update.

### What changed

- Added asynchronous model-package installation with live byte and percentage progress in the embedded UI.
- Exposed GGUF Q8, GGUF BF16/FP16, and Safetensors choices with their package-specific download sizes and availability state.
- Detects already-installed package variants using their declared files, disables the corresponding button, and shows **Already downloaded**.
- Allows precision variants that share a package directory to coexist without treating the directory itself as a conflict.
- Added per-variant trash controls and a server uninstall endpoint. Removal deletes only files declared by that package and preserves sibling variants and unrelated files.
- Corrected Pocket TTS package paths and its gated Safetensors messaging.
- Made model cards, Open controls, and package-choice rows consistent in size; same-row cards now align with uniform vertical gaps.
- Updated the embedded production WebUI artifact and documentation.

### Validation

- `npm --prefix webui/native run check` ? 0 errors, 0 warnings
- `python -m unittest webui.test_model_manager_webui` ? 16 tests passed
- `server_model_installer_test` ? passed
- CPU `audiocpp_server` Release build ? passed
- CUDA `audiocpp_server` Release build ? passed
- Browser layout check ? matching card heights within each row and consistent 12 px row gaps

mirek190:
Pushed follow-up commit `55f495a` 

### Root cause

All `ModelInstaller` instances used the same `%TEMP%/audiocpp-model-installer` directory and reused names such as `package-sizes-1.json`. A test process or second server could overwrite the live server's inventory snapshot. The UI would then lose the real installed-package records and leave an existing package clickable.

### Fix

- Give every installer/server instance a unique temporary job directory.
- Remove that private directory when its installer state is released after normal shutdown.
- Treat a package whose complete declared file set already exists as an idempotent successful install.
- Continue rejecting genuinely partial packages unless `--overwrite` is requested.
- Add regression coverage proving two installer instances cannot overwrite each other's inventory output.

### Validation

- Python model-manager suite: 17 tests passed.
- C++ `server_model_installer_test`: passed, including cross-instance isolation.
- CPU and CUDA Release server builds: passed.
- Live CUDA inventory: 127 packages returned; Pocket English Q8 and BF16 both reported `installed: true`.
- Exact Pocket Q8 install API request against the existing file: completed with exit code 0 and `already installed`, with no download or overwrite.

mirek190:
Pushed follow-up commit `0fe5bca` fixing installed-package precision selection and stale inventory after downloads.

### Changes

- Installed Q8/BF16/FP16/Safetensors buttons remain enabled and now select the model variant instead of trying to download it again.
- Buttons retain the format label and distinguish `Selected ? size` from `Downloaded ? size`.
- `Open` uses the selected installed package path.
- A successful install invalidates the server's cached package inventory.
- The frontend refreshes that inventory once for each newly completed install, so Studio ? Models navigation cannot restore stale installed flags.
- Added server regression coverage for inventory invalidation after a successful preparation.

### Validation

- Svelte check: 0 errors and 0 warnings.
- C++ `server_model_installer_test`: passed.
- CPU and CUDA Release server builds: passed.
- Browser test with Pocket TTS:
  - Q8 and BF16 both remained marked downloaded.
  - Switched Q8 ? BF16, navigated Models ? Studio ? Models, and BF16 remained selected.
  - Switched back to Q8 and pressed Open.
  - Studio resolved `models/PocketTTS-GGUF/english/pocket-tts-english-q8_0.gguf`.

mirek190:
Screenshot 2026-08-06 190453
Screenshot 2026-08-06 190605
Screenshot 2026-08-06 190628


# CPU
`  .\audiocpp_server.exe --ui --backend cpu --threads 8`

# NVIDIA CUDA
`  .\audiocpp_server.exe --ui --backend cuda --device 0`

# Vulkan — Windows or Linux
`  .\audiocpp_server.exe --ui --backend vulkan --device 0`

# AMD HIP/ROCm
`  .\audiocpp_server.exe --ui --backend hip --device 0`

# macOS
`  ./audiocpp_server --ui --backend metal`

0xShug0:
@mirek190 Thanks! I merged it into dev first because there is another relevent PR on UI change.

I will delay release 0.6 because the UI should be the headline feature in 0.6!

mirek190:
> @mirek190 Thanks! I merged it into dev first because there is another relevent PR on UI change.
> 
> I will delay release 0.6 because the UI should be the headline feature in 0.6!

No problem

What the another relevant PR on UI change?

0xShug0:
Here https://github.com/0xShug0/audio.cpp/pull/191

mirek190:
Screenshot 2026-08-06 225224

You mean this one?

0xShug0:
Nice. So the voice lib is manged by the UI? I guess the UI change in #191 is not necessary, while the server voice lib is still good to keep. Could you leave a comment to explain this feature in that PR?

mirek190:
> Nice. So the voice lib is manged by the UI? I guess the UI change in #191 is not necessary, while the server voice lib is still good to keep. Could you leave a comment to explain this feature in that PR?


Yes that webUI based on Svelte is completely independent from the python. 

But https://github.com/0xShug0/audio.cpp/pull/191 is still good for a clean server use

mirek190:
###   **Saved voices stored only in this browser**
  
  
  Saved voices feature: lets the user name and keep a reference voice recording (audio + transcript) for reuse across voice-clone sessions. Voices are persisted ONLY in this browser's IndexedDB.

0xShug0:
@mirek190 Great job! I really like the new UI!

A few things I noticed while testing the new UI:

* In the Models tab, clicking a Q8/Q16 card starts the download immediately. I accidentally ended up with several downloads running at the same time and couldn’t stop them 😭. Would it be possible to add `Stop Download` and `Clean Partial Download` options?

* There are also some duplicate entries in the model list, for example Vevo2. Could we deduplicate the list and show the available model variants inside each card instead? I think that would make the model page much cleaner.

* Do we have any model version checking right now? For example, whether the local model is still the same version as the one in the Hugging Face repo. Models may get updated under the same repo/file name, so it would be nice to show something like `Up to date` or `Update available`.

(**Update**)
* Can the UI include some prebuilt voices/reference text for a quick start? I found this Gradio feature handy.

Some model specific issues I found:

PocketTTS: PocketTTS should allow plain TTS without `voice_ref`, e.g., built in voice `voice_id=alba`

RVC: UI treats every task=vc as needing `voice_ref` but RVC’s target voice is selected by packaged `voice_id` or `voice_model_path`

Irodori, DramaBox: UI always includes `max_tokens` for tts/clon/vdes requests, but these two spec v1 models don't expose this option. Spec v1 validation reject unknown options so UI shows `unknown request option: max_tokens`. 

The real solution is full spec-v1 migration + UI driven by spec metadata, but until then, we need small model-specific UI patches.

mirek190:
Thanks for feedback  ;)

Issue #188, opened 2026-08-05
Irodori-TTS Vulkan codec decoder produces corrupted audio on Intel Iris Xe, CPU codec works correctly


datkun92-code:
**Title:** Irodori-TTS Vulkan codec decoder produces corrupted audio on Intel Iris Xe, CPU codec works correctly

### Environment

* audio.cpp: latest main branch
* Backend: Vulkan
* GPU: Intel Iris Xe Graphics (TGL GT2)
* Driver: Intel Mesa Vulkan
* OS: Linux (Debian via Distrobox)
* Model: Irodori-TTS v4 Small Q8

### Problem

I found a Vulkan-specific issue in the Irodori-TTS codec decoder.

When running the full pipeline on Vulkan, generated WAV output becomes corrupted (mostly noise/static).

However, if I keep:

* Condition Encoder = Vulkan
* RF Sampler = Vulkan
* Codec Decoder = CPU

the generated speech becomes completely correct.

This suggests the issue is inside the codec decoder Vulkan path rather than RF sampling.

### Evidence

RF sampler output looks normal:

```text
LATENT_RANGE -4.644324 4.728831
```

When codec decoder runs on Vulkan:

```text
DECODE_WAV_RANGE -2678900660450312629594952949856993280.000000 31.729752
```

Output audio is corrupted.

When codec decoder runs on CPU:

```text
DECODE_WAV_RANGE -0.810733 0.999992
```

Output audio is correct.

### Additional notes

I traced the issue down to the codec decode path:

```cpp
build_irodori_codec_decode(...)
```

Potentially related operations:

```cpp
Snake1dModule
ConvTranspose1dModule
SliceModule
```

The most suspicious component is:

```cpp
ggml_conv_transpose_1d
```

because:

* latent values entering the decoder are valid
* CPU decoder produces correct audio
* Vulkan decoder produces invalid waveform values

### Temporary workaround

Creating a separate CPU ExecutionContext only for:

```cpp
IrodoriCodec
```

while keeping RF sampler on Vulkan fixes the issue completely.

If needed, I can provide a patch and additional logs.

0xShug0:
@datkun92-code Thank you for reporting this! This is an interesting issue. It may be related to how GGML handles Intel/Mesa Vulkan for `ggml_conv_transpose_1d`. I could not reproduce it on my local Vulkan devices, but your isolation looks solid. 

My test: 

Vulkan device 0, NVIDIA RTX 5090
[...]

PR #189, opened 2026-08-05
Add Irodori codec backend selection

0xShug0:
## Summary
- add `irodori_tts.codec_backend=same|cpu` as an Irodori-only session option
- keep condition encoder and RF sampler on the requested backend while allowing the DACVAE codec decode to run on CPU
- keep old standalone GGUF packages usable when their embedded spec predates this option

## Validation
- `cmake --build build/debug -j 16 --target audiocpp_cli`
- Vulkan default path logs `irodori_tts.codec.backend same`
- Vulkan + `--session-option irodori_tts.codec_backend=cpu` logs `irodori_tts.codec.backend cpu` and produces valid audio
- invalid value rejects with `Invalid irodori_tts.codec_backend: bad`
- Qwen3-ASR on the CPU-codec output produced largely matching Japanese text

### DIFF
--- a/include/engine/models/irodori_tts/session.h
+++ b/include/engine/models/irodori_tts/session.h
@@ -73,6 +73,7 @@ private:
   assets::TensorStorageType codec_weight_storage_type_ =
       assets::TensorStorageType::Native;
   bool mem_saver_ = true;
+  std::unique_ptr<engine::core::ExecutionContext> codec_execution_context_;
   std::unique_ptr<IrodoriConditionEncoder> condition_encoder_;
   std::unique_ptr<IrodoriRfSampler> rf_sampler_;
   std::unique_ptr<IrodoriCodec> codec_;
--- a/model_specs/irodori_tts.json
+++ b/model_specs/irodori_tts.json
@@ -195,6 +195,17 @@
         "default": "native"
       },
       {
+        "name": "codec_backend",
+        "type": "enum",
+        "description": "DACVAE codec execution backend. Use cpu as an opt-in workaround for backend-specific codec decoder issues; default same.",
+        "values": [
+          "same",
+          "cpu"
+        ],
+        "required": false,
+        "default": "same"
+      },
+      {
         "name": "condition_graph_arena_mb",
         "type": "int",
         "description": "Condition encoder graph arena size in MiB; default 256.",
--- a/src/models/irodori_tts/session.cpp
+++ b/src/models/irodori_tts/session.cpp
@@ -29,6 +29,11 @@ namespace {
 using Clock = std::chrono::steady_clock;
 constexpr const char *kFamily = "irodori_tts";
 
+enum class IrodoriCodecBackend {
+  Same,
+  Cpu,
+};
+
 std::shared_ptr<const IrodoriTTSAssets>
 require_assets(std::shared_ptr<const IrodoriTTSAssets> assets) {
   if (assets == nullptr) {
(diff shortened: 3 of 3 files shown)


0xShug0:
@francescobozzo @fedeizzo It’s weird that this model-scoped change failed the Nix/Vulkan CI. Any thoughts on what might be causing this?

francescobozzo:
Interesting, the build "worked on my machine"... I tested using my nix system installation and also reproducing github workflows with containers

```sh
git clone git@github.com:0xShug0/audio.cpp.git audio.cpp-remote
cd audio.cpp-remote
gh pr checkout 189

nix build .#vulkan

nix run nixpkgs#act -- -W .github/workflows/nix-build.yml
```

0xShug0:
@francescobozzo Retry works. “Have you tried turning it off and on again?” I should’ve followed this golden rule :smile: .

datkun92-code:
I tested the PR on Intel Iris Xe. With:

[TRACE ts=20260806-122853] irodori_tts.codec.backend cpu -all work
[TRACE ts=20260806-123137] irodori_tts.codec.backend same -error wav correct


the RF sampler stays on Vulkan, the codec runs on CPU, and the generated audio is clean. The severe static/noise issue I was getting with full Vulkan decoding is gone.

Looks good from my side.
By the way, how do I pass the Caption / Style Prompt (optional) argument into the audio.cpp CLI for Irodori-TTS-v4-Small?

0xShug0:
@datkun92-code You can use `--request-option caption="...."`, optional -`-request-option caption_guidance_scale=...`

Example:
```
audiocpp_cli --task vdes --family irodori_tts \
  --model models/Irodori-TTS-v4-Small-GGUF/irodori-tts-v4-small-q8_0.gguf \
  --backend cuda --language ja \
  --text "本日はお越しいただき、誠にありがとうございます。" \
  --request-option caption="落ち着いた大人の男性。深く響く声で丁寧に話している。" \
  --request-option caption_guidance_scale=3 \
  --request-option no_ref=true \
  --out out.wav
```

[v4_caption_guidance_example.wav](https://github.com/user-attachments/files/30792879/v4_caption_guidance_example.wav)

datkun92-code:
The current `irodori_tts.codec_backend=cpu` workaround fixes the Vulkan decoder issue, but it still requires creating a separate CPU codec context and loading the codec weights again.

Would it make sense to support a standalone codec package (for example `irodori_codec.gguf`) that can be loaded independently from the RF/condition models?

Potential benefits:

* Avoid duplicating large model weights between Vulkan and CPU contexts.
* Reduce memory usage when only the codec needs to run on CPU.
* Faster initialization because the RF sampler and condition encoder remain on Vulkan while the codec is loaded separately.
* Easier debugging of codec-specific backend issues.
* Could potentially allow sharing the same codec across multiple Irodori model variants.

The current CPU codec backend already demonstrates that the codec can operate independently, so separating it into its own GGUF package might be a natural next step.
I did some additional debugging on the Vulkan codec issue.

Findings:

* CPU codec backend always produces correct audio.
* Vulkan codec starts failing only on longer generations.
* The failure appears around latent_steps ~=> 256.
* latent values themselves remain normal:
  LATENT_RANGE ≈ [-4.5, 5.3]
* No NaN or Inf are observed.
* Decoder block 0 works.
* Decoder block 1 works.
* Decoder block 2 works.
* The issue appears after the final decoder block (stride=2).
* With all decoder blocks enabled, Vulkan decode can produce either:

  * all zeros
  * or extremely large invalid values
* CPU decode on the same latent produces correct audio.

This makes me suspect a Vulkan ConvTranspose1d path issue (possibly related to large output lengths), rather than RF sampling or latent generation.

The CPU codec workaround completely fixes the problem for me, but I wanted to share the debugging results in case you decide to investigate the Vulkan path later.

0xShug0:
@datkun92-code Thank for the information! As mentioned in https://github.com/0xShug0/audio.cpp/issues/188, This isn’t a generic Vulkan issue; it only appears with Vulkan under a specific setting. Changing the global Vulkan ConvTranspose1d lowering would therefore be risky and have a large blast radius.

Actually, I think a better solution would be to add an opt-in flag that uses a special `ConvTranspose1d` lowering for the Irodori-TTS codec, assuming we confirm that this is the root cause. The change should remain scoped to the Irodori-TTS codec implementation and bypass the shared framework `ConvTranspose1d` module, so it doesn't not affect other paths and is additive. It's not safe to make the change as default because we don't know the performance and parity impact on other backends/settings.

With this, we can actually remove the CPU backend hack to make the code clean. 

Would you be interested in submitting a PR for this? The issue currently seems reproducible only in your environment, so you’re in the best position to implement and validate the fix.

PR #193, opened 2026-08-07
vulkan: disable integer dot product on AMD proprietary driver without…

### LINKED ISSUE #192: Vulkan: q8_0 Irodori-TTS output saturates to +32767 on AMD RX 6800; disabling integer dot avoids it
## Summary

On an AMD Radeon RX 6800 (Windows, AMD proprietary driver), Irodori-TTS synthesis with a **q8_0** GGUF on the Vulkan backend produces a WAV whose samples are **all exactly +32767** — constant full-scale output rather than degraded audio.

The same model works correctly in the following cases:

* q8_0 on CPU
* f16 on Vulkan
* q8_0 on Vulkan with `GGML_VK_DISABLE_INTEGER_DOT_PRODUCT=1`
* q8_0 loaded with `irodori_tts.weight_type=f16`

Disabling other Vulkan paths such as MMVQ, fusion, or f16 does not change the failure.

This strongly suggests that the failure is in, or triggered by, ggml-vulkan's integer-dot path used by q8_0 on this device. I have not isolated the specific ggml operation or kernel yet.

Everything below was reproduced on **pristine upstream `2795a54`** (current `main`, including #189), built from source with no local patches.

The failure is silent: the duration predictor and every `--metrics` value appear normal. Only the generated waveform is invalid.

## Localization

Same command each time; only the environment variable changes:

| run                                     | distinct sample values |    RMS | result      |
| --------------------------------------- | ---------------------: | -----: | ----------- |
| default                                 |                      1 | 1.0000 | **broken**  |
| `GGML_VK_DISABLE_INTEGER_DOT_PRODUCT=1` |                 25,462 | 0.1291 | **correct** |
| `GGML_VK_DISABLE_MMVQ=1`                |                      1 | 1.0000 | broken      |
| `GGML_VK_DISABLE_FUSION=1`              |                      1 | 1.0000 | broken      |
| `GGML_VK_DISABLE_F16=1`                 |                      1 | 1.0000 | broken      |

Only disabling integer dot changes the outcome.

I also reproduced the same behavior with the v3 600M VoiceDesign q8_0 GGUF: broken by default on Vulkan and correct with `GGML_VK_DISABLE_INTEGER_DOT_PRODUCT=1`. This therefore does not appear to be specific to Irodori-TTS v4.

The affected device reports integer-dot support but no matrix cores:

```text
ggml_vulkan: 0 = AMD Radeon RX 6800 (AMD proprietary driver) | uma: 0 | fp16: 1 | bf16: 0 |
             warp size: 32 | shared memory: 32768 | int dot: 1 | matrix cores: none
```

This may explain why the same issue did not reproduce on the RTX 5090 used in #188: a coopmat-capable device may select a different q8_0 matmul path. This is only a hypothesis; I have not verified the exact dispatch path on either device.

### ggml backend tests do not reproduce it

I built `test-backend-ops` from the vendored `external/ggml` using a standalone configure, since audio.cpp forces `GGML_BUILD_TESTS=OFF`, and ran the full Vulkan suite:

```text
$ test-backend-ops test -b Vulkan0
  13641/13641 tests passed
  Backend Vulkan0: OK
```

`MUL_MAT` alone also passes 947/947 tests, including q8_0 cases.

So the issue does not reproduce with the shapes covered by `test-backend-ops`; it appears to require something about the real model graph, such as a particular shape, size, stride, execution sequence, or another graph-level condition in the integer-dot path.

## Relationship to #188 / #189

[#188](https://github.com/0xShug0/audio.cpp/issues/188) reported corrupted Irodori output on Intel Iris Xe. In that case the problem was traced to the codec decoder and was worked around by the `irodori_tts.codec_backend=cpu` option added in [#189](https://github.com/0xShug0/audio.cpp/pull/189).

That workaround does **not** fix this case:

```text
--session-option irodori_tts.codec_backend=cpu
```

still produces constant +32767 output.

The option is being parsed: setting

```text
irodori_tts.codec_backend=bad
```

correctly fails with:

```text
Invalid irodori_tts.codec_backend: bad
```

Since moving the codec decoder to CPU does not change the result, while running the entire q8_0 model on CPU is correct, this strongly suggests that corruption occurs before or outside the codec decoder — most likely somewhere in the condition encoder or RF sampler.

I have not captured intermediate latent tensors, so I cannot directly confirm the first stage at which they become invalid.

This differs from #188, where the reporter observed approximately:

```text
LATENT_RANGE -4.64 4.73
DECODE_WAV_RANGE -2.6e36 ...
```

That case had apparently healthy latents and corruption appearing during codec decode, making `codec_backend=cpu` sufficient.

On the RX 6800, the failure appears to occur earlier.

It is possible that #188 and this issue still share a lower-level Vulkan cause but manifest first in different stages depending on device and dispatch path. I have not verified that hypothesis.

## Reproduction

No reference audio or additional assets are required; the plain `tts` task is enough.

```sh
# broken: every sample is +32767
audiocpp_cli --task tts --family irodori_tts \
  --model <models>/Irodori-TTS-v4-Small-GGUF/irodori-tts-v4-small-q8_0.gguf \
  --backend vulkan --threads 8 --language ja --seed 1234 \
  --text 'こんにちは、テストです。' \
  --request-option num_inference_steps=24 --out q8_vulkan.wav
```

With integer dot disabled, the same model produces correct speech:

```sh
# Linux/macOS shell syntax shown here.
# On Windows, set GGML_VK_DISABLE_INTEGER_DOT_PRODUCT=1 before running the command.

GGML_VK_DISABLE_INTEGER_DOT_PRODUCT=1 \
audiocpp_cli --task tts --family irodori_tts \
  --model <models>/Irodori-TTS-v4-Small-GGUF/irodori-tts-v4-small-q8_0.gguf \
  --backend vulkan --threads 8 --language ja --seed 1234 \
  --text 'こんにちは、テストです。' \
  --request-option num_inference_steps=24 --out q8_vulkan_nointdot.wav
```

Also correct:

```sh
# same q8_0 file on CPU
audiocpp_cli ... --backend cpu --out q8_cpu.wav

# f16 model on Vulkan
audiocpp_cli ... \
  --model .../irodori-tts-v4-small-f16.gguf \
  --backend vulkan --out f16_vulkan.wav
```

Does **not** fix it:

```sh
audiocpp_cli ... \
  --backend vulkan \
  --session-option irodori_tts.codec_backend=cpu
```

The models are the stock HF packages from `audio-cpp/audio.cpp-gguf`. `irodori_tts_v4_small_q8_0` is currently the `default: true` package for this family.

## Measurements

Irodori-TTS v4 Small, text `こんにちは、テストです。`, seed 1234, `num_inference_steps=24`:

| model / backend                                        | samples | duration |       min |       max | distinct values | RMS (normalized) |
| ------------------------------------------------------ | ------: | -------: | --------: | --------: | --------------: | ---------------: |
| q8_0 / **vulkan**                                      | 178,560 |   3.72 s | **32767** | **32767** |           **1** |       **1.0000** |
| q8_0 / vulkan, `codec_backend=cpu`                     | 178,560 |   3.72 s | **32767** | **32767** |           **1** |       **1.0000** |
| q8_0 / vulkan, `GGML_VK_DISABLE_INTEGER_DOT_PRODUCT=1` | 180,480 |   3.76 s |   −23,860 |    24,067 |          25,462 |           0.1291 |
| f16 / vulkan                                           | 178,560 |   3.72 s |   −26,971 |    24,078 |          25,741 |           0.1312 |
| q8_0 / cpu                                             | 180,480 |   3.76 s |   −23,354 |    23,557 |          25,485 |           0.1302 |
| q8_0 / vulkan, `irodori_tts.weight_type=f16`           | 180,480 |   3.76 s |   −25,943 |    24,020 |          25,325 |           0.1287 |

The last row uses the same q8_0 GGUF but dequantizes the weights to f16 at load time. It also produces correct output, which is consistent with the problem being related to the quantized Vulkan execution path rather than the GGUF file itself or its loader.

The small duration differences are ordinary numerical drift in the duration predictor and do not appear related to the failure.

I did not capture intermediate floating-point buffers, so I cannot determine whether the invalid values are NaN/Inf or simply extremely out of range. The constant +32767 int16 output is consistent with severe positive saturation occurring before or during the final float-to-int16 conversion.

## Why this is worth fixing rather than only working around

`irodori_tts_v4_small_q8_0` is the `default: true` package for `irodori_tts`, and q8_0 is also the default for many families in `model_specs/`.

On an affected device, a user can therefore:

1. download the default model,
2. select the Vulkan backend,
3. receive no error or obviously abnormal metric,
4. get a completely invalid WAV.

The current workaround is:

```text
GGML_VK_DISABLE_INTEGER_DOT_PRODUCT=1
```

or using f16 instead of q8_0.

The root cause may ultimately belong upstream in ggml-vulkan. However, audio.cpp vendors ggml under `external/ggml` and distributes q8_0 as the default model for this family, so the resulting failure is directly user-visible from audio.cpp.

## Environment

* **OS:** Windows 11 Pro 10.0.26200
* **GPU:** AMD Radeon RX 6800
* **Driver:** 32.0.21045.1000 (2026-07-23), AMD Windows proprietary driver
* **audio.cpp:** pristine `2795a54` (`main`), no local patches
* **ggml:** vendored `external/ggml`, version 0.12.0

  * configure output reports `ggml commit: 2795a54`; this appears to inherit the host audio.cpp repository commit because ggml is vendored rather than a separate Git checkout
* **Build:** MSVC 19.44.35228.0, Ninja, `Release`
* **CMake options:**

```text
-DENGINE_ENABLE_VULKAN=ON
-DENGINE_ENABLE_OPENMP=ON
-DGGML_OPENMP=ON
-DENGINE_BUILD_TESTS=OFF
```

Built from source because `release-0.5.1` does not include a Windows Vulkan prebuilt.

* **Models:**

  * `Irodori-TTS-v4-Small-GGUF`
  * `Irodori-TTS-600M-v3-VoiceDesign-GGUF`
  * q8_0 and f16
  * from `audio-cpp/audio.cpp-gguf`

## Not tested

* NVIDIA GPUs
* AMD GPUs under RADV/Mesa on Linux
* Metal
* Other model families using q8_0
* The exact ggml operation or shader at which corruption first appears

Only `irodori_tts` has been exercised so far. If this is a shared q8_0 Vulkan execution issue, other families may also be affected on devices taking the same integer-dot path, but I have not confirmed that.

Because `test-backend-ops` passes completely on this device, further isolation will probably require instrumentation of the real model graph, for example dumping intermediate tensor ranges per node.

I can reproduce the failure reliably and can run an instrumented build if useful.

IIIIIllllIIIIIlllll:
Fix: #192
On AMD GPUs without native dot4 hardware (GCN/RDNA1/RDNA2, e.g. RX 6800), the AMD proprietary Windows driver advertises integerDotProduct4x8BitPackedSignedAccelerated=true, so ggml-vulkan routes q8_0 matmuls through the integer-dot path (dotPacked4x8EXT). That path silently produces wrong results there — with the default q8_0 Irodori-TTS package the output WAV is saturated to a constant +32767 with no error or abnormal metric. RADV reports accelerated=false on the same hardware, and RDNA3+ has native dot4. Same signature reported in [ollama#13108](https://github.com/ollama/ollama/issues/13108) (RX 6650 XT).

**Fix**
- Force integer_dot_product=false for the affected combination (AMD vendor + eAmdProprietary + GCN/RDNA1/RDNA2), mirroring the existing old_amd_windows quirk condition. Equivalent to the known workaround GGML_VK_DISABLE_INTEGER_DOT_PRODUCT=1, applied automatically; also covers MUL_MAT_ID and flash-attention MMQ paths. The device-info path is adjusted too so the int dot log line stays truthful. RDNA3+, RADV, NVIDIA, Intel are unaffected.

**Verification**
- Reporter confirmed on the affected RX 6800: device now reports int dot: 0, q8_0/Vulkan output is bit-identical to the env-var workaround (25,462 distinct samples, RMS 0.1291), f16 unchanged, no measurable perf cost, test-backend-ops still 13,641/13,641.
- ggml-vulkan target compiles cleanly (MSVC 19.44, Vulkan SDK 1.4.350).

AI usage: Kimi K3 helped me find the issue and performed a fix.

### DIFF
--- a/external/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/external/ggml/src/ggml-vulkan/ggml-vulkan.cpp
@@ -395,6 +395,19 @@ static vk_device_architecture get_device_architecture(const vk::PhysicalDevice& 
     return vk_device_architecture::OTHER;
 }
 
+// Workaround for the AMD proprietary driver advertising
+// integerDotProduct4x8BitPackedSignedAccelerated on GPUs without native dot4
+// hardware (GCN/RDNA1/RDNA2): the emulated integer dot product path silently
+// produces incorrect quantized matmul results
+// (https://github.com/0xShug0/audio.cpp/issues/192). RADV reports
+// accelerated=false on the same hardware, and RDNA3+ has native dot4 support.
+static bool ggml_vk_amd_proprietary_emulated_int_dot(uint32_t vendor_id, vk::DriverId driver_id, vk_device_architecture architecture) {
+    return vendor_id == VK_VENDOR_ID_AMD && driver_id == vk::DriverId::eAmdProprietary &&
+           (architecture == vk_device_architecture::AMD_GCN ||
+            architecture == vk_device_architecture::AMD_RDNA1 ||
+            architecture == vk_device_architecture::AMD_RDNA2);
+}
+
 enum vk_conv_shapes {
     CONV_SHAPE_128x128,
     CONV_SHAPE_64x32,
@@ -5286,6 +5299,10 @@ static vk_device ggml_vk_get_device(size_t idx) {
 
         device->integer_dot_product = device->integer_dot_product && shader_integer_dot_product_props.integerDotProduct4x8BitPackedSignedAccelerated;
 
+        if (ggml_vk_amd_proprietary_emulated_int_dot(device->vendor_id, device->driver_id, device->architecture)) {
+            device->integer_dot_product = false;
+        }
+
         device->min_imported_host_pointer_alignment = external_memory_host_props.minImportedHostPointerAlignment;
 
         device->max_workgroup_size_log2 = uint32_t(log2f(float(device->properties.limits.maxComputeWorkGroupInvocations)));
@@ -5947,6 +5964,10 @@ static void ggml_vk_print_gpu_info(size_t idx) {
     integer_dot_product = integer_dot_product
                        && shader_integer_dot_product_props.integerDotProduct4x8BitPackedSignedAccelerated
                        && shader_integer_dot_product_features.shaderIntegerDotProduct;
+
+    if (ggml_vk_amd_proprietary_emulated_int_dot(props2.properties.vendorID, driver_props.driverID, device_architecture)) {
+        integer_dot_product = false;
+    }
 
     coopmat_support = coopmat_support
 #if defined(GGML_VULKAN_COOPMAT_GLSLC_SUPPORT)


0xShug0:
Thank you @IIIIIllllIIIIIlllll!  Great work! I'm really glad to have your help on the project.

PR #197, opened 2026-08-08
Server: support models unloading

nikich340:
**Summary**
Adds two endpoints to server: v1/tasks/unload_models, v1/tasks/unload_all_models.

**Purpose:**
In some situations, I need to free all the possible VRAM while keeping server functional and ready for new tasks without restart. That's where these features will help.

**Confirmation:**
Successfully built and tested on Windows 11 (CUDA sm_120 avx-512 release build) with omnivoice, qwen3-asr models. The server correctly handles new endpoints. When a model is in use, it properly waits until it finishes and then do unloading. The next requests to server work fine, with model loading back to backend device.

### DIFF
--- a/app/server/README.md
+++ b/app/server/README.md
@@ -402,7 +402,7 @@ curl http://127.0.0.1:8080/v1/tasks/run \
 
 Unload specific models from memory to free resources (e.g. VRAM on GPU backends). Subsequent requests to an unloaded model will trigger a transparent reload. The server waits for any in-flight inference on each target model to complete before unloading it.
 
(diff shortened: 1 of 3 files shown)


0xShug0:
@nikich340 Thanks for the PR! Two minor issues:

1. `LoadedModel::unload() `resets model before session. Normal member destruction order would destroy session before model, and unload should preserve that safer lifetime order:
`offline=nullptr, streaming=nullptr, session.reset(), then model.reset().`

2. Typo: `app/server/README.md`: docs say `model-ids`, but code requires` model_ids`.

Could you fix them?

nikich340:
Sure, you are right, done!

0xShug0:
@nikich340 Merged. Thanks!

PR #199, opened 2026-08-08
Improve native WebUI model management

mirek190:
## Summary

Follow-up to #185 and the maintainer review in https://github.com/0xShug0/audio.cpp/pull/185#issuecomment-5219818386.

- require confirmation before starting a model download, allow concurrent package jobs, and add **Stop download** plus **Clean partial download** actions
- group repeated model entries into one family card with nested variants and suppress duplicate package buttons
- persist the resolved Hugging Face revision and report **Up to date**, **Update available**, or **Version unknown** for older installs
- expose spec-defined request controls instead of submitting unsupported generic fields
- ship bundled quick-start reference voices and reference transcripts in Windows packages
- allow PocketTTS to run without an uploaded reference by using its built-in `alba` voice
- avoid requiring `voice_ref` for RVC and preserve its target-voice controls
- avoid sending `max_tokens` to strict model-spec-v1 models that do not declare it, including Irodori/DramaBox

## Package lifecycle

Downloads use package-scoped staging directories and a cooperative cancellation marker checked between 1 MiB chunks. Cancelling cleans staging data; a separate cleanup endpoint removes abandoned partial package directories. Package manifests record requested/resolved revisions, sizes, and ETags.

Precision variants can safely share identical sidecars. For example, PocketTTS Q8 and BF16 both use the public Alba embedding: installing the second precision reuses it, and uninstalling one precision preserves it while the sibling remains installed.

Existing installs made before these manifests are shown as **Version unknown** until they are reinstalled once.

## Validation

Backend: Windows CUDA 12.4, NVIDIA RTX 3090.

- `npm run check` — 0 errors, 0 warnings
- `npm run build` — embedded native WebUI rebuilt successfully
- `python -m py_compile tools/model_manager_v2.py`
- `python -m json.tool model_specs/pocket_tts.json`
- mocked model-manager revision test: matching revision -> `up_to_date`; changed revision -> `update_available`; partial staging cleaned
- mocked shared-sidecar test: second precision reused Alba; removing Q8 preserved Alba while BF16 remained; removing BF16 then removed Alba
- `server_model_installer_test.exe` — passed, including queued/running cancellation and cleanup lifecycle
- complete `audiocpp_server` CUDA debug build — passed
- browser QA: Vevo2 rendered as one family with 3 variants; Qwen3-TTS rendered as one family with 4 variants; PocketTTS selected Alba by default without requiring an uploaded reference
- PocketTTS end-to-end server request using GGUF Q8 + Alba and no `voice_ref`: HTTP 200, `audio/wav`, 169,004-byte generated artifact


mirek190:
Implemented the complete native WebUI follow-up requested in the [review comment](https://github.com/0xShug0/audio.cpp/pull/185#issuecomment-5219818386), together with the subsequent model-library usability fixes.

| Request | Status | Implementation |
|---|---:|---|
| Stop active downloads | ✅ | Active package jobs expose **Stop download** and transition through cancelling/cancelled states. |
| Clean partial downloads | ✅ | Failed or cancelled jobs expose **Clean partial download**, with visible cleaned-state feedback. |
| Deduplicate model entries | ✅ | Models are grouped by family; variants such as Vevo2 and Qwen3-TTS appear inside one family card. |
| Model version checking | ✅ | Install manifests record the resolved Hugging Face revision and ETags. The UI displays **Up to date**, **Update available**, or **Version unknown**. |
| Quick-start voices and transcripts | ✅ | Built-in voices and bundled reference WAV/transcript pairs are exposed in Studio. Voice sources are mutually exclusive to prevent stale conditioning. |
| PocketTTS plain TTS | ✅ | PocketTTS defaults to the built-in `alba` voice and does not require a reference recording. |
| RVC target voice handling | ✅ | RVC no longer requires `voice_ref`; target voices are selected through `voice_id` or `voice_model_path`. |
| Irodori/DramaBox `max_tokens` validation | ✅ | `max_tokens` is shown and sent only when declared by the model's request-option metadata. |
| Download format selection | ✅ | GGUF Q8, GGUF BF16/FP16, and Safetensors are represented independently, including installed/resident state. |
| Model-library layout | ✅ | Families and variants are alphabetically sorted, cards use independent compact columns, and task filters are available beside the model-folder selector. |
| Resident model controls | ✅ | Studio exposes format-specific load/unload buttons; unavailable formats are disabled and only the resident format receives the green indicator. |
| Custom model folder | ✅ | Downloads, local detection, and model loading share a configurable model root with a default beside the server binary. |

Version-note: packages installed before revision manifests were introduced display **Version unknown** until they are reinstalled once. The UI is substantially driven by `model_specs_v1` metadata while retaining compatibility fallbacks for older or incomplete specs.

Validation:

- `python -m py_compile tools/model_manager_v2.py`
- `npm run check` — 0 errors and 0 warnings
- `npm run build` — production static bundle generated successfully

Update commit: `bbb76bf` (`Complete native WebUI model management follow-up`).

mirek190:
Screenshot 2026-08-09 020043


Also I think we could name those example voices better ... any ideas?

0xShug0:
> Screenshot 2026-08-09 020043
> Also I think we could name those example voices better ... any ideas?

Yes we can normalize the naming to `demo_[male/female]_[language]_[num]`, e.g.,
 ```
demo_male_en_01, demo_male_en_02, demo_female_en_01,  demo_female_zh_01
```

**Comment by mirek190** [comment]
> _01, demo_male_en_02, demo_female_en_01,  demo_female_zh_01

Adding suffix en , zh, it, pl, es  has no sense as the audio is generated based on the prompt language  


I will make like that 

```
demo_01_man       demo_1 man
demo_03_man       demo_2 man
demo_02_woman     demo_3 woman
ref-w1            demo_4 woman
```

mirek190:
Added embedded native WebUI demo voices in `bdada27`.

### What changed

- Copied the four quick-start reference WAVs into `webui/native/demo_voices/` and renamed them to the identifiers displayed by the native UI:
  - `demo_1_man`
  - `demo_2_man`
  - `demo_3_woman`
  - `demo_4_woman`
- Added a matching `prompt_text` entry for every WAV. The original-language transcripts are intentionally preserved because they are model-conditioning data, not interface labels.
- Added a platform-neutral CMake asset-generation step. Each WAV and `prompt_text` is converted to a generated byte array and linked into `audiocpp_server` on Windows, Linux and macOS builds.
- Added a small embedded-demo API beside the existing embedded HTML asset API.
- When native UI management starts and no explicit `voice_dir` is configured, the server materializes the compiled assets into its per-run temporary voice library. This reuses the existing WAV decoder, voice-list endpoint and reference-transcript injection flow.
- Temporary demo files are removed with the normal server runtime directory during shutdown. An explicitly configured `voice_dir` still takes precedence.
- Updated the quick-start selector to use the embedded names directly and removed the old external aliases/default selection.
- Kept visible catalog, hint and parameter text English-only while retaining non-English reference transcripts needed for voice cloning.

### Result

The native WebUI quick-start voices are now self-contained in `audiocpp_server.exe`; distributing the server binary no longer requires the repository-level `webui/voice` directory for these demos.

### Verification

- `npm run check`: 0 errors, 0 warnings
- `npm run build`: passed
- CUDA Release `audiocpp_server` build: passed
- `/v1/audio/voices` returned exactly the four renamed embedded identifiers
- SHA-256 comparison confirmed every copied WAV is byte-identical to its source
- Embedded UI contains the new identifiers and no old demo filenames
- Ctrl+C stopped the test server, released its port and removed the per-run extracted demo directory

mirek190:
Added multilingual native WebUI support in commit `01b85c2`.

### What changed

- Added an embedded English fallback dictionary and automatic discovery of external language packs from `webui/native/lang/lang_*.json`.
- Added complete Italian, Polish, Russian, and Simplified Chinese language packs.
- Added an interface-language selector in the top-right header, next to the backend indicator.
- The initial language follows the browser preferences when supported; the user's selection is persisted in `localStorage` and restored on later visits.
- Localized the Studio, Models, Runtime, folder browser, workflow navigation, request controls, voice-library controls, output controls, package state/version labels, download controls, and common empty/status text.
- Replaced the visually exposed browser-native file labels with translated file-picker labels.
- Kept model names and technical identifiers such as GGUF, BF16, CUDA, VRAM, and JSON unchanged.
- Added `webui/native/lang/README.md` documenting the language-pack format and automatic discovery behavior.
- Rebuilt `webui/native/dist/index.html` so all translations remain embedded in `audiocpp_server.exe`; no external language files are needed at runtime.

### Adding another language

Add a file such as `webui/native/lang/lang_de.json` with `code`, `name`, and `translations`. The Vite build discovers it automatically and adds it to the selector; missing keys fall back to built-in English.

### Validation

- `npm run check` — 0 errors, 0 warnings
- `npm run build` — passed
- Windows CUDA Release `audiocpp_server` build — passed
- Browser QA confirmed switching and localized primary surfaces for Italian, Polish, Russian, and Chinese
- Translation coverage audit — 0 missing English keys in all four language packs

Is fully multi language now ;)



Screenshot 2026-08-09 144949
Screenshot 2026-08-09 145014
Screenshot 2026-08-09 145034
Screenshot 2026-08-09 145050

mirek190:
[0xShug0](https://github.com/0xShug0) can you sync that to dev for testing?

0xShug0:
@mirek190 Done! I'll test it later. AFK now

mirek190:
> @mirek190 Done! I'll test it later. AFK now

thanks

0xShug0:
@mirek190 

I think the UI is in pretty good shape overall.  Multilingual support is a really nice improvement!

A few issues left:

1. The model status check seems too strict. I uploaded some new models, and after that all models showed `reinstall`, which makes it look like every installed model is outdated.

2. It is not obvious how to unload a model. And I found when switching between different models, VRAM was not released. Maybe we could add a quick action bar at the top that shows the currently loaded models, so users can quickly unload a model with one click?

3. PocketTTS looks like has a race condition issue. Sometimes plain TTS without a reference voice fails with:
   `PocketTTS session prepare() requires a session voice via --voice-id or --voice-ref`

4. DramaBox fails if `reference_text` is sent: `unknown DramaBox request option: reference_text`

5. I think we should remove the safetensor option entirely. Currently, there are some mismatches between the model specs and what actually gets downloaded. E.g., loading OmniVoice safetensor downloaded via UI shows something like `failed to load model resources using model spec model_specs/omnivoice.json' source 'safetensors': missing model package file 'audio_tokenizer_preprocessor': models/OmniVoice/audio_tokenizer/preprocessor_config.json`.

I actually found several issues in the model specs and framework while testing the UI, and I've already fixed some of them :)

I'm ready to merge UI into main. 

Release 0.6 will include 4 models: DotTTS, NeuTTS-2e, MuScriptor (audio to MIDI), and (community model, experimental) Minimax-H3 text to audio (with text to video as a free bonus). We may need a new tab for H3 :). I can help with the UI after finishing testing H3.

mirek190:
about 2 - i think better will be just unload model switching to another one as modes are quite small so a  loading time is literal a second but bigger ones still have to unload everything to fit in the memory ;)  

I also planning to implement a streaming data to vram straight .... in the core audio.cpp.

PR #205, opened 2026-08-10
Fix handling of pocket_tts in model_manager

### LINKED ISSUE #204: model_manager_v2.py install fails for all pocket_tts packages: FileNotFoundError on final rename
### Summary

  `tools/model_manager_v2.py install` crashes at the very end — after fully downloading the model — for any package whose `target_directory` contains a `/`. Today that is every `pocket_tts` package (11 of them), since `model_specs/pocket_tts.json` is the only spec using nested targets (`PocketTTS-GGUF/english`, `.../german`, `.../italian`, `.../portuguese`, `.../spanish`).

  ### Reproduction

  ```
  $ python3 tools/model_manager_v2.py install pocket_tts
  selected pocket_tts_english_q8_0 (pocket_tts)
  repo audio-cpp/audio.cpp-gguf@main
  target models/PocketTTS-GGUF/english
  file remote=PocketTTS-GGUF/english/pocket-tts-english-q8_0.gguf local=PocketTTS-GGUF/english/pocket-tts-english-q8_0.gguf
  Traceback (most recent call last):
    File "tools/model_manager_v2.py", line 339, in <module>
      raise SystemExit(main())
    File "tools/model_manager_v2.py", line 329, in main
      install_package(select_package(records, args), args)
    File "tools/model_manager_v2.py", line 237, in install_package
      staging.rename(final_dir)
    File "/usr/lib/python3.14/pathlib/__init__.py", line 1076, in rename
      os.rename(self, target)
  FileNotFoundError: [Errno 2] No such file or directory: '.../models/.PocketTTS-GGUF_english.ci702_fn' -> 'models/PocketTTS-GGUF/english'
  ```

  `--dry-run` and `--check` both pass, so the failure only shows up after the download completes (~128 MB wasted for the default q8_0 package).

  ### Cause

  In `install_package`, only the models root is created before staging:

  ```python
  models_root.mkdir(parents=True, exist_ok=True)
  staging = Path(tempfile.mkdtemp(prefix=..., dir=models_root))
  ...
  staging.rename(final_dir)
  ```

  For a flat `target_directory` (e.g. `HTDemucs-GGUF`) the rename target's parent is `models/`, which exists. For a nested one, the parent is `models/PocketTTS-GGUF/`, which nothing ever creates — so `os.rename` fails on the missing intermediate directory. The staging tree
  itself is fine; `download_file` creates its parents via `output_path.parent.mkdir(parents=True, ...)`.

  The `except` handler then removes the staging directory, so the downloaded file is discarded and a retry re-downloads from scratch.

  ### Suggested fix

  Create the target's parent instead of just the models root, and stage alongside the final location so the rename is always an intra-directory move:

  ```python
  final_dir.parent.mkdir(parents=True, exist_ok=True)
  staging = Path(tempfile.mkdtemp(prefix=f".{package.target_directory.replace('/', '_')}.", dir=final_dir.parent))
  ```

  I applied this locally and `install pocket_tts` now succeeds; the installed file matches the size reported by `--check` (127,856,704 bytes) and no staging directory is left behind. Happy to open a PR if useful.

  ### Possibly worth fixing at the same time

  `tempfile.mkdtemp` creates its directory with mode `0700`, and the staging directory *becomes* the installed directory via the rename — so installed model directories end up `drwx------` rather than following the user's umask. This affects flat targets too, so it's not
  specific to this bug, but it can break setups where models are read by another user or service. A `chmod` on the staging directory before the rename would resolve it.

  ### Environment

  - audio.cpp at `a48cc34`, branch `release-0.5.1`
  - Python 3.14, Linux

P.S. Bug report generated with Claude.

yegorius:
Fix #204

### DIFF
--- a/tools/model_manager_v2.py
+++ b/tools/model_manager_v2.py
@@ -227,8 +227,8 @@ def install_package(package: PackageRecord, args: argparse.Namespace) -> None:
         return
     if final_dir.exists() and not args.overwrite:
         raise ManagerError(f"target already exists: {final_dir} (use --overwrite)")
-    models_root.mkdir(parents=True, exist_ok=True)
-    staging = Path(tempfile.mkdtemp(prefix=f".{package.target_directory.replace('/', '_')}.", dir=models_root))
+    final_dir.parent.mkdir(parents=True, exist_ok=True)
+    staging = Path(tempfile.mkdtemp(prefix=f".{package.target_directory.replace('/', '_')}.", dir=final_dir.parent))
     try:
         for remote, output in plan:
             download_file(package, remote, staging / output.relative_to(final_dir))


0xShug0:
@yegorius Merged. Thank you for the PR!

PR #206, opened 2026-08-10
Fix native WebUI package and request handling

mirek190:
## Summary

Continuation of #199, addressing the follow-up issues reported after merge.

| Area | Root cause | Fix | User impact |
|---|---|---|---|
| Model version status | The UI compared Hugging Face's repository-wide `X-Repo-Commit`, which changes when any file is uploaded to a shared repository. | Compare package-owned file ETags recorded in the install manifest. Older manifests without comparable ETags remain `Version unknown` without showing a reinstall action. | Uploading an unrelated model no longer marks every installed model as outdated. |
| Model and quant switching | Resident state was keyed primarily by model ID, allowing an old model or another precision of the same model to remain in VRAM. | Before loading, unload every different resident model and unload the same model when its selected GGUF path changes. Resident state now includes the exact selected model path. | Switching models or Q8/BF16/FP16 releases the previous allocation before loading the replacement. No additional unload controls were added. |
| PocketTTS plain TTS | A request without `voice_ref` could reach `prepare()` without a session voice. | When no reference or quick-start voice is selected, send the spec-defined default voice `alba`; the server maps it to `cached_voice_id`. | Plain PocketTTS requests no longer depend on stale session state and avoid the intermittent missing-voice error. |
| DramaBox request validation | The generic TTS form sent `reference_text` even though DramaBox does not declare that request option. | Respect explicit spec-v1 request-option metadata before forwarding `reference_text`, with backward-compatible behavior for specs not yet declaring the full option list. | DramaBox no longer fails with `unknown DramaBox request option: reference_text`; Qwen3-TTS reference transcripts continue to work. |
| Safetensors model-manager choices | Some one-click Safetensors packages are incomplete without source-tree sidecars. | Remove Safetensors install/load choices from the native UI and expose complete GGUF packages only. Safetensors-only managed entries are omitted, while bundled and locally managed entries remain visible. | The UI no longer installs packages that cannot be loaded, and models such as Qwen3-TTS 0.6B do not appear until a GGUF package exists. CLI/framework Safetensors support is unchanged. |

The installer also records metadata for package-owned sidecars reused from sibling precisions, so later ETag checks cover the complete package payload.

## Validation

- synchronized with `dev` at `3b1bbae`
- `python -m py_compile tools/model_manager_v2.py`
- per-file version-state test:
  - unrelated repository revision + matching file ETag -> `up_to_date`
  - changed package file ETag -> `update_available`
  - missing legacy manifest -> `unknown`
- `npm run check` -> 0 errors, 0 warnings
- `npm run build` -> embedded native WebUI rebuilt successfully
- Windows CUDA 12.4 Release `audiocpp_server` build -> passed
- rebuilt server: `build/windows-cuda-release/bin/audiocpp_server.exe`

## Compatibility

- Native WebUI downloads are GGUF-only after this change.
- Existing CLI/server Safetensors loading remains available outside the native model-manager UI.
- Bundled models and catalog entries without a managed download package remain visible.


mirek190:
[0xShug0](https://github.com/0xShug0) You have add qwen 0.6b gguf models to the repo as was only as a safetensor.  So the model is not visible on the WebUi now currently.

mirek190:
Added in `462b0aa`:

| UI area | Change |
|---|---|
| Studio header | Removed the generic “One native server…” slogan. |
| Workflow context | Added a concise, task-specific explanation for Text to speech, ASR, Music generation, Voice conversion, Source separation, Audio analysis, and Voice design. |
| Localization | Added matching descriptions for English, Polish, Italian, Russian, and Chinese. |



Screenshot 2026-08-10 211351

mirek190:
I meant Qwen3-TTS  0.6b

0xShug0:
@mirek190 Qwen3 TTS 0.6B bf16 and q8 have been uploaded to HF!

PR #210, opened 2026-08-11
model: add IndexTTS-2.5 family

IIIIIllllIIIIIlllll:
Just for test now.

Add support for IndexTTS-2.5 (GPT + DiT CFM + BigVGAN, zh/en/ja/es/ar) as a new model family alongside index_tts2:

The architecture of IndexTTS-2.5 has many similarities to IndexTTS-2, but I created a new architecture to avoid interfering with the existing code.

GGUF is also supported (GGUF weights are not uploaded).

Test on my 2080TI gfx1151 CPU

The generated content was consistent with the official deployment method.

About convert:
```
python tools/convert_index_tts2_5.py \
    --model-dir /path/to/IndexTTS-2.5 \
    --output-dir /path/to/staging \
    --native-dir /path/to/IndexTTS-2.5-native \
    --run-converter /path/to/gguf_file --type q8_0
```

AI usage: Kimi k3 developed this implementation based on the original project.

Side note (not fixed in this PR): the existing index_tts2 family produces noise on the CPU backend, same root cause as the 2.5 fix here — src/models/index_tts2/s2mel.cpp builds the wavenet zero accumulator via sub(ctx, input_bct, input_bct) where input_bct is a permuted view; ggml's CPU binary-op kernels only check src1 row-contiguity and silently misread a permuted src0 (asserts are compiled out under NDEBUG; CUDA/HIP kernels handle it, so only CPU is affected). The same one-line materialize-before-sub fix applies. A more thorough fix belongs in ggml: check row contiguity for both operands in the binary-op path (add/sub/mul/div family).


IIIIIllllIIIIIlllll:
@0xShug0  This is a real big project. I'd appreciate it if you could do a quick check to see which parts are reusable. Thanks!

0xShug0:
@IIIIIllllIIIIIlllll Thanks for the work! Based on your code, I think IndexTTS2.5 is close enough to IndexTTS2 that we should extend the existing IndexTTS2 implementation additively instead of copying the whole model into a new family.

The main 2.5 differences seem to be variant-level pieces: multilingual tiktoken/BPE tokenizer + language embedding, CAMPPlus-projected GPT speaker conditioning, changed GPT prefill condition layout, and semantic-codec 2x upsample + up conv. Most other files are identical or near-identical to IndexTTS2 after renaming. Most request/session options seem to be shared with existing IndexTTS2.

Could you refactor this as an IndexTTS2 variant rather than adding copied index_tts2_5 files? Suggested shape:

1. Keep shared audio features, wav2vec2bert, qwen emotion, style encoder, BigVGAN vocoder, and most S2Mel code in the current IndexTTS2 implementation.

2. Add a variant-specific text tokenizer behind one common IndexTTS2 tokenizer interface:
   - v2 keeps the current IndexTTS2 tokenizer behavior.
   - v2.5 uses the multilingual tiktoken/BPE tokenizer, language special tokens, and language-id handling.

3. Split GPT speaker-conditioning into explicit variant modes:
   - v2 mode should preserve the current IndexTTS2 GPT speaker-conditioning path: use speaker semantic features with the existing `conditioning_encoder` + `perceiver_encoder`.
   - v2.5 mode loads and uses `spk_emb_proj` with the CAMPPlus speaker/style embedding, plus `lang_embedding` during text prefill.
   - Keep the shared GPT decode/cache logic in one place.

4. Extend semantic codec with an explicit variant option:
   - v2 keeps the current code embedding path.
   - v2.5 applies the extra 2x nearest upsample and `up` convolution after the decoder projection.
   - The branch should be selected from model config/variant, not by probing tensor names in hot paths.

5. Add 2.5 packages/spec entries under the existing IndexTTS2 family/variant surface.

6. Remove debug-only production code such as dump paths, injected noise files, and environment-variable parity hooks before PR is ready.

This should keep the implementation smaller, reduce maintenance cost, and avoid two copies drifting independently.

IIIIIllllIIIIIlllll:
During the development phase, we encountered a bug: a coverage gap in the English text normalization (TN) module. The same issue affects all backends (my CUDA reproduction also reads it as "DSA").

The official wetext English TN expands digits that are attached to letters: DS4 → "DS four", H264 → "H two hundred sixty four", R2D2 → "R two D two".

Our normalize_english_numbers uses \b(\d+)\b (digits must form a standalone word). In "DS4", the '4' is preceded by the letter 'S', so \b does not match → the raw form is fed to the tokenizer. Since the model was trained on the officially normalized text, when it sees the raw "ds4", it mispronounces it as "D-S-A".

Test:
`I'll start pushing to Github soon. It's still evolving very very quickly. My first run is something akin to antirez' DS4.`

[tts-3a888699.wav](https://github.com/user-attachments/files/30970543/tts-3a888699.wav)

0xShug0:
@IIIIIllllIIIIIlllll I'm going to do some tests and may update the PR directly. Do you have any WIP changes that haven’t been committed yet?

IIIIIllllIIIIIlllll:
> @IIIIIllllIIIIIlllll I'm going to do some tests and may update the PR directly. Do you have any WIP changes that haven’t been committed yet?

No other changes for the time being; you can go ahead and test it.

0xShug0:
> > @IIIIIllllIIIIIlllll I'm going to do some tests and may update the PR directly. Do you have any WIP changes that haven’t been committed yet?
> 
> No other changes for the time being; you can go ahead and test it.

Cool. My pass will foucs on parity and performance. Language normalization has always been an issue in audio.cpp. It’s easy to handle in the UI, while fully porting wetext or other normalization libraries to C++ would distract from the project’s main focus.

0xShug0:
@IIIIIllllIIIIIlllll 

```
Two parity fixes found while investigating the v2.5 ja greedy divergence:

- The repetition penalty was applied to log-softmax scores; since log-probs
  are all <= 0, every seen token was multiplied by the penalty. The official
  pipeline (HF RepetitionPenaltyLogitsProcessor) penalizes the raw logits
  before softmax, dividing positive and multiplying negative ones. Move the
  penalty ahead of the softmax to match.
```
Could you check the ja test case again and explain how you checked the divergence?  I think the official Python code uses a vendored sampler that applies the repetition penalty through `logits_processor` after `log_softmax`. 

I will leave your changes to penalty as they are. I don’t know whether this is a bug in the Python implementation, but it would be great to get to the bottom of it. If changing the penalty order improves parity, it may mean there is another mismatch elsewhere that this change is accidentally offsetting.

```
indextts/infer_v2_5.py
  IndexTTS2.infer()
    -> IndexTTS2.infer_generator()
       sets do_sample/top_p/top_k/temperature/num_beams/repetition_penalty defaults
       calls GPT with do_sample=True

indextts/gpt/model_v2.py
  UnifiedVoice.inference_speech()
    -> self.inference_model.generate(..., logits_processor=...)

indextts/gpt/transformers_generation_utils.py
  GenerationMixin._beam_sample()
    -> next_token_scores = log_softmax(next_token_logits)
    -> next_token_scores_processed = logits_processor(input_ids, next_token_scores)
```

IIIIIllllIIIIIlllll:
> @IIIIIllllIIIIIlllll
> 
> ```
> Two parity fixes found while investigating the v2.5 ja greedy divergence:
> 
> - The repetition penalty was applied to log-softmax scores; since log-probs
>   are all <= 0, every seen token was multiplied by the penalty. The official
>   pipeline (HF RepetitionPenaltyLogitsProcessor) penalizes the raw logits
>   before softmax, dividing positive and multiplying negative ones. Move the
>   penalty ahead of the softmax to match.
> ```
> 
> Could you check the ja test case again and explain how you checked the divergence? I think the official Python code uses a vendored sampler that applies the repetition penalty through `logits_processor` after `log_softmax`.
> 
> I will leave your changes as they are. I don’t know whether this is a bug in the Python implementation, but it would be great to get to the bottom of it. If changing the penalty order improves parity, it may mean there is another mismatch elsewhere that this change is accidentally offsetting.
> 
> ```
> indextts/infer_v2_5.py
>   IndexTTS2.infer()
>     -> IndexTTS2.infer_generator()
>        sets do_sample/top_p/top_k/temperature/num_beams/repetition_penalty defaults
>        calls GPT with do_sample=True
> 
> indextts/gpt/model_v2.py
>   UnifiedVoice.inference_speech()
>     -> self.inference_model.generate(..., logits_processor=...)
> 
> indextts/gpt/transformers_generation_utils.py
>   GenerationMixin._beam_sample()
>     -> next_token_scores = log_softmax(next_token_logits)
>     -> next_token_scores_processed = logits_processor(input_ids, next_token_scores)
> ```

OK，to be continued tomorrow; tonight I'm engaged in a passionate battle with DeepSeek 4 Pro 0813. (lol)

0xShug0:
@IIIIIllllIIIIIlllll Did some optimizations. The current code is ~40.0% faster on CUDA. The 16-bit kvcache is a big win but I  gated it to CUDA only. It may work on other backends. 

```
before: RTF 0.164427  (6.08x)
after: RTF 0.117673 (8.50x)
```
The code for models added in release 0.1 is pretty rough :sweat:  I really need to find time to refactor it.

IIIIIllllIIIIIlllll:
> @IIIIIllllIIIIIlllll Did some optimizations. The current code is ~40.0% faster on CUDA. The 16-bit kvcache is a big win but I gated it to CUDA only. It may work on other backends.
> 
> ```
> before: RTF 0.164427  (6.08x)
> after: RTF 0.117673 (8.50x)
> ```
> 
> The code for models added in release 0.1 is pretty rough 😓 I really need to find time to refactor it.

Regarding performance issues, I also observed something yesterday: on the 2080 Ti, F32-quantized GGUF model performs just as well as F16. I identified the root cause, fixed it, and pushed it to a separate branch, but I don't plan to continue working on it for now.

0xShug0:
@IIIIIllllIIIIIlllll I'm going to merge the PR so 2.5 will be included in release 0.6. Keep me posted if you have any updates on the penalty case.

IIIIIllllIIIIIlllll:
> @IIIIIllllIIIIIlllll I'm going to merge the PR so 2.5 will be included in release 0.6. Keep me posted if you have any updates on the penalty case.

Ok

IIIIIllllIIIIIlllll:
@0xShug0 Re-checked the ja case end-to-end on the reference box. Short version: the pre-softmax penalty is exactly aligned with the official pipeline for the num_beams=1 path — verified token-for-token, not an accidental offset.

**The official penalty order is mode-dependent** (vendored `indextts/gpt/transformers_generation_utils.py`):
- `_sample` (num_beams=1, both greedy and multinomial): penalty on raw logits, pre-softmax (line 3224)
- `_beam_search` (num_beams>1, both beam search and beam sample): log_softmax first, then penalty (lines 3475-3479)

Your trace through `_beam_sample` is the official default config (`num_beams=3, do_sample=True`); my ja divergence investigation ran num_beams=1, which routes to `_sample`.

**One gotcha found while re-testing**: `infer_v2_5.py`'s `infer_generator` hardcodes `do_sample=True` in the `inference_speech()` call — a user-passed `do_sample=False` is popped into a local variable and never used. So all my earlier "official greedy" runs were actually unseeded sampling (which also explains the run-to-run code divergence I saw during the original debugging).

**Definitive check** (ja sentence, wakachi segmentation disabled, num_beams=1, true greedy forced by monkeypatching the `inference_speech` kwargs):
- text tokens: 32/32 identical (incl. `<|ja|>`)
- GPT codes: 99/99 identical, token for token (official emits one extra trailing stop_mel_token that the C++ side excludes)
- flip ablation (penalty moved post-softmax in `_sample`) produces a different sequence, confirming the order does matter
- ASR (whisper) transcribes the C++ output perfectly

One residual divergence to be aware of: the C++ beam path also applies the penalty pre-softmax, while the official beam path applies it post-softmax. So num_beams>1 still differs from the official default. I'd argue the C++ behavior is the semantically correct HF form and the official beam path looks like a quirk of the vendored generation utils — happy to align either way if you prefer bug-for-bug parity there.

0xShug0:
@IIIIIllllIIIIIlllll Thank you for investigating this! Let's just keep your fix for correctness.

PR #211, opened 2026-08-11
Fix CUDA BF16 cuBLAS output on Ampere

mirek190:
## Problem

The embedded GGML BF16 cuBLAS path wrote BF16 inputs directly into a BF16 destination with CUBLAS_GEMM_DEFAULT_TENSOR_OP. On an RTX 3090 (SM 8.6, CUDA 12.4), MiniMax-H3 prompt encoding aborted the server with:

`
CUDA error: the requested functionality is not supported
in ggml_cuda_op_mul_mat_cublas
cublasGemmEx(... CUDA_R_16BF ... CUDA_R_16BF ... CUBLAS_COMPUTE_32F, CUBLAS_GEMM_DEFAULT_TENSOR_OP)
`

This made the native WebUI report only Failed to fetch because the CUDA failure terminated the server process.

## Fix

- Require Ampere or newer before selecting native NVIDIA BF16 GEMM.
- Keep BF16 input operands and FP32 accumulation.
- Write NVIDIA cuBLAS results directly to the existing FP32 destination.
- Use CUBLAS_GEMM_DEFAULT, matching the direction of current upstream GGML behavior.
- Preserve the existing HIP/hipBLASLt BF16-output path and its BF16-to-FP32 conversion.

This avoids a BF16-to-FP16 fallback and therefore does not introduce an unnecessary precision reduction.

## Impact

Ampere CUDA devices such as RTX 30-series cards can execute compatible BF16 matrix multiplications without the unsupported cuBLAS configuration. Ada and newer devices retain native BF16 inputs with FP32 output. HIP behavior is unchanged.

## Validation

- Windows 11
- NVIDIA RTX 3090, compute capability 8.6
- CUDA Toolkit 12.4
- CMAKE_CUDA_ARCHITECTURES=86-real
- Release CUDA build of udiocpp_server: passed
- MiniMax-H3 BF16 prompt-encoder reproducer: passed the former failing cuBLAS operation
- Full 12-step/241-frame CUDA generation, with the separate model-side SM 8.6 attention fallback applied locally:
  - GGUF Q4: 14.74 s, complete 1,728,060-character audio payload
  - GGUF Q4 ConvRot: 14.55 s, complete 1,728,060-character audio payload

The MiniMax SAGE-attention architecture selection is intentionally excluded from this GGML-only PR and will be submitted separately.

### DIFF
--- a/external/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/external/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -1784,7 +1784,7 @@ static void ggml_cuda_op_mul_mat_cublas(
                                  src1_ptr,       CUDA_R_16BF, ne10,
                     &beta_f32,   dst_dd_i,       CUDA_R_32F,  ldc,
                     CUBLAS_COMPUTE_32F,
-                    CUBLAS_GEMM_DEFAULT));
+                    CUBLAS_GEMM_DEFAULT_TENSOR_OP));
 #endif // defined(GGML_USE_HIP) && defined(GGML_HIP_USE_HIPBLASLT)
     } else if (fast_fp16_hardware_available(cc) && use_fp16) {
         // convert src0 and src1 to fp16, multiply as fp16, convert dst to fp32
--- a/external/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/external/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -1739,7 +1739,8 @@ static void ggml_cuda_op_mul_mat_cublas(
 
     const int cc = ggml_cuda_info().devices[id].cc;
 
-    const bool supports_bf16 = GGML_CUDA_CC_IS_NVIDIA(cc) || GGML_CUDA_CC_IS_AMD(cc) ||
+    const bool supports_bf16 =
+        (GGML_CUDA_CC_IS_NVIDIA(cc) && cc >= GGML_CUDA_CC_AMPERE) || GGML_CUDA_CC_IS_AMD(cc) ||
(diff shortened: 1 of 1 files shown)


0xShug0:
@mirek190 The direction looks right and aligns with the latest upstream ggml . One small question: upstream still passes `CUBLAS_GEMM_DEFAULT_TENSOR_OP`, while this patch uses `CUBLAS_GEMM_DEFAULT`. Was that intentional/measured? If not, matching upstream may be safer for BF16 Tensor Core performance.

mirek190:
Good catch. CUBLAS_GEMM_DEFAULT was a conservative diagnostic choice while isolating the original Ampere failure; it was not selected from a performance comparison.

I rechecked current upstream ggml (30bf8685): NVIDIA BF16 uses an FP32 destination while retaining CUBLAS_GEMM_DEFAULT_TENSOR_OP. I have now matched that behavior in 180e06.

I also reran the full MiniMax-H3 CUDA reproducer on RTX 3090 / SM 8.6 / CUDA 12.4 with CUDA_LAUNCH_BLOCKING=1:

- 12 inference steps, Spectrum acceleration
- 241 requested frames / 243 aligned frames
- completed successfully
- 1,728,060-character audio payload
- no CUBLAS_STATUS_NOT_SUPPORTED / 
equested functionality is not supported failure

So the unsupported combination was the BF16 destination with Tensor Op mode. BF16 inputs + FP32 accumulation/output + Tensor Op mode works and preserves the upstream Tensor Core selection.

0xShug0:
@mirek190 Merged! Thanks!

PR #214, opened 2026-08-11
Fix CUDA linking on Nixos

utsl42:
Fixes a problem building with CUDA from the Nix flake. Without it, the build errors with:

```
$ ./result/bin/audiocpp_cli --list-devices
ggml_cuda_init: failed to initialize CUDA: CUDA driver is a stub library
available_devices=1
CPU:0 "13th Gen Intel(R) Core(TM) i7-13700" [CPU]
select with: --backend <cuda|hip|vulkan|metal|cpu> --device <index>
```

With it my GPU is detected.

Fix based on how the `llama-cpp` package in Nixos does it:
https://github.com/NixOS/nixpkgs/blob/ef33b28e97233dac04666e34f26027c44f5db08b/pkgs/by-name/ll/llama-cpp/package.nix#L115

### DIFF
--- a/.devops/nix/package.nix
+++ b/.devops/nix/package.nix
@@ -14,6 +14,7 @@
   python-scripts,
   config,
   version,
+  autoAddDriverRunpath,
 
   # Overridable feature flags
   cudaSupport ? config.cudaSupport or false,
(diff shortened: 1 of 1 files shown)


0xShug0:
@utsl42 Merged. Thanks!

PR #226, opened 2026-08-13
server: accept base64-encoded voice reference in speech requests

IIIIIllllIIIIIlllll:
Add base64 support to the speech API. During use, I noticed that every time I clone a voice, I have to upload the audio file to the server first, and then use the file path for audiocpp_server to read it—which is quite inflexible. Therefore, I plan to add this feature so that clients can directly upload audio files encoded in base64.

Move base64 helpers into app/server/base64.{h,cpp}, add a strict decoder, and cover it with a server_base64_test unittest.

The maximum file size needs to be considered—I'm concerned that users might inadvertently upload extremely large audio files, causing machines with limited memory to run out of memory (OOM) directly.

What do you think?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1399,6 +1399,7 @@ configure_file(
 
 add_executable(audiocpp_server
     app/server/main.cpp
+    app/server/base64.cpp
     app/server/config.cpp
     app/server/http.cpp
     app/server/model_installer.cpp
@@ -2143,6 +2144,17 @@ if (ENGINE_BUILD_TESTS)
     add_test(
         NAME server_multipart_test
         COMMAND server_multipart_test
+    )
+
+    add_executable(server_base64_test
+        tests/unittests/test_server_base64.cpp
+        app/server/base64.cpp
+    )
+    target_include_directories(server_base64_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/app/server)
+
+    add_test(
+        NAME server_base64_test
+        COMMAND server_base64_test
     )
 
     add_executable(server_busy_guard_test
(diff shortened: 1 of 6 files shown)


0xShug0:
@IIIIIllllIIIIIlllll Thanks! I like the feature. For reference voice, models usually recommands 5 to 15s. We can assume a safe max is 30s (base64 < 5MB).

I wonder if we can avoid adding a separate top-level `voice_ref_b64` option by overloading the `voice_ref` option? 

```
"voice_ref": {
  "type": "base64",
  "data": "UklGRh..."
}

"voice_ref": {
  "type": "path",
  "path": "voices/alice.wav"
}

"voice_ref": "voices/alice.wav"
```

IIIIIllllIIIIIlllll:
> I wonder if we can avoid adding a separate top-level `voice_ref_b64` option by overloading the `voice_ref` option?

OK, I think we can distinguish between a file path and base64-encoded data by checking the beginning of the string. 

Edited: 
Your method is more reliable after all.

0xShug0:
@IIIIIllllIIIIIlllll  Sorry, just for clarification: `"voice_ref": "voices/alice.wav" `is backward-compatible, so existing users can continue using it, or they can use:

```
"voice_ref": {
  "type": "path",
  "path": "voices/alice.wav"
}
```
And for Base64, the user must explicitly send it.

```
"voice_ref": {
  "type": "base64",
  "data": "UklGRh..."
}
```

So you don't need to distinguish between them based on prefix. 

PS: In fact, we can’t prevent users from sending random stuff.  Comprehensive checks are better handled in the frontend or gateway before requests reach the HTTP server so the server remains lightweight.

Issue #227, opened 2026-08-13
Core dump while inferencing with old GPU (CUDA_ARCHITECTURES=61)


jasonchen31:
Hi,

After clone the latest version of audio.cpp, all gguf models (STT, TSS) no longer work on old Nvidia GPU (CUDA_ARCHITECTURES=61, CUDA, Driver Version: 535.54.03, CUDA Version: 12.2). Core dumped always.

## Environment

audio.cpp version: Release 0.6 WIP (d4d684b41286db7baf83f5a9ee17cc8fd6ed3610)
Nvidia container: nvidia/cuda:12.4.1-devel-ubuntu22.04
g++ (conda-forge gcc 12.4.0-2) 12.4.0
CUDA 12.2
Driver Version: 535.54.03
GTX 1080TI (11G)
Linux (Nvidia container, Ubuntu 22.04.1)

## Build commands
1. ./scripts/build_linux.sh --backend cuda
2. CMAKE_CUDA_ARCHITECTURES=61 ./scripts/build_linux.sh --backend cuda
3. Manually
  cmake -S . -B build -DENGINE_ENABLE_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=61 
  cmake --build build --config Release -j

## Observation
1. All binary built from the above 3 build commands failed with the same core dump
2. At first the building process failed due to sage attention introduction
3. After commits "Gate CUDA SageAttention by architecture" and "Fix HIP build without Sage Attention 2 (#217)", built successfully. But none of the binary work with my old GPU. I guess it was caused by the SAGE Attention 2 relevant updates, even it is gated during compilation phase.
4. Core dump while perform inference.

## Execution Log
./audiocpp_server --ui --ui-management --host 0.0.0.0
native WebUI model root: "/workspace/audio.cpp/models"
native WebUI package resources: "/workspace/audio.cpp"
audiocpp_server listening on http://0.0.0.0:8080
ggml_cuda_init: found 1 CUDA devices (Total VRAM: 11172 MiB):
  Device 0: NVIDIA GeForce GTX 1080 Ti, compute capability 6.1, VMM: yes, VRAM: 11172 MiB
ggml_cuda_graph_set_enabled: disabling CUDA graphs due to GPU architecture
ggml_cuda_compute_forward: CPY failed
CUDA error: the provided PTX was compiled with an unsupported toolchain.
[...]

PR #231, opened 2026-08-13
Fix core dump issue for binary built for old GPUs (cm<89)

jasonchen31:
Fixing https://github.com/0xShug0/audio.cpp/issues/227

## Problem

Two bugs broke CUDA builds targeting **old GPUs (compute capability 6.1, e.g. Pascal GTX 10xx)**:

1. **SageAttention2 is unconditionally compiled and can't build for old GPUs.**
   The vendored ggml CUDA backend globs `sage-attn2.cu` into every build. Its
   headers include `<cuda/barrier>` / `<cuda/pipeline>` unconditionally, which
   CUDA hard-errors on sub-sm_70 targets:

   ```
   #error "CUDA synchronization primitives are only supported for sm_70 and up."
   ```

2. **`scripts/build_linux.sh --backend cuda` never sets the target GPU architecture.**
   It silently inherits `CMAKE_CUDA_ARCHITECTURES=native` (the *build machine's*
   GPU). Building on any modern GPU produced a single-arch image with no sm_61
   code, and left the Sage gate "ON" (arch >= 89), so kernels with SM89+
   instructions were packaged into a binary that core-dumped when run on a 6.1 GPU.

A direct `cmake -S . -B build -DENGINE_ENABLE_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=61`
worked because it pins the arch (and 61 < 89 disables Sage at compile time).

### `external/ggml/src/ggml-cuda/CMakeLists.txt` (build gate)

Add a fail-closed build-time gate for SageAttention2:

- Default `ON`, but turned `OFF` if **any** `CMAKE_CUDA_ARCHITECTURES` entry
  parses to `< 89` (or is unparseable, e.g. `native`).
- When `OFF`, `sage-attn2.cu` is removed from the source list and a
  `GGML_CUDA_SAGE_ATTN2_ENABLED` compile definition is **not** added; a
  `message(STATUS ...)` reports `CUDA SageAttention2 disabled`.

### `external/ggml/src/ggml-cuda/sage-attn2.cuh` (stub definitions)

Because `ggml-cuda.cu` unconditionally references the four Sage entry points
(`ggml_cuda_sage_attn2`, `_i8`, `_supported`, `_i8_supported`), the header now:

- keeps the real declarations when `GGML_CUDA_SAGE_ATTN2_ENABLED` is defined
  (arch >= 89), and
- provides `static inline` stubs when it is not: `supported()` returns `false`
  (op never attached to a graph) and the compute entry points `GGML_ABORT`.
  Since `ggml-cuda.cu` already includes this header, no dispatcher edits are
  needed.

### `scripts/build_linux.sh` (target architecture control)

- New `--cuda-arch <spec>` flag that forwards `-DCMAKE_CUDA_ARCHITECTURES=<spec>`
  (with `-UCMAKE_CUDA_ARCHITECTURES` to avoid the sticky cache entry, mirroring
  the existing HIP `--gpu-targets` handling).
- Status output prints the effective CUDA architectures (or
  `<auto: machine-native ...>` when unset).

### Docs

- `docs/build/linux.md`: document `--cuda-arch`, including the
  old-GPU example and the equivalence with the working direct CMake invocation.
  Noted that `ENGINE_ENABLE_NATIVE_CPU=OFF` is **not** required, since
  `GGML_NATIVE` affects CUDA arch only as a fallback when
  `CMAKE_CUDA_ARCHITECTURES` is undefined.

## Why this is safe for correctness

- Disabling Sage does **not** leave MiniMax-H3 (the only Sage consumer) without
  an attention path: `src/community_models/minimax_h3/dit_denoiser.cpp` already
  falls back to `ggml_flash_attn_ext` when the Sage op is unsupported.
- Flash attention has a real Pascal (cc 610) kernel path
  (`BEST_FATTN_KERNEL_TILE` in `external/ggml/src/ggml-cuda/fattn.cu`), which is
  compiled into the sm_61 build.
- Sage physically cannot run below SM89 anyway; the runtime `supported()`
  guard already returns `false` for cc < 890, so a 6.1 GPU would have used flash
  attention regardless. The real fix is compiling the whole backend for sm_61.

## Usage

```bash
# Old GPU (compute capability 6.1)
./scripts/build_linux.sh --backend cuda --cuda-arch 61 --target audiocpp_cli --target audiocpp_server
```

## Validation status

- `git diff` for CMakeLists.txt is a pure additive change (22 insertions /
  0 deletions); CRLF line endings of the vendored file are preserved.
- Verified locally via a `cmake -P`-equivalent simulation that the gate resolves
  correctly: `61`/`52`/mixed-low/`native` -> Sage OFF; `89-real`/`120a-real` -> ON.


jasonchen31:
I compiled again while fixing the SensVoice ASR Webui missing issue, Sage Attention was somehow included and the build failed again. Still working on it.

jasonchen31:
It should be fixed this time. clean build success, and GPU inference correctly.

0xShug0:
@jasonchen31 The changes are good and work correctly on Linux. Could you rebase onto the current `main` and resolve the conflicts? `main` is the correct target now. Once it rebases cleanly, we should be good to go.

jasonchen31:
rebased with https://github.com/0xShug0/audio.cpp/pull/240

jasonchen31:
we may close this with https://github.com/0xShug0/audio.cpp/pull/240

PR #237, opened 2026-08-14
Document macOS CPU-only builds

Blakeolson21:
## What changed

- Add CMake to the macOS source-build prerequisites.
- Clarify that the Metal compiler is only required for Metal builds.
- Add an explicit CPU-only CMake recipe for macOS. Apple builds otherwise enable Metal by default.
- Add a device-list check so users can confirm that the resulting binary sees the CPU backend.

## Validation

Tested on an Apple M3 Max with macOS 26.3.1, Xcode 26.3, AppleClang 17.0.0, and CMake 4.4.2.

The documented configure command completed in 8.89 seconds. The documented build command completed in 55.76 seconds and produced arm64 `audiocpp_cli`, `audiocpp_server`, and `audiocpp_gguf` binaries.

`audiocpp_cli --list-devices` reported:

```text
available_devices=2
BLAS:0 "Accelerate" [ACCEL]
CPU:0 "Apple M3 Max" [CPU]
```

The loader/catalog checks also passed:

```text
active_loaders=49 commented_loaders=0 specs=47 packages=140 manager_packages=140
ok: runtime loaders, model_specs, and model_manager_v2 are in sync
```


0xShug0:
Thanks @Blakeolson21! Merged.

PR #240, opened 2026-08-14
merge for: Fix core dump issue for binary built for old GPUs (cm<89)- #231

jasonchen31:
This is a new PR to rebased https://github.com/0xShug0/audio.cpp/pull/231 to current main. 
As the following files are already merged:
1. external/ggml/src/ggml-cuda/sage-attn2.cuh
2. external/ggml/src/ggml-cuda/CMakeLists.txt
So this PR only update build_linuix.sh to support `--cuda-arch <archi id>` and add some guidance.


0xShug0:
@jasonchen31 Merged! Thanks!

PR #245, opened 2026-08-15
Honor required request options in native Studio

JoeMattie:
## Summary

- propagate required request-option metadata from model specs into native Studio
- label required lyrics and reference transcripts correctly
- reject blank required fields before model loading or task submission
- regenerate the embedded native UI bundle

## Why

MiniMax Music3 declares lyrics as required and rejects empty lyrics at runtime, but Studio labeled lyrics optional and omitted the field when blank. That made prompt-only submissions fail at the server. Studio now uses model-spec metadata as the source of truth, which also honors GLM-TTS's required reference transcript.

## Port notes

This ports the applicable review fix from `717fade` onto the implementation merged by #243. The other findings are already satisfied or no longer present on this branch: it uses a portable `kTwoPi`, retains runtimes when `mem_saver=false`, publishes a valid 32 MiB weight-context default with `min: 1`, and uses the file-based `audiocpp_gguf` converter instead of a raw stdio helper.

## Validation

- `npm run check`
- `npm run build`
- `cmake --build build/linux-cuda-release --target engine_model_minimax_music3 --parallel 4`
- `python3 tools/check_loader_catalog_sync.py --self-test`
- `python3 tools/check_loader_catalog_sync.py`


0xShug0:
@JoeMattie Merged, Thanks!

PR #247, opened 2026-08-15
Fix: IndexTTS-2 F16 weight storage

IIIIIllllIIIIIlllll:
While running IndexTTS on the 2080 Ti, I noticed that the performance of the F16‑quantized GGUF and the F32‑quantized GGUF was almost identical, so I made a fix to address the performance difference.

Here is the comparison before and after the fix:

| Stage | Before fix (f16) | After fix (f16) | Improvement |
|-------|------------------|-----------------|-------------|
| GPT decoding | 3126 ms | 2555 ms | –18.3% |
| CFM | 2617 ms | 1930 ms | –26.3% |
| BigVGAN | 982 ms | 742 ms | –24.4% |
| Total | 6889 ms | 5393 ms | –21.7% |

Need your input: should we go ahead with this change?

### DIFF
--- a/src/framework/modules/vocoders/bigvgan_vocoder.cpp
+++ b/src/framework/modules/vocoders/bigvgan_vocoder.cpp
@@ -138,7 +138,6 @@ Conv1dWeights load_weight_norm_conv1d(
     int64_t dilation,
     bool use_bias,
     assets::TensorStorageType storage_type) {
-    (void) storage_type;
     const auto weight_v = source.require_f32(prefix + ".weight_v", {out_channels, in_channels, kernel});
     const auto weight_g = source.require_f32(prefix + ".weight_g", {out_channels, 1, 1});
     Conv1dWeights out;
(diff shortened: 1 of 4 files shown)


0xShug0:
@IIIIIllllIIIIIlllll This is a very good improvment, but I need some time to do regression test for the existing 5 consumers of BigVGAN.

0xShug0:
@IIIIIllllIIIIIlllll Could you test the audio quality change before and after the PR for indextts 2/2.5 using the q8 gguf?

IIIIIllllIIIIIlllll:
> @IIIIIllllIIIIIlllll Could you test the audio quality change before and after the PR for indextts 2/2.5 using the q8 gguf?

No problem. Are there any criteria for the evaluation?

0xShug0:
> > @IIIIIllllIIIIIlllll Could you test the audio quality change before and after the PR for indextts 2/2.5 using the q8 gguf?
> 
> No problem. Are there any criteria for the evaluation?

Just human judgment :). On CUDA Q8 before and after the patch has log-mel similarity and frame-count drift (while F16 is identical). But the spoken words are correct, and the emotion sounds good. I don’t know whether the BigVGAN change will affect the behavior for quantized types on HIP. If they sound good to you, I'm ready to merge.

IIIIIllllIIIIIlllll:
@0xShug0 I can clearly perceive the difference between F16 and Q8 — it’s a fleeting, extremely jarring yet indescribable sensation, a difference that only becomes noticeable upon careful comparison. However, there is virtually no difference between Q8(main branch) and Q8(this pr). (I used to be really into HiFi.)

0xShug0:
@IIIIIllllIIIIIlllll Merged!

PR #257, opened 2026-08-16
Retire stale WebUI service workers on startup

mirek190:
## Summary

Fix the embedded Native WebUI remaining stuck on a previously installed application when audio.cpp reuses the same local origin, commonly `http://127.0.0.1:8080`.
**In short - If you had other webpage on http://127.0.0.1 and port 8080 loaded before the cache was not released and you could nor see Audio.cpp Studio on working server until you do not clean cache manually.** 

## Root cause

A prior llama.cpp WebUI registered a root-scoped `/sw.js` Workbox service worker. That worker continued serving its cached application shell before requests reached `audiocpp_server`, so normal Edge sessions displayed the old UI while InPrivate sessions correctly displayed Native Studio. Ordinary HTTP cache headers alone cannot retire a service worker that intercepts the navigation first.

## Changes

- Serve no-store/no-cache headers and `Clear-Site-Data: cache` for the embedded UI shell.
- Provide retirement workers at `/sw.js` and `/service-worker.js` when the UI is enabled.
- The retirement worker replaces a stale registration, clears Cache Storage, unregisters itself, and reloads controlled windows.
- Native Studio also unregisters any remaining workers and clears Cache Storage during startup.
- Deliberately preserve `localStorage` and IndexedDB so saved voices, model-folder selection, language, and UI preferences survive.
- Regenerate the embedded SvelteKit output.

## Validation

- `npm run check`: 0 errors and 0 warnings.
- `npm run build`: successful static Native Studio build.
- Windows CUDA Release `audiocpp_server` build: successful.
- Verified `/` returns `Cache-Control: no-store, no-cache, must-revalidate, max-age=0` and `Clear-Site-Data: cache`.
- Verified `/sw.js` returns the no-store retirement worker with root service-worker scope.
- Browser regression: a synthetic legacy Cache Storage entry was removed after reload while a `localStorage` preference remained intact; no service-worker registrations remained.
- Reproduced with a normal Microsoft Edge profile containing llama.cpp's old `/sw.js` and Workbox cache; reopening `127.0.0.1:8080` automatically recovered to `audio.cpp · Native Studio`. InPrivate mode was not required.


0xShug0:
Merged! Thanks!

PR #258, opened 2026-08-16
Clarify ACE-Step GGUF package labels

mirek190:
## What changed

Updates the native WebUI labels for all six ACE-Step GGUF packages so users can distinguish Turbo, Base, XL Turbo, and XL SFT variants at a glance.

The two 14.2 GB packages are explicitly identified as **GGUF Turbo XL BF16** and **GGUF Turbo XL SFT BF16**. Base packages keep the concise **GGUF Q8/BF16** labels, while regular Turbo packages use **GGUF Turbo Q8/BF16**.

The embedded static WebUI asset was regenerated so the labels ship inside audiocpp_server.

The shared compact package-button height fix is already present in main through PR #257 and is therefore not duplicated here.

## Validation

- npm run check: 0 errors and 0 warnings
- npm run build: passed


0xShug0:
@mirek190 Little confused about this PR. It also has a commit "Normalize multi-variant package button heights", while PR https://github.com/0xShug0/audio.cpp/pull/257 (just merged) also has the commit of the same name? Are they different?

mirek190:
the same ... I messed a bit here ;) 
No difference anyway in the height buttons.


Here are fixed lables.

0xShug0:
Thanks @mirek190, merged!

PR #262, opened 2026-08-17
Accept a recognition prompt on multipart transcriptions

### LINKED ISSUE #261: Multipart /v1/audio/transcriptions silently drops the recognition prompt
`build_openai_transcription_request()` reads a `text` field into `request.text_input`, and `qwen3_asr/prompt_asr.cpp` passes `request.context` to `tokenizer_.build_prompt()` — so Qwen3-ASR recognition-context biasing works end to end on the JSON path.

`handle_transcription_multipart()` can't reach it. It builds its own `fields` object from the parsed parts and only ever emplaces `model` and `language`, so a multipart client has no way to bias recognition. Since unrecognised parts are ignored, the caller gets HTTP 200 and an unchanged transcript — the failure is invisible.

Multipart is the path the OpenAI SDKs and every `curl -F` example use, and OpenAI's `/v1/audio/transcriptions` has a `prompt` field for exactly this purpose.

### Repro

Qwen3-ASR-0.6B, CUDA. Speech saying *"Tell me about cell-free systems"*:

```
curl -s :8816/v1/audio/transcriptions -F file=@clip.wav -F model=qwen3-asr \
     -F language=en -F prompt='cell-free systems, plasmid, Benchling'
{"text":"Tell me about self-free systems."}      # prompt had no effect
```

The same audio and the same context through the JSON path transcribes "cell-free" correctly, so the model and the request plumbing are fine — only the multipart handler drops it.

### Expected

A `prompt` part on the multipart endpoint biases recognition the same way `text` does on the JSON path.

derekja:
## Summary

- Accept a `prompt` (or `text`) part in `handle_transcription_multipart` and forward it as `text` to `build_openai_transcription_request`, which already routes it to `request.text_input`.
- `prompt` matches OpenAI's `/v1/audio/transcriptions` field name; `text` is accepted so the multipart and JSON paths take the same spelling.
- Document the accepted fields in the server usage line.

Fixes #261

## Validation

Built `audiocpp_server` and ran it on a spare port beside the unmodified one. CUDA, Qwen3-ASR-0.6B, H100 12GB slice. Identical audio, identical request except the field:

| spoken | without `prompt` | with `prompt` |
|---|---|---|
| Tell me about cell-free systems. | Tell me about **self-repair** systems. | Tell me about **cell-free** systems. |
| How do cell-free systems help medicine in space? | How do **self-repair** systems help medicine in space? | How do **cell-free** systems help medicine in space? |

Context used: `cell-free systems, cell-free protein synthesis, cell lysate, PURE system, plasmid, Benchling, Opentrons, ...`

Sanity check that the context genuinely reaches `build_prompt`: with `prompt="cell-free systems"` over **one second of random noise**, the server returns `{"text":"Cell-free systems."}`. Without the field, the same audio returns empty.

Behaviour is unchanged when the field is absent (left column above) and on the JSON path.

## Note — not fixed here

The context is built into the thinker prefill graph, so its size costs VRAM. On a 12GB slice a prompt over roughly 1200 characters fails the whole request:

```
ggml_backend_cuda_buffer_type_alloc_buffer: allocating 13559.65 MiB on device 0: cudaMalloc failed: out of memory
alloc_tensor_range: failed to allocate CUDA0 buffer of size 14218318336
500 {"error":{"message":"failed to allocate Qwen3 ASR thinker prefill graph","type":"server_error"}}
```

That is pre-existing and applies to the JSON path too, but a 400 with a clear message would beat a 500 — a caller currently discovers the ceiling by taking transcription down. Happy to follow up separately if that would be welcome.

### DIFF
--- a/app/server/main.cpp
+++ b/app/server/main.cpp
@@ -94,6 +94,7 @@ void print_help() {
         << "  GET  /v1/audio/voices?model=<id>\n"
         << "  POST /v1/audio/speech\n"
         << "  POST /v1/audio/transcriptions\n"
+        << "       fields: file, model, language, prompt, stream\n"
         << "       OpenAI-style streaming: speech stream_format=sse|audio, transcription stream=true\n"
         << "  POST /v1/audio/transcriptions/live?model=<id>\n"
         << "       raw PCM in a chunked body, transcript deltas as SSE on the same connection\n"
--- a/app/server/runtime.cpp
+++ b/app/server/runtime.cpp
@@ -1955,6 +1955,7 @@ HttpResponse ServerState::handle_transcription_multipart(const std::string & bod
     const MultipartPart * file_part = nullptr;
     std::string model_id;
     std::string language;
+    std::string prompt;
     std::optional<int> busy_timeout_ms;
     bool stream = false;
     for (const auto & part : parts) {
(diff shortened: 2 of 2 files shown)


0xShug0:
Thanks @derekja!  I will do some tests quickly.

0xShug0:
@derekja Merged. A follow-up PR on the 500 error would be very welcome!

PR #264, opened 2026-08-17
Report over-large requests as 400, not 500

derekja:
## Summary

Follow-up to #262, as suggested there.

Some rejections that are the caller's to fix surface as `500 server_error`. The message names an internal graph or cache, so there is nothing actionable in it, and the caller goes looking for a server fault that isn't there.

- Add `engine::runtime::CapacityError` for "this device or model cannot serve a request *of this size*" — distinct from a genuine fault, because the caller can fix it by sending less.
- Convert two sites across two engines: the Qwen3-ASR prefill allocation, and VoxCPM2's prompt and generation cache-length checks. Each message now carries the numbers, so the remedy is arithmetic rather than guesswork.
- Map it to `400 invalid_request_error` in the server dispatch.

This follows the `ServerBusyError` → 503 precedent: the layer that knows the condition throws a typed exception, the server decides the status. `CapacityError` lives in the engine runtime headers because the throw sites are under `src/models/` and the mapping is in `app/server/` — both already include `engine/framework/*`.

**Deliberately not a sweep.** Around 30 models throw `failed to allocate`, but they are not uniform — several allocate model weights at load time, which is a genuine fault and should stay a 500. Whether a given site is caller-size-driven needs per-site judgement, so this converts only the two engines I could exercise end to end on real hardware. The type and the mapping make the rest mechanical for whoever knows each model.

## Validation

Two servers side by side, CUDA / H100 12GB slice, one at `main` and one patched.

| engine | request | `main` | this branch |
|---|---|---|---|
| Qwen3-ASR | no prompt | 200 | 200 |
| Qwen3-ASR | 400-char prompt | 200 | 200 |
| Qwen3-ASR | 4000-char prompt | `500 server_error`<br>`failed to allocate Qwen3 ASR thinker prefill graph` | `400 invalid_request_error`<br>`Qwen3 ASR prefill graph does not fit in device memory at this size (1173 prompt steps, of which 13 are audio tokens); shorten the transcription prompt or the audio` |
| VoxCPM2 | 60-char text | 200, 506924 B pcm | 200, **506924 B pcm** (identical) |
| VoxCPM2 | 20000-char text | `500 server_error`<br>`VoxCPM2 generation exceeds model cache length` | `400 invalid_request_error`<br>`VoxCPM2 generation exceeds the model cache length (4104 prefill rows + 4096 requested tokens, limit 8192); shorten the input text` |

The VoxCPM2 numbers are the point: `4104 + 4096 = 8200` against a limit of `8192` tells the caller exactly how much to cut.

`CapacityError` is thrown from three sites in two files, so no existing 500 changes meaning elsewhere.

### DIFF
--- a/app/server/runtime.cpp
+++ b/app/server/runtime.cpp
@@ -11,6 +11,7 @@
 #include "engine/framework/debug/trace.h"
 #include "engine/framework/io/json.h"
 #include "engine/framework/model_spec/metadata.h"
+#include "engine/framework/runtime/errors.h"
 #include "engine/framework/runtime/registry.h"
 
 #include <algorithm>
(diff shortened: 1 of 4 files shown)


0xShug0:
@derekja Merged. Thanks!

PR #275, opened 2026-08-18
Add F5-TTS community model (M0 scaffolding; Habibi Arabic aliases)

tareko:
**M2/M3 landed: verified DiT parity + full inference pipeline.**

## Parity (M2) — all stages at cosine 1.000000

Stage-by-stage comparison against golden tensors generated from the real `f5_tts` PyTorch module running the actual Habibi EMA weights (goldens themselves validated at cosine 0.9997):

| Stage | cosine |
|---|---|
| text embed (lookup) | 1.000000 |
| text ConvNeXt x4 | 0.999999 |
| text padded | 0.999999 |
| input embed (proj + CPE) | 1.000000 |
| time embed | 1.000000 |
| DiT block 0 / 21 | 1.000000 |
| final out (proj_out) | 1.000000 |

Parity harness: `tests/f5_parity_main.cpp` (golden generator + instructions in `docs/community_models/f5_tts.md` follow-up).

Non-obvious ggml details worth review:
- ne0 is the fastest axis: column-layout `[features, time]` throughout; conv weights loaded raw in torch order `[out, in/g, k]`
- `ggml_mean` on `[1, N]` is the identity (row-wise over ne0) — GRN needs `sum + scale` for a true scalar mean
- RoPE applies on `[DH, heads, N]` with positions at ne2, `GGML_ROPE_TYPE_NORMAL` = x_transformers interleaved pairs, theta 10000 — then permute to `[DH, N, H]` for `ggml_flash_attn_ext`
- Grouped conv1d (groups=16, k=31) lowered to per-group `ggml_im2col` + matmul

## Inference (M3)

`f5_synthesize` (src/community_models/f5_tts/synthesize.cpp):
- Log-mel frontend bit-exact vs torchaudio MelSpectrogram (htk scale, norm=None, power=1, reflect-padded center STFT, `1 + floor(len/hop)` frames) — verified cosine 1.000000, maxabs 0.0
- Char tokenizer with Habibi dialect tokens (⓪-⑫ + 〈...〉)
- CFM Euler sampler with sway sampling schedule (coef -1) and classifier-free guidance (cond/uncond pair)
- Vocos vocoder in pure C++: backbone (embed conv + 8 ConvNeXt + final LN) verified exact against torch forward hooks; ISTFT head with exp/clipped magnitudes; roundtrip mel-correlation 0.9963 — identical to torch vocos itself (waveform correlation is near zero for any vocoder due to phase regeneration; mel-space is the correct metric)

## E2E

Arabic text → 24 kHz WAV via the pure C++ stack; verified by Whisper round-trip transcription matching the input text (loanwords audio.cpp/GPU rendered recognizably).

## Known limits (M4 roadmap)

- CPU-only for now (~33x RTF on an i7-3930K); CUDA backend wiring is straightforward since all weights go through BackendWeightStore
- Long texts: duration capped at 1024 frames; needs F5-style text chunking + crossfade
- Session wiring to the audiocpp_server HTTP API pending
- Reference audio must currently be 24 kHz mono float (resampler is linear)

## CUDA (new)

`F5ComputeDevice` selects CPU threads or a CUDA device index. Weights upload to a CUDA `BackendWeightStore`; the graph runs via gallocr + `compute_backend_graph` with staged leaf uploads and `tensor_get` readback. Model cache and backends intentionally leak at exit (CUDA buffers cannot be freed after driver shutdown during static destruction).

- Parity on GPU: all 8 stages cosine 0.999995–1.0 (CPU unchanged at 1.0)
- E2E on an RTX 3090: 5.37 s audio in 14.3 s = **2.67x RTF**, vs 33x on CPU — 12x speedup
- The CUDA build also surfaced a latent CPU bug (partially-initialized GELU ones tensor), now fixed

## Performance work (new): 0.75x RTF — faster than real time

Three staged optimizations over the CUDA port (2.67x RTF):

| Stage | Wall (5.37s clip) | RTF |
|---|---|---|
| CPU (12 threads) | 177s | 33x |
| CUDA, per-call graphs | 14.3s | 2.67x |
| + DiT graph reuse | 9.5s | 1.77x |
| + GPU Vocos vocoder | **4.0s** | **0.75x** |

1. **DiT graph reuse** (roformer `FixedShapeGraph` pattern): graph + gallocr allocation + constants built once per (model, N, NT, taps); each Euler step uploads only 4 leaves (x, cond, text_ids, time-embedding). Both CFG passes share one graph.
2. **Vocos as a ggml graph** (`vocos_decode_gpu`): cached per (frames, device); verified cosine 0.999987 against the torch-exact host implementation. Host keeps only the O(n) ISTFT tail.
3. **CUDA graphs** (`-DGGML_CUDA_GRAPHS=ON`): verified via nsys — 1 instantiate + 31 replays per synthesis. The DiT is now compute-bound (cutlass GEMMs dominate), so this mostly future-proofs launch overhead.

GPU 1 (RTX 3090) at 99% util, ~7.7 GiB VRAM, coexisting with the existing qwen3-tts container on the same box. Whisper round-trip transcription identical before/after all three changes.

## Batched CFG + measurement-driven notes (new)

- **Batched CFG**: the conditioned/unconditioned pair now runs as ONE ne3=2 graph (`f5_dit_forward_cfg`) — verified cosine 1.000000 on both halves against the sequential path. Halves launches; weights read once per step.
- **FP16 weights** (`F5ComputeDevice::fp16_weights`): parity 0.999963, but *measured slower* end-to-end on RTX 3090 (4.9s vs 4.0s) — ggml converts F32 activations to F16 per GEMM and that overhead outweighs tensor-core gains at F5's GEMM sizes. Off by default; the honest data is in the commit message.
- **Duration bucketing** (64-frame multiples): graphs + CUDA captures survive across requests.

**Final numbers** (5.37s Arabic clip, one RTX 3090 sharing with the production stack):
- cold: 4.2s (one-time ~2.6s graph build)
- warm/steady-state: **1.68s = 0.31x RTF** (DiT 98ms/step, Vocos 44ms)
- Whisper round-trip identical throughout

## Review feedback implemented (module reuse)

Per @0xShug0's [comment](https://github.com/0xShug0/audio.cpp/pull/275#issuecomment-5336602093): the DiT is now composed from framework modules following the dev-branch patterns — `LinearModule`, `LayerNormModule`, `EmbeddingModule`, `DepthwiseConv1dModule`, `GeluModule`, `SiluModule`, `RoPEModule`, `ScaledDotProductAttentionModule` (Flash lowering), and the structural/primitive modules — with module-typed weights (`LinearWeights` etc.) loaded through `BackendWeightStore` like parakeet_tdt/roformer. Only ops with no framework equivalent remain local helpers: grouped conv1d (lowered exactly like `Conv1dModule`'s own im2col+mul_mat), GRN, adaLN modulate, and the sinusoidal pe table. Verified against the previous raw-ggml implementation: cosine 1.000000 (CPU) / 0.999995 (CUDA).

## Long-text support

The 1024-frame cap no longer truncates: text is chunked at sentence/clause boundaries (Arabic break characters included), each chunk is synthesized within the graph budget, and chunks chain — chunk N+1 is conditioned on the last ~2s of chunk N's latent with a matched transcript tail. Verified 4×-length Arabic input → 10.2s of generated audio, Whisper round-trip recognizes every chunk.


Abubakabir34 (approved):

0xShug0:
@tareko Thank you for contributing new models! Just a quick comment: It would be great if you could reuse the framework helpers, modules and runtimes, following the patterns used by the newer models. The `dev` branch is a better reference that `main`, as it reflects the latest framework patterns and changes.

tareko:
Thanks! Implemented in 4f4a119: the DiT graph is now composed from the framework modules (Linear/LayerNorm/Embedding/DepthwiseConv1d/Gelu/Silu/RoPE/SDPA + structural/primitives) with module-typed weights through BackendWeightStore, following the parakeet_tdt/roformer dev-branch patterns. The only local helpers left are ops without a framework equivalent: grouped Conv1d (lowered like Conv1dModule's own im2col+mul_mat), GRN, adaLN modulate, and the sinusoidal pe table. Verified against the previous raw-ggml implementation at cosine 1.000000 (CPU) / 0.999995 (CUDA). The FlowSamplerRuntime adoption for the CFM loop is next (minimax_music3 pattern) — I kept the hand-rolled Euler loop this round to keep the parity diff minimal.

tareko:
Follow-up in 3b78bf9: long-form pacing fixed (chunks sized by the duration budget from the reference's frames/char; character-based pacing with bounds only against pathological refs), chunker made UTF-8-safe, and f5_dit_forward_cfg now also uses the module-composed graph — no raw ggml graph bodies remain in the runtime. Also switched CUDA allocation to the gallocr-only flow; the module graph's larger ctx made the old alloc_ctx_tensors + gallocr double-allocation OOM at N~1000.

tareko:
I'm still actively working on this, as there seems to be a bug with the letter "ح" vs. "ه" that I haven't been able to track down. I'll update when this is fixed.

0xShug0:
@tareko Just ping me when it is ready :).

tareko:
OK, this is ready to go now. It works very well and is probably at the state of the art for Arabic TTS.

0xShug0:
@tareko Could you convert the safetensors weights to GGUF (at least for the default package)? The default package need to be GGUF otherwise it will fail CI. We no longer actively maintain safetensors support. The UI now only handles GGUF models and safetensors packages will no longer be shown. You can host the GGUF files in your own HF repo, or add a conversion script and I can convert and host them.

Another issue with the PR is that it shows a lot of changes to the GGML kernels. Please keep GGML untouched. Almost all issues encountered during model development can be addressed by carefully rewriting the graphs or following the correct patterns/primitives used by the other released models.

tareko:
> @tareko Could you convert the safetensors weights to GGUF (at least for the default package)? The default package need to be GGUF otherwise it will fail CI. We no longer actively maintain safetensors support. The UI now only handles GGUF models and safetensors packages will no longer be shown. You can host the GGUF files in your own HF repo, or add a conversion script and I can convert and host them.

In process and I'll push once ready.

> Another issue with the PR is that it shows a lot of changes to the GGML kernels. Please keep GGML untouched. Almost all issues encountered during model development can be addressed by carefully rewriting the graphs or following the correct patterns/primitives used by the other released models.

It looks like a lot of changes, but fundamentally just a few lines (18 to be exact). However, the file somehow got a CRLF in it, which I noted was anyway against your style guide, so it was just converting it back to LF that triggered that bigtime diff. If you ignored the CRLF -> LF changes, it was a few lines that had to do with preventing an escape condition and adding a bit of debugging when needed. No problem to completely drop those edits.

tareko:
Thanks for the review! Both points addressed in the latest commits:

**GGUF packaging** — the default package (and all dialect packages) are now GGUF:
- `tools/convert_f5_tts.py` drives `audiocpp_gguf` to produce one self-contained GGUF per checkpoint with two namespaces: `transformer.*` (DiT, raw EMA torch names) and `vocos.*` (Vocos vocoder), plus a standalone `vocos-mel-24khz` package.
- The model spec has a real gguf sources entry (`transformer`/`vocos` namespaces) and all 9 packages are `gguf`/`orig` now; the safetensors source entry stays for development only.
- Runtime loads GGUF through namespace-prefixed tensor views, so safetensors and GGUF converge on identical tensor names. Verified against the GGUF on CUDA: DiT stage parity 0.9997, batched-CFG parity 0.99999, tokenizer id parity, and e2e synthesis passes our whisper-based pronunciation suite. Fresh-package CLI synthesis works from the package directory with no extra options (vocoder is bundled).
- The GGUF files are converted and ready; I'll upload them to `tareko/audio.cpp` on HF shortly (the spec's download repo points there). If you'd rather host them on the org side, the conversion script reproduces every package exactly (`python3 tools/convert_f5_tts.py --checkpoint-root <Habibi-TTS> --vocos <vocos.safetensors> --converter build/bin/audiocpp_gguf`).

**GGML untouched** — `external/ggml` is now byte-identical to upstream and drops out of the PR diff entirely. The two edits were defense-in-depth/debug leftovers from the arena-aliasing hunt; the actual fix always lived in the F5 graph code (staged constants get private backend buffers before `gallocr_reserve`, per-call inputs live in a persistent io buffer). I re-ran the full verification with upstream ggml: DiT parity 0.9997 (CUDA), CFG parity 0.99999, tokenizer parity, and 16/32-step long-form sampler output verified speech-clean via ASR — the original corruption does not come back.

tareko:
GGUF packages are now uploaded and installable: all 9 (default `habibi_unified` + 7 specialized dialects + `vocos_mel_24khz`) live at https://huggingface.co/trklou/audio.cpp — the spec's download repo points there. Verified end-to-end from a clean models dir: `model_manager_v2.py install habibi_unified` → `audiocpp_cli --family habibi` synthesis with the bundled vocoder, plus `sizes` checks pass for every package (CI's download path). One bug surfaced and fixed along the way: the bundled-vocoder probe used a dotted name while packed GGUF namespaces are slash-separated — GGUF-only installs without a fallback vocoder now work.

0xShug0:
@tareko Could you fix the windows failure? Then we are almost good to go.

0xShug0:
@tareko Some other issues during testing:

1. `cfg_strength` is exposed in `model_specs/f5_tts.json` but ignored by the runtime.

2. `F5_MEL_TEST` is compiled into the normal model target. This exposes test hooks in the production model library.

3. Production code still has hidden env toggles:
   - F5_FRAME_BUDGET
   - F5_NO_RMS_NORM
   - F5_SINGLE_CHUNK
   - F5_DUMP_STAGES
   These should not control normal production behavior. If needed, make them real request/session options or remove debug-only paths.

4. Docs are stale.

5. There are many unused-variable/function warnings when compile.

tareko:
> @tareko Some other issues during testing:
> 
>     1. `cfg_strength` is exposed in `model_specs/f5_tts.json` but ignored by the runtime.
> 
>     2. `F5_MEL_TEST` is compiled into the normal model target. This exposes test hooks in the production model library.
> 
>     3. Production code still has hidden env toggles:
>        
>        * F5_FRAME_BUDGET
>        * F5_NO_RMS_NORM
>        * F5_SINGLE_CHUNK
>        * F5_DUMP_STAGES
>          These should not control normal production behavior. If needed, make them real request/session options or remove debug-only paths.
> 
>     4. Docs are stale.
> 
>     5. There are many unused-variable/function warnings when compile.

Working on these now. Could I ask if these were all done by human review from your expertise, or if you used AI to uncover them? I'm trying to improve my skills and know where best to focus. Thank you for the mentorship and guidance through this!

tareko:
Commit 184d6fdfc122ecc3d9695b877fce220e0d50b8d4 dealt with a bug I spotted during use of the last phoneme occasionally being cut.

0xShug0:
Human + AI :). I’ll build and run your models myself, so I can catch some blockers immediately. I also know some common issues from mistakes I’ve made before, so I know where to have AI focus its checks. If you unleash AI on the whole codebase, it’ll nitpick everything and waste a lot of time (some can be just hallucinations).

BTW, the Windows CI failure is probably because you used `M_PI`. You can follow the approach used by other models to address this issue.

tareko:
Both addressed in `1a8bf22`:

**Windows failure**: `M_PI` is not defined by MSVC — replaced with a local `kPi` constant in synthesize.cpp. Should be green on the re-run.

**Round-2 issues**:
1. `cfg_strength` is now honored by the session (previously only `guidance_scale` was read; both work, spec updated).
2. `F5_MEL_TEST` no longer touches the production model target — hooks and parity binaries are gated behind a new `ENGINE_F5_TEST_HOOKS` cmake option (default OFF). Debug stage-dump taps are compiled out of production builds.
3. Env toggles removed from production behavior: `F5_FRAME_BUDGET` became the validated session option `f5_tts.frame_budget` (spec-documented); `F5_NO_RMS_NORM` and `F5_SINGLE_CHUNK` deleted (parity behavior is unconditional); `F5_DUMP_STAGES` is test-hooks-only.
4. Docs updated (milestones, session options, `cfg_strength` alias, GGUF status).
5. Deleted the dead pre-module raw-ggml graph helpers and unused variables — the `f5_tts` target now compiles warning-free (GCC; the full CUDA docker image builds clean too).

Re-verified after the cleanup: DiT parity 0.9997, CFG parity 0.99999, tokenizer parity, e2e pronunciation suite — all green.

tareko:
One small note is that while using this model, I noticed a bug with the use of diacritics (highly formalized arabic). I'm trying to debug it now.

tareko:
> Human + AI :). I’ll build and run your models myself, so I can catch some blockers immediately. I also know some common issues from mistakes I’ve made before, so I know where to have AI focus its checks. If you unleash AI on the whole codebase, it’ll nitpick everything and waste a lot of time (some can be just hallucinations).
> 
> BTW, the Windows CI failure is probably because you used `M_PI`. You can follow the approach used by other models to address this issue.

Thank you for the pointers.

tareko:
OK, the diacritic bug is now addressed. It seems to be a model deficiency, so now the audio.cpp version runs *better* than the "original". However, it's kind of embarrassing that we don't have a great training corpus for diacritics..

0xShug0:
Thank you @tareko! PR merged :tada:

PR #280, opened 2026-08-19
Fix CUDA arch selection and provide a portable arch list

xashr:
### Problem

All CUDA builds based on the CMakeLists.txt that do not explicitly set `CMAKE_CUDA_ARCHITECTURES` are only built for nvcc's default architecture: sm_75 on CUDA 13, sm_52 on CUDA 12.

The fallback to set `CMAKE_CUDA_ARCHITECTURES=native` is never reached *(and in addition is only supported on CMake >= 3.24)*.

**=> For local builds this means:** built for sm_75 (CUDA 13) / sm_52 (CUDA 12), even on an RTX 5090 (sm_120).
**=> For Docker builds this also means:** no portable set of architectures is built, only the minimal sm_52 or sm_75.

The Windows prebuilts are not affected (they pass a curated arch list); `build_windows.ps1 auto` explicitly targets the local GPU when one is present.

### Root cause

We set `cmake_minimum_required(VERSION 3.20)` which activates policy [CMP0104](https://cmake.org/cmake/help/latest/policy/CMP0104.html): during `enable_language(CUDA)`, CMake auto-initializes `CMAKE_CUDA_ARCHITECTURES` from nvcc's compile default whenever the user doesn't set it. The seed comes from the compiler-id probe - it does not query the local GPU. It is sm_75 on CUDA 13, sm_52 on CUDA 12.

Example (CUDA 12):
- On an RTX 5090 we see:
`-- Using CMAKE_CUDA_ARCHITECTURES=52 CMAKE_CUDA_ARCHITECTURES_NATIVE=120a-real`
- On a host without GPU (e.g. GitHub runner)
`-- Using CMAKE_CUDA_ARCHITECTURES=52 CMAKE_CUDA_ARCHITECTURES_NATIVE=No CUDA devices found.-real`

Since `CMAKE_CUDA_ARCHITECTURES` is initialized by CMake, the fallback to "native" is never used:
```
if (CMAKE_CUDA_ARCHITECTURES)
    set_target_properties(engine_runtime PROPERTIES CUDA_ARCHITECTURES "${CMAKE_CUDA_ARCHITECTURES}")
else()
    set_target_properties(engine_runtime PROPERTIES CUDA_ARCHITECTURES native)
endif()
```

### Solution

This PR brings the implementation closer to llama.cpp:

- We define a set of architectures to include in a default build. This is applied automatically if `CMAKE_CUDA_ARCHITECTURES` is not set.
  - Defaults are:
    - 50/61/70-virtual, *if < 13.0*
    - \+ 75/80-virtual
    - \+ 86-real;
    - \+ 89-real 90-virtual, *if ≥ 11.8*
    - \+ 120a-real, *if ≥ 12.8*
    - \+ 121a-real, *if ≥ 12.9*
- Users can reduce build times by explicitly setting `CMAKE_CUDA_ARCHITECTURES=native` (CMake >= 3.24) or by defining a custom set, e.g. `CMAKE_CUDA_ARCHITECTURES=120a-real`

For Docker builds we do not set `CMAKE_CUDA_ARCHITECTURES` and thus build for the full defined arch set.

### Details

- The default list is set **before** `enable_language(CUDA)` in CMakeLists.txt. That is the only point where "the user didn't set anything" is still detectable, because the CMake seed happens inside that call. The old if/else fallback quoted above sat after the seed and was therefore dead code - it is removed, since `CMAKE_CUDA_ARCHITECTURES` automatically initializes the `CUDA_ARCHITECTURES` property of every CUDA target.
- Suffixes: `-virtual` = PTX (compiled on the fly by the driver, forward-compatible), `-real` = SASS (native machine code).
- The default list is identical to llama.cpp's `ggml-cuda` default.
- Plain `12X` values (e.g. `120`, or `native` resolved on a Blackwell GPU) are upgraded to `12Xa` - the `a` variant unlocks Blackwell's FP4 tensor cores. Each replacement is logged.
- `native` requires CMake >= 3.24. On CMake 3.20-3.23 it falls back to the default list with a status message (same gate as llama.cpp's `GGML_NATIVE`, but visible instead of silently ignored).
- The configure log prints the final list (`-- Using CMAKE_CUDA_ARCHITECTURES=...`), so the result is easy to verify.
- The `CUDAARCHS` environment variable keeps working (CMake reads it before the seed).
- `.devops/cuda.Dockerfile` contains no arch list at all - CMakeLists.txt is the single source of truth. Custom images can still override via `--build-arg CUDA_DOCKER_ARCH="89-real;..."`.

### Consequences / Remarks

- Since we now define a default arch set, build times for default builds (without `CMAKE_CUDA_ARCHITECTURES`) increase by roughly 4.5x (CUDA 12 Docker build on a 24-core CPU).

### Open issues / TODO

- Update docs ✅
- Check workflows ✅
- Follow up? Avoid having two separate curated arch lists in CMakeLists.txt vs windows build scripts 📌

### DIFF
--- a/.devops/cuda.Dockerfile
+++ b/.devops/cuda.Dockerfile
@@ -17,6 +17,10 @@ ARG BASE_CUDA_RUN_CONTAINER=docker.io/nvidia/cuda:${CUDA_VERSION}-runtime-ubuntu
 FROM ${BASE_CUDA_DEV_CONTAINER} AS build
 
 ARG GCC_VERSION=14
+# CUDA architectures to compile for.
+# - default = the portable default list from CMakeLists.txt
+# - for a custom arch set build with --build-arg CUDA_DOCKER_ARCH="89-real;...".
+ARG CUDA_DOCKER_ARCH=default
 
 # Install build toolchain
 RUN apt-get update && \
@@ -30,8 +34,10 @@ ENV CC=gcc-${GCC_VERSION} CXX=g++-${GCC_VERSION} CUDAHOSTCXX=g++-${GCC_VERSION}
 WORKDIR /app
 COPY . .
 
-# Configure and build
-RUN cmake -S . -B build \
+RUN if [ "${CUDA_DOCKER_ARCH}" != "default" ]; then \
+        ADDITIONAL_CMAKE_ARGS="-DCMAKE_CUDA_ARCHITECTURES=${CUDA_DOCKER_ARCH}"; \
+    fi && \
+    cmake -S . -B build \
         -DCMAKE_BUILD_TYPE=Release \
         -DAUDIOCPP_MODEL_SET=full \
         -DENGINE_ENABLE_CPU_ALL_VARIANTS=ON \
@@ -43,6 +49,7 @@ RUN cmake -S . -B build \
         -DENGINE_BUILD_EXAMPLES=OFF \
         -DENGINE_BUILD_TESTS=OFF \
         -DENGINE_BUILD_WARMBENCH=OFF \
+        ${ADDITIONAL_CMAKE_ARGS} \
         -DCMAKE_EXE_LINKER_FLAGS=-Wl,--allow-shlib-undefined && \
     cmake --build build --parallel $(nproc) \
         --target audiocpp_cli \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,17 +166,74 @@ if (ENGINE_HIP_STRIX_HALO_OPTIMIZATIONS)
 endif()
 
 if (ENGINE_ENABLE_CUDA AND NOT ENGINE_ENABLE_HIP)
-    enable_language(CUDA)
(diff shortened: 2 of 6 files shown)


xashr:
@0xShug0 PR is tested now, docs updated. Verified with local builds and Docker images (fatbin arch sets checked with cuobjdump).
      
As mentioned above, this PR also changes the default build behavior. The default now builds the portable arch set defined in CMakeLists.txt - the same set llama.cpp uses. To reduce build times locally, `native` or a custom arch set can be specified; docs have been updated accordingly. Docker images now include the full portable arch set as well.

Note: PR #288 independently documented this same bug during a Jetson Orin bring-up. Its "always pass CMAKE_CUDA_ARCHITECTURES explicitly" advice is superseded by this fix.

📌 Possible follow-up: the Windows build script could use this default set instead of its own curated list → single source of truth.

0xShug0:
@xashr Merged. Thanks!

PR #293, opened 2026-08-22
cuda: make graph-cache eviction reachable; qwen3_asr: survive concurrent-load memory pressure

derekja:
Follow-up to #276. That fix made **serial** qwen3_asr load memory-stable (verified: three identical single-stream sweeps, VRAM byte-identical). Under **concurrent** `/v1/audio/transcriptions` load a distinct failure remained: VRAM grows on every run — including identical repeats of the same shapes — until allocation fails, after which **every request returns 500 until the process is restarted**.

Measured on a 12 GB H100L vGPU slice (no VMM), Qwen3-ASR-0.6B Q8, 8 concurrent streams of mixed 1–8 s utterances posting accumulated audio ~1/s: 6.2 → 8.9 → 10.5 GB (ceiling) across repeated sweeps, then permanent 500s. `nemotron_asr` under the identical workload is byte-stable, which localized the problem to the qwen3_asr graph lifecycle. With `GGML_CUDA_DISABLE_GRAPHS=1` the same workload is also byte-stable, which localized the reservoir.

Four changes, in causal order:

1. **Export `ggml_backend_cuda_clear_graph` through `get_proc_address`.** Every graph destructor already calls `engine::core::release_backend_graph_resources`, which resolves this name — but the CUDA registry never exported it, so eviction has been a silent no-op since it was added (`docs/build/HIP.md` claims it works; now it does). Single-threaded servers got away with it because the freed graph arena's address is typically reused by the next same-size `ggml_init`, overwriting the stale `cuda_graphs` entry in place; concurrent requests (one detached thread per request) perturb the address space, so each rebuild minted a fresh key and orphaned a `cudaGraph_t` + `cudaGraphExec_t`.

2. **Release the old graph before constructing its replacement** (prefill/decode/classification/encoder). Assigning `make_unique` over a live `unique_ptr` holds both arenas at the rebuild peak — and interleaved streams force a rebuild on nearly every request because the prefill/encoder caches are exact-shape single slots.

3. **Hold `ggml_gallocr_t` in a `unique_ptr`** (the existing `voxtral_realtime` pattern) so the `CapacityError`/`runtime_error` throws in the graph constructors stop leaking the partially reserved arena. Previously every failed rebuild after an OOM deepened the OOM.

4. **Trim idle pool memory and retry once when graph allocation fails.** The legacy pool (no-VMM path) caches every buffer it ever allocated and only flushes when *its own* `cudaMalloc` fails; graph arenas allocate through `ggml_backend_cuda_buffer_type_alloc_buffer`, which flushes nothing — so once the pool ratchets to the ceiling, every graph build fails forever while gigabytes of idle cached buffers sit reclaimable. Adds `ggml_cuda_pool::clear()`, an exported `ggml_backend_cuda_trim_pools()`, a framework resolver, and trim-and-retry on the four qwen3_asr allocation-failure paths. The trim costs one device sync and only runs on a previously-fatal failure.

**Validation** (same slice/model/harness): the workload that previously produced the terminal all-500 state now self-heals — at the ceiling the trim reclaimed 2.3 GB mid-load, 30/32 utterances in the pressure-point sweep still transcribed, and the server returned to full health (WER identical to baseline, finals p50 9 ms) with no restart. Serial behavior is unchanged. Transcripts across the load runs are unchanged (corpus WER 0.05 at every level, same as pre-patch).

Repro used throughout: N concurrent realtime websocket streams → accumulated-utterance POSTs of mixed lengths; happy to share the harness scripts if useful.

### DIFF
--- a/external/ggml/include/ggml-cuda.h
+++ b/external/ggml/include/ggml-cuda.h
@@ -23,6 +23,7 @@ extern "C" {
 GGML_BACKEND_API ggml_backend_t ggml_backend_cuda_init(int device);
 
 GGML_BACKEND_API bool ggml_backend_is_cuda(ggml_backend_t backend);
+GGML_BACKEND_API void ggml_backend_cuda_trim_pools(ggml_backend_t backend);
 GGML_BACKEND_API void ggml_backend_cuda_clear_graph(ggml_backend_t backend, const struct ggml_cgraph * graph);
 
 // device buffer
--- a/external/ggml/src/ggml-cuda/common.cuh
+++ b/external/ggml/src/ggml-cuda/common.cuh
@@ -1131,6 +1131,9 @@ struct ggml_cuda_pool {
 
     virtual void * alloc(size_t size, size_t * actual_size) = 0;
     virtual void free(void * ptr, size_t size) = 0;
+    // Release cached device memory back to the driver. Buffers handed out by
+    // alloc() are unaffected; only idle cached capacity is dropped.
+    virtual void clear() {}
 };
 
 template<typename T>
--- a/external/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/external/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -380,6 +380,10 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
         GGML_ASSERT(pool_size == 0);
     }
 
+    void clear() override {
+        clear_pool();
+    }
+
     void clear_pool() {
         ggml_cuda_set_device(device);
         for (int i = 0; i < MAX_BUFFERS; ++i) {
@@ -5034,6 +5038,21 @@ bool ggml_backend_is_cuda(ggml_backend_t backend) {
     return backend != NULL && ggml_guid_matches(backend->guid, ggml_backend_cuda_guid());
 }
 
+void ggml_backend_cuda_trim_pools(ggml_backend_t backend) {
+    if (!ggml_backend_is_cuda(backend)) {
+        return;
+    }
+    ggml_backend_cuda_context * cuda_ctx = (ggml_backend_cuda_context *) backend->context;
+    CUDA_CHECK(cudaDeviceSynchronize());
+    for (int device = 0; device < GGML_CUDA_MAX_DEVICES; ++device) {
+        for (int stream = 0; stream < GGML_CUDA_MAX_STREAMS; ++stream) {
+            if (cuda_ctx->pools[device][stream] != nullptr) {
+                cuda_ctx->pools[device][stream]->clear();
+            }
+        }
+    }
+}
+
 void ggml_backend_cuda_clear_graph(ggml_backend_t backend, const ggml_cgraph * graph) {
 #ifdef USE_CUDA_GRAPH
     if (!ggml_backend_is_cuda(backend) || graph == nullptr || graph->n_nodes <= 0) {
@@ -5856,6 +5875,9 @@ static void * ggml_backend_cuda_reg_get_proc_address(ggml_backend_reg_t reg, con
     }
     if (strcmp(name, "ggml_backend_cuda_clear_graph") == 0) {
         return (void *)ggml_backend_cuda_clear_graph;
+    }
+    if (strcmp(name, "ggml_backend_cuda_trim_pools") == 0) {
+        return (void *)ggml_backend_cuda_trim_pools;
     }
     return nullptr;
 }
--- a/external/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/external/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -5854,6 +5854,9 @@ static void * ggml_backend_cuda_reg_get_proc_address(ggml_backend_reg_t reg, con
     if (strcmp(name, "ggml_backend_get_features") == 0) {
         return (void *)ggml_backend_cuda_get_features;
     }
+    if (strcmp(name, "ggml_backend_cuda_clear_graph") == 0) {
+        return (void *)ggml_backend_cuda_clear_graph;
+    }
     return nullptr;
 }
 
--- a/include/engine/framework/core/backend.h
+++ b/include/engine/framework/core/backend.h
@@ -44,6 +44,9 @@ bool is_host_backend(ggml_backend_t backend);
 bool uses_host_graph_plan(BackendType type);
 bool uses_host_graph_plan(ggml_backend_t backend);
 bool requested_backend_uses_host_graph_plan(const BackendConfig & config);
+// Drop the CUDA/HIP context's cached (idle) pool memory back to the driver.
+// No-op on other backends. For use on allocation-failure paths before a retry.
+void trim_backend_pools(ggml_backend_t backend);
 void release_backend_graph_resources(ggml_backend_t backend, ggml_cgraph * graph);
 void release_backend_graph_resources(BackendType backend_type, ggml_backend_t backend, ggml_cgraph * graph);
 void validate_backend_graph_supported(ggml_backend_t backend, ggml_cgraph * graph, const char * label);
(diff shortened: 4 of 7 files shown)


derekja:
RTF data, `audiocpp_cli --metrics` as in the #276 review. Same slice as the PR description (12 GB H100L vGPU, CUDA backend). Method: one `--request-sequence` session per family per binary — 6 requests per audio length, first request per length excluded as graph build, p50 (min–max) of the 5 warm runs. Baseline is `main @ 54ea3d1`, patched is this branch.

qwen3_asr 0.6B Q8 (the family this PR touches):

| audio | main | patched |
|---|---|---|
| 2 s | 0.0326 (0.0326–0.0394) | **0.0258** (0.0258–0.0285) |
| 6 s | 0.0402 (0.0400–0.0429) | **0.0354** (0.0354–0.0368) |
| 15 s | 0.0243 (0.0242–0.0252) | 0.0242 (0.0241–0.0250) |
| 30 s | 0.0259 (0.0259–0.0264) | 0.0258 (0.0257–0.0262) |

nemotron_asr 0.6B Q8 (regression check for the shared ggml/framework changes):

| audio | main | patched |
|---|---|---|
| 2 s | 0.0098 (0.0096–0.0157) | 0.0096 (0.0095–0.0155) |
| 6 s | 0.0097 (0.0094–0.0129) | 0.0092 (0.0087–0.0121) |
| 15 s | 0.0067 (0.0066–0.0080) | 0.0065 (0.0060–0.0071) |
| 30 s | 0.0058 (0.0057–0.0072) | 0.0065 (0.0064–0.0070) |

Long-file RTF is unchanged on both families and nemotron sits inside its baseline spread everywhere. Short/medium qwen is measurably *faster* on the patched build (2 s: 0.0326 → 0.0258; ranges don't overlap) — plausibly the reset-before-rebuild change relieving allocator pressure, though I haven't isolated it; no length is slower. Consistent with the patch adding no code to the warm path (eviction runs at graph destruction, the pool trim only on a previously-fatal allocation failure).

0xShug0:
@derekja Could you clean up the churn in the PR caused by the CRLF changes? Changes to GGML are usually high-risk because of their large blast radius and generally require a regression sweep, unless they’re purely additive and opt-in. The current PR shows full-file churn, which makes it difficult to tell what actually changed.

derekja:
you bet, thanks. running regression tests overnight and will post when complete.

derekja:
Thanks for the careful look, and apologies for the churn — that was my tooling silently normalizing line endings on the three `external/ggml` files (they're stored CRLF, so a text-mode edit rewrote every line). I've rebuilt both commits from `main`'s exact bytes with only the real insertions and force-pushed; the PR now shows the actual change: **+110/−22 across 7 files**.

On blast radius, agreed — here's the breakdown and the regression sweep.

**What the ggml side actually changes (+29 lines, no deletions):** two proc-address table entries, `ggml_backend_cuda_trim_pools()` (new, reachable only via the new proc entry), a `clear()` override on the legacy pool that calls its existing `clear_pool()`, and a `virtual void clear() {}` with a default no-op body on the pool base. Nothing existing is modified. The one behavioral change to pre-existing code is the point of the fix: `release_backend_graph_resources` — already called by every family's graph destructors — stops being a silent no-op on CUDA, so the sweep below deliberately covers multiple families, not just qwen3_asr.

**Regression sweep** (12 GB H100L vGPU, CUDA backend, this branch):

- `ctest` (`-DENGINE_BUILD_TESTS=ON`): **56/56 passed**.
- `run_audiocpp_cli_path_tests.py` + `compare_audiocpp_cli_path_results.py`, baseline `main @ 54ea3d1` CLI vs patched CLI, every family I have local models for: qwen3_asr (3 cases), nemotron_asr (2), citrinet_asr (1), hviske_asr (1), voxcpm2 (2 — a TTS family, included since the eviction change fires in all families' destructors): **9/9 compared cases, 0 differences** — artifacts and text byte-match between the two builds. Caveats for reproducibility: the `resources/*.wav` fixtures aren't in the repo, so I substituted local audio (identical for both runs, which is what the A/B needs); the qwen forced-aligner case exercises the aligner model I don't have installed and the voxcpm2 voice-clone case failed identically on both builds against my substitute reference clip — both failures are byte-identical base-vs-patched and counted in the 0-difference comparison.
- RTF: the `--metrics` A/B in the comment above (equal-or-faster, no length regressed).
- The concurrent-load soak from the PR description, re-run on the clean branch: previously-fatal workload self-heals; serial byte-stability unchanged.

One pre-existing note from running the unit suite: `codec_dequant_parity` (tests/moss_tts_local) failed to compile on the first full-tests build against current `main` headers — unrelated to this PR (neither file is in this diff); mentioning it in case it's already known.

Happy to run any additional sweep you'd like — the concurrent-load harness that produced the leak repro is scriptable and I can share it.

0xShug0:
@derekja What worries me most is the `ggml_backend_cuda_clear_graph` change 

```
if (strcmp(name, "ggml_backend_cuda_clear_graph") == 0) {
    return (void *)ggml_backend_cuda_clear_graph;
}
```

I agree this is a correctness fix: before this PR, `release_backend_graph_resources`(...) was effectively a no-op for CUDA. The concern is that this correctness fix can expose lifecycle/performance issues in models that destroy and rebuild graphs between requests. A common use case (mostly for DiT models) is sending the same request multiple times with only a different seed to find the best results. The graph shape does not change, so before this PR CUDA graph cache state could be reused **accidentally**  -- a happy coincidence. E.g. I tested MiniMax-H3 text-to-audio with default settings in one server session, using repeated requests. The PR introduced about a 9% warmed-request regression in that case.

main warmed avg  (across 5 requests): 4460.38 ms
PR warmed avg:                 4858.80 ms
delta:                         +398.42 ms
regression:                    +8.93%

Depending on the GPU, model, and request shape, the CUDA graph rebuild cost may be small enough to look like noise, but it can also be noticeable.

So what could delay the merge now is that we need to audit which models are affected by this lifecycle change, then refactor and validate the affected models before merging the PR.

I think a safer approach is to make the new CUDA graph cleanup opt-in for now, with only Qwen3-ASR enabling it initially. Any thoughts?

derekja:
Sure, that makes sense. I'm happy to take on some of the testing to make it more generalizable if you like, although I'd want some guidance on what could be most convincing in that regard. Alternatively, I could see the opt-in as an optional flag on the existing call, or by putting the new behaviour in a new call, evict_backend_graph_cache(backend, graph) or something like that.

Do you have a preference? I'll push something into this PR to optionalize it and re-run the tests, if you like.

0xShug0:
@derekja Thanks! I’d prefer a simple opt-in flag on the existing call, something like:

```
void release_backend_graph_resources(
    ggml_backend_t backend,
    ggml_cgraph * graph,
    bool evict_cuda_graph_cache = false);

void release_backend_graph_resources(
    BackendType backend_type,
    ggml_backend_t backend,
    ggml_cgraph * graph,
    bool evict_cuda_graph_cache = false);
```

Then just: 
```
void release_backend_graph_resources(
    ggml_backend_t backend,
    ggml_cgraph * graph,
    bool evict_cuda_graph_cache) {
    if (!evict_cuda_graph_cache) return; // existing callsite/behavior unchanged
     ....
}
```
Default should stay `false`, so existing models keep the current CUDA graph-cache behavior. Then Qwen3-ASR can opt into the real cleanup path in this PR

`release_backend_graph_resources(backend, graph, true);`

and we can audit/refactor other models separately later.

derekja:
Done — implemented exactly as you sketched: `evict_cuda_graph_cache = false` on both overloads with the early return, so every existing call site keeps current behavior, and qwen3_asr's four graph destructors pass `true`. Pushed as one additional commit. Re-validated on this branch: ctest 56/56; the 5-family path-test comparison against `main @54ea3d1` still reports 0 differences; and the qwen concurrent-load repro still self-heals with the opt-in in place. Happy to help with the wider audit whenever you get to it — the concurrent-load harness is yours if useful.

0xShug0:
Thank you @derekja! PR merged :tada:

PR #297, opened 2026-08-22
Make embedded WebUI work behind a path-prefix reverse proxy

### LINKED ISSUE #294: WebUI unusable behind a path-prefix reverse proxy (renders its own 404; API calls escape the prefix)
## Use case

llama-swap (https://github.com/mostlygeek/llama-swap) can manage `audiocpp_server`
as an on-demand TTS backend: it starts/stops one instance per configured model and
proxies it under a path prefix, e.g. `https://host/upstream/<model>/`. The OpenAI
endpoints work great this way - but the embedded WebUI does not.

## Problem

The WebUI is a SvelteKit app built for the domain root:

1. **Pathname routing** - the document is served fine at
   `/upstream/<model>/` (HTTP 200), but the client-side router resolves the
   current `location.pathname`, finds no matching route, and renders the
   app's own 404 page. So any path-prefix deployment shows "404" even though
   the server is healthy.
2. **Root-absolute API calls** - `src/lib/api.ts` fetches `/health`,
   `/v1/models`, `/v1/audio/speech`, … with absolute paths, so behind a prefix
   they escape to the proxy root instead of reaching the instance.

## What fixed it for me (tested)

- `svelte.config.js`: `kit.router = { type: 'hash' }` - the pathname stops
  mattering, and `paths.relative: true` (already set) keeps assets working.
  Requires dropping the `prerender`/`ssr` page options from
  `src/routes/+page.ts` (hash router rejects them at build time).
- Rebasing the root-absolute fetches onto the page's directory. I used a
  small `window.fetch` wrapper in `app.html` (no-op when served at a real
  root); making the `api.ts` paths relative would be the cleaner in-tree fix.

With those two changes the full UI (model load, TTS, cloning) works behind
`/upstream/<model>/` with zero proxy configuration. Happy to send a PR if
you're open to either approach.

SelfRef:
Fixes #294 

The embedded WebUI was unusable when `audiocpp_server` runs behind a path-prefix
reverse proxy (e.g. llama-swap serving an instance at `/upstream/<model>/`):
the SvelteKit router resolved `location.pathname`, found no matching route, and
rendered the app's own 404 page - and the root-absolute API calls in
`src/lib/api.ts` escaped the prefix to the proxy root.

## Changes

- **`svelte.config.js`** - switch the client router to hash mode
  (`router: { type: 'hash' }`), so routing no longer depends on the pathname
  the page is served under. Together with the existing `paths.relative: true`
  this makes the document itself fully location-independent.
- **`src/routes/+page.ts`** - removed; the hash router rejects the
  `prerender`/`ssr` page options at build time, and the static adapter's
  `index.html` fallback already produces the same single-file SPA output.
- **`src/lib/api.ts`** - add a small `apiUrl()` helper that resolves the
  endpoint paths against `document.baseURI`, applied at the two `fetch` call
  sites, so all API traffic stays inside the prefix. Serving at the domain
  root is unaffected (the resolved URL is identical).
- **`dist/index.html`** - rebuilt, since CMake embeds this artifact into
  `audiocpp_server`.

## Testing

- `npm run build` and `npm run check` pass (0 errors, 0 warnings).
- Headless-browser smoke test with the built page served at
  `/upstream/testmodel/` behind a stub proxy: the app renders normally (no
  404 page) and the request log shows `health` and `v1/models` fetched as
  `/upstream/testmodel/health` and `/upstream/testmodel/v1/models` - nothing
  escapes to the proxy root.
- Full workflow (model load, TTS, cloning) previously verified manually behind
  llama-swap's `/upstream/<model>/` prefix, per the issue.


0xShug0:
@SelfRef Merged. Thank you for the PR!

Issue #173, opened 2026-08-01
[Feature Request] TTS models that clone - option to persist embeddings to disk and upload voices


langfod:
Add `POST/DELETE /v1/audio/voices` to upload named reference files and store as embeddings.

option to persist embeddings to disk and not just in-memory cache.  
option to populate embeddings to in-memory cache on server start or first use ( preload_lazy ).

Server would at least get what is available in order to populate `GET /v1/audio/voices?model=<id>`

Send error is requested speaker/name is not found in cache.

Latter parts are probably similar to PocketTTS.

0xShug0:
@langfod Would you like to explain your use cases? Building embedings are usually very fast. My biggest concern is that the inference server should not handle asset management, which should be the responsibility of the frontend/UI.

langfod:
@0xShug0 

mainly from an API perspective.

Current use case may involve many voice types and many different calls. So it saves a good bit of  bandwidth and time to not need to resend the reference wav constantly.

This is fairly typical with other `/v1/audio/voices` style APIs.

So the flow is generally:  
  call GET `/v1/audio/voices` 
  when about to make a request and your voice wasn't in the list then upload your voice.
  call `/v1/audio/speech` with the voiceid and speech request.


While embeddings are usually very fast they still can add to latency.

0xShug0:
@langfod Thank you for the explanation. You mentioned “upload your voice.” In that flow, who authenticates and authorizes that operation?

I’m open to in-memory named reference voice cache for a running session, but I’d like to keep persistence, ownership, and access control outside the inference server.

0xShug0:
Just some more context on why I don’t want to blur the responsibility: A naive local-folder implementation is easy to add, but hard to evolve once exposed as an API contract. The moment clients depend on POST /voices persisting data, the server has to define storage lifetime, naming conflicts, overwrite/delete behavior, cleanup after failed uploads, backup/migration behavior, and more. Later, replacing local disk with S3, a database, or a frontend-managed store becomes a breaking change or a compatibility layer we have to maintain indefinitely.

langfod:
so a multi-user scenario? That does get tricker. Is that in-scope of the current server or just for safety in-case someone uses it that way?

That would probably involve adding something like a workspace field to audio.cpp server and a startup option to allow writes.
 
Admittedly, I had been looking at this from a single-user setup.

Currently working on a wrapper that will intercept the "voices" and "speech" calls to at least cache the "voice_ref".

[...]

PR #300, opened 2026-08-23
very basic API-keys

erik-elmgren:
Let me know if this is interesting.

I think it would disable the web gui at the moment as there is no easy way for a browser to add a authorization bearer header to all the requests.

Obviously https is a next nice thing to have to make it fully useful.

### DIFF
--- a/app/server/config.cpp
+++ b/app/server/config.cpp
@@ -260,6 +260,14 @@ ServerConfig load_server_config(const std::filesystem::path & path) {
         throw std::runtime_error("server threads must be positive");
     }
 
+    const auto * apikeys = root.find("apikeys");
+    if (apikeys && !apikeys->is_array()) {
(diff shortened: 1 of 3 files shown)


0xShug0:
@erik-elmgren Thanks for PR. There are some issues with the PR, as it breaks existing pipelines that use the audio.cpp server and the implementation is incomplete. However, my biggest concern is that it goes against the current design principles of the server. In general, the server is intended to stay lightweight. Features that require additional application-level state or infrastructure are generally better handled by a layer in front of audio.cpp (e.g., a reverse proxy) rather than being built into the inference server itself. You can check the discussion here for more context: https://github.com/0xShug0/audio.cpp/issues/173

That said, I’m happy to reopen the PR if authentication is implemented as a complete, self-contained, and fully opt-in subsystem that does not affect existing server behavior or pipelines when disabled.

And please don’t let this discourage you from contributing!

Issue #249, opened 2026-08-15
Forced aligner: --words-out + fixed chunking throws "timestamp outside the chunk span" at ~the 10th chunk of a process; the same audio passes in a fresh process


bjhengen:
## Summary

`--words-out` (Qwen3 forced aligner) with fixed-mode chunking throws

```
audiocpp_cli failed: Audio chunker word merge received a timestamp outside the chunk span
```

at roughly the **9th–10th chunk of a single process**, aborting the whole run so no timestamps are
produced for any chunk.

The important part: **the same audio passes in a fresh process.** The failure needs *both*
triggering content *and* enough accumulated per-process state — neither alone is sufficient. That
points at per-call state in the aligner rather than at any particular audio or timestamp value.

This is **long-standing, not a regression** — `release-0.4` fails identically (see below). Please
don't search a recent commit window for it.

## Environment

| | |
|---|---|
| Version | `release-0.5` (`3178daf`), built from source |
| Also fails on | `release-0.4` (archived binary, clean 0.4 worktree + 0.4 `model_specs/`), 3/3 |
| Still present on | `main` @ `f08e61d` — `src/framework/audio/chunking.cpp:602` is unchanged |
| Backend | CUDA, RTX 5090 (sm_120), CUDA 13.3 |
| Models | `Qwen3-ASR-1.7B-hf` + `Qwen3-ForcedAligner-0.6B` |
| Input | 16 kHz mono WAV (ffmpeg-normalised) |

## Reproduce

```bash
# FAILS at ~chunk 10
./build/bin/audiocpp_cli --task asr --family qwen3_asr \
  --model models/Qwen3-ASR-1.7B-hf --backend cuda \
  --audio repro-A-150s-fails.wav \
  --text-out /tmp/t.txt --words-out /tmp/w.json \
  --session-option qwen3_asr.forced_aligner_model_path=models/Qwen3-ForcedAligner-0.6B \
  --audio-chunk-mode fixed --audio-chunk-seconds 15

# PASSES -- same length, same chunk count, different content
#   ... --audio repro-B-150s-PASSES-negative-control.wav ...

# PASSES -- same content as A, fewer chunks
#   ... --audio control-60s-passes.wav ...
```

## What isolates it to per-process state, not the audio

The **same 15 s of audio** (135–150 s of `repro-A`):

| Presented as | Result |
|---|---|
| chunk 1 of 1 (isolated) | **pass** |
[...]

Issue #167, opened 2026-07-31
VibeVoice-ASR: chunked (VAD/fixed) timestamps are in the input sample rate, not the 24 kHz single-pass domain — inconsistent


sataz-ehl:
## Summary

For `vibevoice_asr`, the sample-rate domain of the returned `start_sample`/`end_sample` **differs between single-pass and chunked modes**, so the same audio yields timestamps that mean different things depending on `audio_chunk_mode`. Because the ASR response has no `sample_rate` field, a client cannot detect which domain it got.

- **`audio_chunk_mode=none` (single-pass):** `frontend_.normalize()` resamples to `config.sample_rate` (24000) before the span math → timestamps in the **24000** domain. ✅
- **`audio_chunk_mode=vad`/`fixed` (chunked):** the chunk planner runs on the **raw input** (`plan_vad_audio_chunks(*request.audio_input, …)`), so each chunk's `source_span.start_sample` is in the **input** rate, and `append_offset_speech_metadata` adds it to the per-chunk (24000-domain) position **without rate conversion**. The returned absolute timestamps come back in the **input** rate.

## Reproduce (9B q8_0, engine release-0.4.2 `27d87ba`)

Feed a **16 kHz** 30 s clip:
- `mode=none` → `max(end_sample) = 720000` (= 30 s × 24000). ✅
- `mode=vad, audio_chunk_seconds=15` → `max(end_sample) = 480000` (= 30 s × **16000**). Dividing by 24000 gives a wrong 20 s; dividing by 16000 gives the correct 30 s.

Feeding **24 kHz** to make the domains match is not possible for chunked mode: the VAD rejects it —
```
{"error":{"message":"Silero VAD 16k model only supports sample_rate=16000","type":"server_error"}}
```
So `mode=vad` effectively **requires 16 kHz input**, which then makes its timestamps land in the 16 kHz domain while single-pass uses 24 kHz. There is no feed rate that makes both modes agree.

## Impact

Clients that convert `start_sample → seconds` need a single, reliable divisor. Today they must special-case `mode` (÷24000 for single-pass, ÷ input-rate for chunked) and *know the fed rate*, which is fragile. Segment/SRT timing is silently wrong (×1.5) if the single-pass divisor is assumed for chunked output.

## Suggested fix

Either of:
[...]

PR #301, opened 2026-08-24
Drop out-of-span chunk speech metadata instead of aborting the run

bjhengen:
Found while investigating #249. `src/framework/audio/chunking.cpp` has two sibling merge paths that compute the identical out-of-span guard and then disagree about what it means:

| path | function | on out-of-span | reached by |
|---|---|---|---|
| word timestamps | `append_chunk_word_timestamps` | warn + `continue` (since #252) | `qwen3_asr` forced aligner |
| speech metadata | `append_chunk_speech_metadata` → `merge_span` | **`throw`**, aborting the whole run | `vibevoice_asr` |

#252 chose warn-and-drop for the word path. This PR applies that same decision to the metadata path, which still aborts. It is not proposing a new policy — it is finishing the one already merged.

`merge_span` covers **both** `speech_segments` and `speaker_turns`, so either can abort a run.

## The change

`merge_span` already returns `std::optional<TimeSpan>`, and both call sites already `continue` on `std::nullopt`. So the throw becomes a warning plus `return std::nullopt` — the "not applicable" signal the function was already built around. No signature change, no call-site change. The warning carries the same diagnostic fields the word path emits (`local_start`, `local_end`, `source_samples`, the source/keep spans, both sample rates).

## Test

Adds `test_chunk_speech_metadata_merge_drops_outside_spans`, mirroring #252's `test_chunk_word_timestamp_merge_drops_outside_words`: one kept span and one outside span for **both** `speech_segments` and `speaker_turns`, asserting the outside ones are dropped and the valid ones survive with correct global spans.

Only `test_chunk_speech_metadata_merge_rescales_chunk_domain` existed before, so the out-of-span behaviour on this path had no coverage at all.

## Scope: which routes were checked

Shared framework code, so to be explicit about blast radius:

- **Affected:** the `vibevoice_asr` chunked path (`speech_segments` and `speaker_turns`). Behaviour change: a run that previously aborted now completes, with the out-of-span entry dropped and a warning logged.
- **Not affected:** the `qwen3_asr` word-timestamp path — untouched; it has behaved this way since #252. `append_chunk_speech_metadata` calls `append_chunk_word_timestamps` at the end, so today a vibevoice user gets the lenient behaviour for words and the fatal one for segments within the same call.
- No output, performance or memory change on any path that does not hit the guard.

## Verification

Verified on this PR's base, `main` @ `288a271`, CPU-only:

```bash
cmake -S . -B build-cpu -DENGINE_ENABLE_CUDA=OFF -DENGINE_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release
cmake --build build-cpu --target audio_chunking_test -j8
./build-cpu/bin/audio_chunking_test
```

| state | result |
|---|---|
| new test, **stock** `chunking.cpp` | ❌ `audio_chunking_test failed: Audio chunker speech metadata merge received a speech segment outside the chunk span` (exit 1) |
| new test, **with the fix** | ✅ `audio_chunking_test passed` (exit 0) |

So it is a real regression test: it fails without the change and passes with it, on the exact commit this PR targets.

Earlier I ran the same pair at `release-0.6.1` in both configurations — CPU-only and CUDA (`-DENGINE_ENABLE_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=120`, sm_120 / RTX 5090) — with identical results. I did not repeat the CUDA build on `main`; `src/framework/audio/chunking.cpp` is byte-identical between `release-0.6.1` and `288a271` (unchanged since #252), and this path has no backend-dependent code.

## Known limitation, stated plainly

**Reachability of the metadata case is unproven.** I have not observed an out-of-span *segment or turn* in the wild — my own trigger (#249) is on the *word* path. The argument for this change is the inconsistency itself: the same guard on the same arithmetic in the same file should not abort in one place and drop in the other.

The supporting evidence that it is not theoretical:

- The aligner demonstrably emits spans outside the chunk — that is exactly #249 (`local_start=272640` vs `source_samples=272000`, 40 ms past the end of the chunk's audio).
- The `vibevoice_asr` chunked-timestamp handling already has a third-party report against it: #167.

If you would rather see a reproduction on the metadata path before changing behaviour there, that is a fair ask and I am happy to keep digging instead.

### DIFF
--- a/src/framework/audio/chunking.cpp
+++ b/src/framework/audio/chunking.cpp
@@ -648,8 +648,19 @@ void append_chunk_speech_metadata(
         const int64_t local_start = std::max<int64_t>(local_span.start_sample, 0);
         const int64_t local_end = std::min<int64_t>(local_span.end_sample, source_samples);
         if (local_start >= local_end) {
-            throw std::runtime_error(std::string("Audio chunker speech metadata merge received a ") +
-                label + " outside the chunk span");
+            std::ostringstream warning;
+            warning << "dropping " << label << " outside chunk span"
+                    << " local_start=" << local_span.start_sample
+                    << " local_end=" << local_span.end_sample
+                    << " source_samples=" << source_samples
+                    << " source_start=" << source_span.start_sample
+                    << " source_end=" << source_span.end_sample
+                    << " keep_start=" << keep_span.start_sample
+                    << " keep_end=" << keep_span.end_sample
+                    << " source_sample_rate=" << source_sample_rate
+                    << " timestamp_sample_rate=" << timestamp_sample_rate;
+            debug::log_message(debug::LogLevel::Warning, "audio.chunking", warning.str());
+            return std::optional<runtime::TimeSpan>{};
         }
         runtime::TimeSpan global{
             timestamp_source_span.start_sample + local_start,
--- a/tests/unittests/test_audio_chunking.cpp
+++ b/tests/unittests/test_audio_chunking.cpp
@@ -879,6 +879,41 @@ void test_chunk_word_timestamp_merge_drops_outside_words() {
     require_span(merged[1].span, 1050, 1080, "second valid word span");
 }
 
+void test_chunk_speech_metadata_merge_drops_outside_spans() {
+    engine::runtime::TaskResult chunk;
+    auto kept_segment = speech(20, 40);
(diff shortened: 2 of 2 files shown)


0xShug0:
Thanks @bjhengen! PR merged.

PR #306, opened 2026-08-25
feat(server): add --idle-unload-ms and pre-load memory guard

gqf2008:
## Motivation
On a 16GB Apple Silicon machine, a long-running batch transcription loaded model after model with no eviction or idle unload, exhausting unified memory and surfacing as `kIOGPUCommandBufferCallbackErrorOutOfMemory` + HTTP 500s. This PR gives `audiocpp_server` two server-side memory controls that complement the existing `max_loaded_models` (#298).

## Changes
1. **`--idle-unload-ms <ms>`** (`idle_unload_ms` in server.json, default `0` = disabled): a background thread unloads every resident non-busy model once the server has gone that long without a model load/run. The next request reloads lazily. Replaces the previous external Python monitor (which polled the log mtime and required `--log` trace spam); no wrapper process needed.
2. **`--min-free-memory-mb <mb>`** (`min_free_memory_mb` in server.json, default `512`, `0` disables the extra headroom): before every lazy load, estimate the model footprint (weights, including session auxiliary files and directory trees, × 1.5 + 128 MiB floor) and refuse with HTTP 503 `insufficient_memory` when estimate + headroom does not fit the free host memory (Windows `GlobalMemoryStatusEx` / Linux `MemAvailable` / new macOS Mach VM stats) or the backend device memory (`ggml_backend_dev_memory`, CUDA/HIP/Vulkan/Metal; CPU skips the device check).
3. The load path is serialized even when `max_loaded_models = 0`, so concurrent lazy loads cannot both pass the pre-check.
4. macOS implementation of `engine::core::available_host_memory_bytes()`.

## Verification
- Local macOS (Metal): build + `server_config_test` passed; smoke tests for 503 refusal on low host/GPU memory, normal load, and idle auto-unload (15 s timeout).
- Cross-platform CI (new workflow in this PR): `audiocpp_server` build + `server_config_test` passed on ubuntu-24.04 / windows-2022 / macos-14.
- Config unit tests added for the new fields (defaults, overrides, negative rejection).
- Independent code review performed before submission; all blocking findings fixed.

## Notes
- `--idle-unload-ms` and `--min-free-memory-mb` both default to safe values (0 / 512) so existing deployments see no behavior change unless configured.
- CUDA note: after unload, ggml pooled memory may not be returned to the driver immediately; Metal/Vulkan release buffers directly. Consider `trim_backend_pools` on the unload path as a follow-up.
- On directory-style model paths the estimate sums the tree with depth/file limits.


gqf2008:
Hi! This is my first contribution to this repo, so the workflow runs are waiting for maintainer approval (marked `action_required`). Could you approve the runs when you get a chance?

For reference, the same matrix already passed on my fork (ubuntu-24.04 / windows-2022 / macos-14, `audiocpp_server` build + `server_config_test`), and a full code review was done before submitting. Thanks!

0xShug0:
@gqf2008 Thanks! I'm glad the Metal backend is starting to get more attention. Will review the PR this morning.

0xShug0:
@gqf2008 Some issues during testing: 

1. `min_free_memory_mb` changes the default behavior because model loading always calls `ensure_model_fits_memory`(). Existing configs that never opted into this feature may see surprising load failures.

2. The estimator recursively sums the whole model directory. This can overestimate packages that contain multiple GGUFs or variants, even when a real run would use only one selected GGUF/package choice.

It can also mask the real loader error. I validated this with a multi-GGUF Chatterbox directory:

- Main, same directory load: `500: model directory contains 2 GGUF files...`
- PR, same directory load: `503 insufficient_memory`
  - `estimated 8.27 GiB + 50000 MiB headroom exceeds available host memory`
- PR with direct Q8 file under the same config: `200 OK`

So the direct selected file can load, but the directory path is rejected earlier by the new memory estimator.

3. Relative option paths are resolved against `model.path.parent_path()`, which is correct when `model.path` is a file, but likely wrong when `model.path` is a directory.

4. `idle_unload_ms` currently appears to measure from request start/load rather than request completion. With `idle_unload_ms=1000`, a longform Chatterbox request took ~34.6s, and the server unloaded the model immediately after the response with:
`[server] idle 35721 ms: unloaded 1 model(s)`
So long requests can make the model look idle as soon as they finish.

5. Lazy loads are now globally serialized through `model_load_mutex_`, even when `max_loaded_models` == 0. It can reduce concurrency for unrelated first-load requests. E.g.,

Same two concurrent /v1/models/load requests: Chatterbox Q8 and F16.

Main:
```
WALL_MS=579
TIME_A=0.534779
TIME_B=0.574533
```

PR:
```
WALL_MS=1067
TIME_A=0.526078
TIME_B=1.063086
```

0xShug0:
@gqf2008 Thanks a lot! I’m going to merge the PR. There is still one minor issue, but it’s not a blocker. A follow-up PR would be appreciated.

For issue (2) 

My test:
Create a dummy directory containing two .gguf files:

With `min_free_memory_mb: 0,` the real loader error surfaces correctly:
```
500
model directory contains 2 GGUF files...
```
With `min_free_memory_mb: 1000000`, the PR still returns the memory guard error first:
```
503 insufficient_memory
estimated 0.12 GiB + 1000000 MiB headroom exceeds available host memory
```

gqf2008:
@0xShug0 Thanks for the thorough testing — all five points are addressed in `0082244` (now on the PR head):

**1. Guard is now opt-in.** `min_free_memory_mb` defaults to `0`, and `0` now disables `ensure_model_fits_memory` entirely (not just the headroom). Configs that never opted in see no load-behavior change.

**2. Directory estimation no longer over-sums or masks the loader error.** The estimator now measures only what the loader will read: a single file; the one GGUF a directory selects via `find_directory_gguf` (`model.gguf` or the sole `*.gguf`); a no-GGUF directory (safetensors/HF checkpoint) as a whole tree; and a multi-GGUF directory with no `model.gguf` contributes nothing to the estimate, so the loader's own `contains N GGUF files` error surfaces instead of a 503. I reproduced your scenario with a 2-GGUF dir: before → `503 insufficient_memory`; now → `model directory contains 2 GGUF files: … found: …`, with the guard both on and off. A single-GGUF dir still estimates just the selected file (e.g. `10.83 GiB` → 503 when the guard is on).

**3. Relative option paths** now resolve against `model.path` when it's a directory, and `parent_path()` only when it's a file.

**4. Idle clock measures from completion.** `run_model` and the streaming path stamp `last_activity_ms_` when a request finishes, so a long inference no longer reads as idle the moment it returns (your `idle_unload_ms=1000` / 34.6 s case).

**5. Lazy loads serialize only when a guard needs it.** `model_load_mutex_` is taken only when `max_loaded_models > 0` or `min_free_memory_mb > 0`; with both off, unrelated first loads run concurrently again (your Q8+F16 case).

CI is green across the board (Linux cpu/vulkan, Nix cpu/vulkan, Windows, macOS, and the memory-guard build+test matrix). The earlier `Nix (vulkan)` failure was a flaky ggml-vulkan shader-gen link error unrelated to this PR — the same tree builds green on vulkan on my fork. Happy to adjust anything further.

gqf2008:
Thanks for merging! The minor follow-up on issue (2) is addressed in #308: an ambiguous multi-GGUF directory now reports an indeterminate footprint, so the guard skips it and the loader's own `contains N GGUF files` error surfaces regardless of the configured headroom (verified with `min_free_memory_mb: 1000000`).

PR #308, opened 2026-08-26
fix(server): skip memory guard when the model footprint is indeterminate

gqf2008:
## Motivation

Follow-up to #306 per the post-merge review: with `min_free_memory_mb: 1000000`, a model directory holding two `.gguf` files still returned `503 insufficient_memory` (`estimated 0.12 GiB + 1000000 MiB headroom ...`) before the loader could report its own `contains 2 GGUF files` error. The estimator already contributes no weights for such an ambiguous directory, but the guard still compared the remaining fixed floor plus the configured headroom against free memory, so a large headroom masked the real loader error.

## Changes

- `estimate_model_memory_bytes` returns `std::optional<size_t>`: `std::nullopt` when `model.path` is a directory with several GGUFs and no `model.gguf` (the footprint is indeterminate).
- `ensure_model_fits_memory` skips the guard entirely when the estimate is `nullopt`, so the loader's own error surfaces no matter how large the headroom is. The skip is logged to stderr rather than silent. Determinate footprints (single file, selected GGUF, safetensors/HF tree) still guard exactly as before.
- The estimator moved to a free function in `app/server/model_memory.{h,cpp}` so it is unit-testable, and now lists the directory once (mirroring the loader's selection: `model.gguf` wins, the sole `*.gguf` is used alone) instead of two passes.
- Unreadable file sizes now contribute 0 instead of being folded in as failures.
- `server_config_test` gains estimator cases: single file, sole GGUF, `model.gguf` disambiguation, ambiguous directory → `nullopt`, checkpoint tree sum, relative aux resolution.
- Fixed the stale `--min-free-memory-mb` help text (still said "default 512").

## Known limitation

The ambiguity skip is deliberately permissive: model families with custom directory layouts (e.g. `minimax_music3`, which assembles several named GGUFs) may load a multi-GGUF directory successfully, and the guard then makes no estimate for it — the load proceeds unguarded. There is no loader-level signal to distinguish those layouts from the spec-driven ones the loader rejects, so fully covering them would need per-family layout metadata in the loader interface; left for a follow-up if you want it.

## Verification

macOS (Apple M4), real GGUF symlinks:

- 2-GGUF dir, `min_free_memory_mb: 1000000` → `model directory contains 2 GGUF files: ...` plus `[server] memory guard skipped for model ...` (was `503 insufficient_memory`).
- single-GGUF dir, `min_free_memory_mb: 1000000` → still `503 insufficient_memory` (guard intact for determinate footprints).
- `server_config_test` passes (including the new estimator cases).
- Cross-platform CI: ubuntu / windows / macos build + test green on my fork and on this PR.


0xShug0:
@gqf2008 Thanks! PR merged.

PR #319, opened 2026-08-27
Accept PCM8/32, float64, A-law, mu-law and WAVEFORMATEXTENSIBLE in the WAV reader

5uck1ess:
Follow-up to #180, split out at your request.

The reader handled PCM16, PCM24 and float32. Everything else — including ordinary PCM16 that happens to carry the `0xFFFE` tag — came back as `unsupported WAV encoding (need PCM16, PCM24, or float32)`, which is a confusing thing to read when PCM16 is exactly what is inside the file. Encoders emit WAVEFORMATEXTENSIBLE routinely for more than two channels or whenever a channel mask is set, and the real format tag then lives in the SubFormat GUID rather than in `wFormatTag`.

## What changes

- **WAVEFORMATEXTENSIBLE** (`0xFFFE`) unwraps to whatever its GUID names
- **PCM8** (unsigned, biased by 128 — the sign convention differs from every other PCM width), **PCM32**, **float64**
- **G.711 A-law and mu-law**
- Files that still cannot be decoded now say what they are: a FLAC, Ogg, MP3, MP4/M4A, AIFF, RF64 or CAF handed to the reader is named as such instead of failing with `invalid WAV RIFF header`

No behaviour change for PCM16/24 or float32.

## Origin

The decoders are @dignome's, from the tree that arrived at `fa8206c` in #180 — they existed so the CLI and server path would accept the same files the web UI already did. That was out of scope for Echo, so it was reverted out and is landing here on its own merits instead.

## Tests

The original had none; this adds 213 lines covering each added format.

The G.711 cases are pinned to the **published** decode values — mu-law `0x00` → -32124 and `0x80` → +32124, both `0x7F` and `0xFF` → 0, A-law `0x55` → +8, `0xD5` → -8, `0x2A` → +32256, including A-law's absence of an exact zero. Re-deriving the same bit manipulation in the test and comparing it to itself would prove nothing.

Two negative cases check that widening the accepted set did not turn into accepting everything: ADPCM is still rejected with the unsupported-encoding message, and a FLAC is still identified as a FLAC.

## Verification

Run by execution, not by inspection:

- The new cases **fail** against the current reader with the old message, and **pass** against this one — so they are not vacuous
- `wav_reader_chunk_bounds_test` still passes; the allocation bound from #143 is untouched
- Full `ctest` green (59/59) on Linux/CUDA
- End to end: a hand-built `0xFFFE`-tagged PCM16 file, which the current reader rejects, works as an `--voice-ref` through `audiocpp_cli`

### DIFF
--- a/src/framework/audio/wav_reader.cpp
+++ b/src/framework/audio/wav_reader.cpp
@@ -76,6 +76,13 @@ constexpr uint16_t kFormatALaw = 0x0006;
 constexpr uint16_t kFormatMuLaw = 0x0007;
 constexpr uint16_t kFormatExtensible = 0xFFFE;
 
+// Bytes 2..15 of every KSDATAFORMAT_SUBTYPE_* GUID:
+// XXXXXXXX-0000-0010-8000-00aa00389b71.
+constexpr std::array<char, 14> kKsDataFormatSubtypeTail = {
+    0x00, 0x00, 0x00, 0x00, 0x10, 0x00, static_cast<char>(0x80),
+    0x00, 0x00, static_cast<char>(0xAA), 0x00, 0x38, static_cast<char>(0x9B), 0x71,
+};
+
 // Names a container we can recognise but not decode, so the error can say what
 // the file actually is instead of "invalid WAV RIFF header".
 const char * identify_foreign_container(const std::array<char, 12> & header) {
@@ -124,7 +131,11 @@ float decode_mu_law(uint8_t value) {
 
 float decode_a_law(uint8_t value) {
     value ^= 0x55;
-    const int sign = (value & 0x80) != 0 ? -1 : 1;
+    // Note the inversion relative to mu-law above: in A-law the sign bit marks a
+    // POSITIVE sample. Getting this backwards is silent -- the audio decodes at
+    // the right amplitude, just phase-inverted -- so it is pinned by an
+    // exhaustive 256-code table in the tests rather than by spot checks.
+    const int sign = (value & 0x80) != 0 ? 1 : -1;
     const int exponent = (value >> 4) & 0x07;
     const int mantissa = value & 0x0F;
     int magnitude = 0;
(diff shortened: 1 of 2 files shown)


5uck1ess:
Pushed two rounds of fixes after review. Six defects, all in code this PR added — worth writing up since one of them was invisible by construction.

**A-law was sign-inverted on all 256 codes.** G.711 sets the sign bit for *positive* samples in A-law and *negative* in mu-law; I treated both the same way. Decoded audio came out at the correct amplitude, phase-inverted — inaudible on its own.

It survived review because my tests were circular. I'd written that the expected values were "pinned to the published G.711 values" and that re-deriving the same bit manipulation "would prove nothing" — then derived the four anchors from exactly that arithmetic. They confirmed the bug instead of catching it.

Both codings are now checked against full 256-entry tables produced by ffmpeg 9.0.1 and cross-checked against the ITU-T G.711 segment definitions. Reinstating the old sign fails on `A-law code 0`.

**The extensible SubFormat GUID was trusted on its first two bytes.** Only those carry the format tag; the other fourteen are a fixed suffix shared by every `KSDATAFORMAT_SUBTYPE_*`. Any codec whose GUID merely started `0x0001` decoded as PCM16. Now compared, with `cbSize >= 22` enforced.

**PCM32 and float64 silently dropped a trailing partial sample**, so a truncated download decoded as valid audio. Both reject it now, matching PCM24. PCM16 and float32 keep their existing behaviour — tightening those isn't this PR's business.

**The 12-byte header sniff rewound absolutely.** `read_wav_f32(std::istream &)` is public, and given a pipe or a stream not starting at offset 0 it failed with `incomplete WAV file` on a valid WAV. The bytes are already consumed, so the rewind is gone.

**The RIFF pad byte was required at EOF.** Seeking past the end is legal on an `ifstream` but not on the buffer behind the `string_view` overload — so the same bytes parsed from disk and threw from an upload. This PR is what makes it bite: PCM8, A-law and mu-law are one byte per sample, so odd data chunks go from rare to routine.

**A `fmt ` chunk under 16 bytes** read on into whatever followed. Now rejected.

The PCM8/PCM32/float64 expectations were still derived from the implementation — same construction that hid the A-law bug — so they're frozen from `ffmpeg -f f32le` and extended to the endpoints that separate a correct conversion from a plausible one: `INT32_MAX`, which float32 rounding maps to exactly 1.0, and a float64 value not representable in float32.

**Verification.** Every fix has a test that fails against the code as it stood before it. Full ctest green (59/59). Our decode of a real ffmpeg-transcoded A-law file matches ffmpeg on all 153280 samples, zero mismatches.

Reviewed with two models from different vendors; the second found the two parse-loop bugs the first missed, so a third pair of eyes on the chunk iteration wouldn't be wasted.

0xShug0:
Thanks @5uck1ess! PR merged.

PR #328, opened 2026-08-28
server: resolve the model contract once per model, not per request

XythQ:
### Problem

`ServerState::build_speech_request` calls `model_accepts_request_option()` on every
`POST /v1/audio/speech`. That resolves the model contract through
`engine::model_spec::model_contract(family)`, which reads the model's embedded spec with
`gguf_init_from_file()` on the full GGUF.

The resolution cache is `thread_local` (`src/framework/model_spec/package.cpp:22-26`), and
`serve_http` handles each request on a fresh detached thread, so the cache is cold every time. Each
request re-opens and re-parses the whole model file.

With a 298 MB GGUF this adds a fixed **~900 ms** to every request, independent of backend, text
length and audio duration. It is invisible in `X-AudioCPP-Wall-Ms` because `run_model` starts its
clock after `build_speech_request`, and it does not affect `audiocpp_cli`, which resolves the spec
once on a single thread.

### Fix

Resolve the option once per model instead of once per request:

- `LoadedModel` gains `bool accepts_reference_text` (defaults to `true`, the existing permissive
  behaviour for models with no contract).
- `refresh_model_option_flags()` sets it in `make_model`, and again in `handle_model_load` when a
  model is reconfigured, so dynamic model management stays correct.
- `build_speech_request` reads the cached flag.

A contract that fails to resolve is caught at registration, logged to `stderr`, and falls back to
the permissive default — same behaviour as before, but reported once instead of throwing per
request.

The one-off resolution cost moves to registration, where the model load already performs a
`gguf_init_from_file`.

### Measurements

`v0.7.0`, Windows x64, Vulkan backend, `supertonic-3-f16.gguf` (298 MB), same `server.json`, same
request:

| | ttfb | `X-AudioCPP-Wall-Ms` |
|---|---|---|
| before | 0.956–1.153 s | 44–176 ms |
| after | **0.045–0.049 s** | 44–50 ms |

Also reproduced before the fix on `--backend cpu` (ttfb 1.268–1.414 s against 341–362 ms reported),
confirming it is not backend-specific.

Response bytes are unchanged: the same request returns `Content-Length` 114090 with an identical
MD5 (`1514cf1fcf6bfc4e865a1cc24d5b8948`) before and after.

### Notes

- Long utterances now scale with synthesis time rather than sitting on a fixed floor.
- This caches the `reference_text` lookup, the only contract option the speech path uses today.
  Any other per-request contract lookup added later would hit the same cold `thread_local` path, so
  it may be worth resolving the contract once per model more generally.

### DIFF
--- a/app/server/runtime.cpp
+++ b/app/server/runtime.cpp
@@ -1180,7 +1180,29 @@ std::unique_ptr<ServerState::LoadedModel> ServerState::make_model(ServerModelCon
         engine::runtime::parse_run_mode(loaded->config.mode),
     };
     load_voice_presets(*loaded);
+    refresh_model_option_flags(*loaded);
     return loaded;
+}
+
+void ServerState::refresh_model_option_flags(LoadedModel & model) {
+    const auto effective_override = model.config.model_spec_override.has_value()
+        ? model.config.model_spec_override
+        : config_.model_spec_override;
+    try {
+        model.accepts_reference_text = model_accepts_request_option(
+            model.config.family,
+            "reference_text",
+            effective_override,
+            model.config.path);
+    } catch (const std::exception & ex) {
+        // Keep registration robust: a model whose contract cannot be resolved gets
+        // the same permissive behavior as a model with no contract; the failure is
+        // reported so the misconfiguration stays visible.
+        std::cerr << "[server] model '" << model.config.id
+                  << "': reference_text option check failed (" << ex.what()
+                  << "); assuming the option is accepted\n";
+        model.accepts_reference_text = true;
+    }
 }
 
 HttpResponse ServerState::handle_model_load(const std::string & body_text) {
@@ -1221,6 +1243,7 @@ HttpResponse ServerState::handle_model_load(const std::string & body_text) {
                 engine::runtime::parse_run_mode(existing->config.mode),
             };
             load_voice_presets(*existing);
+            refresh_model_option_flags(*existing);
         }
         ensure_model_loaded_locked(*existing);
         return json_response(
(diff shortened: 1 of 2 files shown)


0xShug0:
@XythQ Thanks, this is a useful improvement! One small issue before merge: the new fallback swallows real model-spec errors.

With an invalid spec override, startup logs the real error but continues permissively:
```
[server] model 'bad-contract-smoke': reference_text option check failed (model spec '/tmp/audiocpp_pr328_bad_spec.json'.schema_version: expected 1); assuming the option is accepted
audiocpp_server listening on http://127.0.0.1:18090
```

Then the request fails later with an unrelated error:
```
HTTP/1.1 500 Internal Server Error
{"error":{"message":"model path does not exist: /tmp/nonexistent-supertonic-smoke.gguf","type":"server_error"}}
```

A better approach may be to remove the broad catch and let invalid specs fail during registration or load.

XythQ:
Removed rather than narrowed — `model_accepts_request_option` already returns `true` for a model with no contract and rethrows the rest (`runtime.cpp:107`), so the catch only ever covered real misconfigurations. An invalid spec now throws at registration and reaches `main.cpp:235` as `audiocpp_server failed: <error>`, exit 1; a throw during dynamic load hits the existing handler at `http.cpp:777`.

One consequence: a configured-but-absent override file now fails loudly. Intentional, but I didn't build to reproduce your case — the log line traces statically to `fail()` at `schema.cpp:16-18`.

0xShug0:
@XythQ Thanks! PR merged :rocket:

PR #330, opened 2026-08-28
Add optional server frontend pipeline

### LINKED ISSUE #313: [Feature Request / Bug]: Support non-WAV / browser audio uploads (WebM/Opus, MP3, OGG) on `/v1/audio/transcriptions` for Open WebUI compatibility
### Description

Currently, the `POST /v1/audio/transcriptions` endpoint strictly expects raw WAV PCM audio uploads (`RIFF....WAVE` header). 

When integrating `audio.cpp` with popular web clients like **Open WebUI**, transcription fails because modern web browsers (Chrome, Edge, Firefox, Safari) using the HTML5 `MediaRecorder` API record microphone input in **`audio/webm;codecs=opus`** (or `audio/ogg`). Browsers do not support recording native WAV via `MediaRecorder`.

As a result, sending microphone recordings from Open WebUI to any ASR model (e.g. `parakeet_tdt`, `granite5asr`, `whisper`) fails with:
`Error transcribing chunk: External: 500, message='Internal Server Error', url='http://:8080/v1/audio/transcriptions'`

In contrast, clients that buffer and convert audio to WAV beforehand (such as Home Assistant via `wyoming_openai` or `curl -F file=@sample.wav`) succeed with exceptional sub-30ms latency.

---

### Steps to Reproduce

1. Start `audiocpp_server` with any ASR model (e.g., `parakeet_tdt` or `granite5asr`):
   ```json
   {
     "models": [
       {
         "id": "parakeet_tdt",
         "family": "parakeet_tdt",
         "path": "/models/parakeet-tdt-0.6b-v3-q8_0.gguf",
         "task": "asr",
         "mode": "offline"
       }
     ]
   }

2. In Open WebUI (Admin Settings ➔ Audio):
- Set Speech-to-Text Engine: OpenAI
- Set API Base URL: http://<server-ip>:8080/v1
- Set STT Model: parakeet_tdt
3. Click the microphone button in Open WebUI chat to record speech.
4. Open WebUI fails with HTTP 500.

### Root Cause
handle_transcription_multipart in app/server/runtime.cpp checks is_wav_upload_filename and directly parses bytes using engine::audio::read_wav_f32(), rejecting .webm, .mp3, .ogg, and .m4a files.
When browsers send MediaRecorder chunks or blobs (often named audio.webm or blob), the server throws an unhandled exception or returns a 400/500 error instead of decoding the compressed audio stream.

## Suggested Fixes & Solutions
### Solution 1: Lightweight In-Tree Decoders (Recommended)
Embed header-only C/C++ audio decoders (similar to how ggml or llama.cpp bundle dependencies) so audio.cpp remains self-contained without external runtime dependencies:

miniaudio (Public Domain / MIT, single-file) — handles MP3, FLAC, WAV, and Vorbis decoding natively.
opusfile / libopus — decodes WebM/Ogg Opus streams directly to 16 kHz float32 PCM in memory.

### Solution 2: Optional Server-Side Transcoder Bridge
If external tools are acceptable when present, allow audiocpp_server to detect and pipe incoming non-WAV multipart bytes through ffmpeg or libavcodec to convert to 16 kHz mono WAV PCM in memory before running inference.

### Solution 3: Graceful Handling of Empty / Incomplete Chunks
When MediaRecorder begins recording, the initial chunk may sometimes be 0 bytes or header-incomplete. Returning {"text": ""} (HTTP 200) instead of throwing an unhandled exception ensures the client does not crash before the full audio is delivered.

### LINKED ISSUE #392: [feature] https support for server (TLS with SSL cert)
It'd be nice for `audiocpp_server` to be able to serve HTTPS. It would need configuration options to pass in paths to the certificate and keyfile.

0xShug0:
Resolves https://github.com/0xShug0/audio.cpp/issues/313
Resolves https://github.com/0xShug0/audio.cpp/issues/392: Adds opt-in frontend-scoped HTTPS support

## Summary
- add an opt-in in-process server frontend pipeline with pre/post module hooks
- add explicit module contracts for request and response envelope states
- add `audio_decode` for MP3/FLAC ASR input via miniaudio
- add `mp3_encode` for `response_format=mp3` TTS output via libmp3lame
- add `https` as a frontend capability, not a pre/post module, using the existing vendored `cpp-httplib` TLS stack
- keep frontend dependencies opt-in; the default server build does not include codec/TLS frontend dependencies
- split frontend docs into an overview plus module-specific docs, including Windows/Linux LAME setup for `mp3_encode`

## High-Level Design
The core server keeps its stable HTTP API shape: WAV/text in and WAV/text out. Optional frontend modules run in-process around that core handler. Pre-processing modules adapt client requests before core routing, and post-processing modules adapt core responses before returning to the client.

`https` is separate from the pre/post module contract. It is a frontend-owned listener capability that terminates TLS in-process and forwards the request envelope to the same core handler. The existing plain HTTP transport is unchanged.

At configure time, users select only the frontend capabilities/modules they need:

```bash
cmake -S . -B build/debug \
  -DCMAKE_BUILD_TYPE=Debug \
  -DAUDIOCPP_BUILD_SERVER_FRONTENDS=ON \
  '-DAUDIOCPP_SERVER_FRONTEND_MODULES=audio_decode;mp3_encode;https'
```

## Docs
- `app/server/frontends/README.md`: frontend overview, supported frontend table, pipeline, and contracts
- `app/server/frontends/docs/audio_decode.md`: MP3/FLAC ASR input module
- `app/server/frontends/docs/mp3_encode.md`: MP3 TTS output module and LAME install/configuration notes
- `app/server/frontends/docs/https.md`: HTTPS frontend listener capability
- `app/server/frontends/docs/adding_modules.md`: developer guide for adding new frontend modules

## Validation

| Case | Request / command shape | Response / behavior |
|---|---|---|
| Default build, no frontend | `AUDIOCPP_BUILD_SERVER_FRONTENDS=OFF`, build `audiocpp_server` | Build passed |
| HTTPS flags on non-HTTPS build | `--https-cert-file ... --https-key-file ...` | Exit 1: `HTTPS frontend support is not available in this build; configure with -DAUDIOCPP_BUILD_SERVER_FRONTENDS=ON -DAUDIOCPP_SERVER_FRONTEND_MODULES=https` |
| HTTPS frontend build | `AUDIOCPP_BUILD_NATIVE_MODEL_MANAGER=OFF`, `AUDIOCPP_BUILD_SERVER_FRONTENDS=ON`, `AUDIOCPP_SERVER_FRONTEND_MODULES=https` | Build passed; HTTPS uses frontend-selected `audiocpp_cpp_httplib` without native model manager |
| CLI missing key pair | `--https-cert-file cert.pem` only | Exit 1: `--https-cert-file and --https-key-file must be set together` |
| Bad cert/key files | cert/key paths do not exist | Exit 1: `could not initialize HTTPS frontend server from certificate/key files` |
| HTTPS `/health` | `GET https://127.0.0.1:18443/health` | `200 OK`, `{"status":"ok","backend":"cpu","models":0,"ui":true,"ui_management":false}` |
| HTTPS UI root | `GET https://127.0.0.1:18443/` | `200 OK`, `Content-Type: text/html; charset=utf-8`, embedded UI returned |
| HTTPS models list | `GET https://127.0.0.1:18443/v1/models` | `200 OK`, `{"object":"list","data":[]}` |
| Plain HTTP to HTTPS port | `GET http://127.0.0.1:18443/health` | Fails before handler routing; curl exit `52`, `Empty reply from server` |
| Config relative cert/key | `https_cert_file: "cert.pem"`, `https_key_file: "key.pem"` in config directory | `GET /health` over HTTPS returned `200 OK`; config-relative path resolution works |
| Config model list | Same config with one lazy dummy model | `GET /v1/models` returned one unloaded model entry |
| Config missing key pair | `https_cert_file` without `https_key_file` | Exit 1: `server https_cert_file and https_key_file must be set together` |
| HTTPS oversized request body | `max_request_body_bytes: 8`, POST body larger than 8 bytes | `413 Payload Too Large` before core routing |
| HTTPS-capable binary without TLS flags | `GET http://127.0.0.1:18080/health` | `200 OK`, normal HTTP path still works |
| HTTPS request to plain HTTP port | `GET https://127.0.0.1:18080/health` | Fails before handler routing; curl exit `28`, connection timed out |
| `mp3_encode` dependency lookup without LAME | Configure with `AUDIOCPP_SERVER_FRONTEND_MODULES=mp3_encode` and no LAME install | Fails with cross-platform message naming `lame/lame.h`, `libmp3lame`, conda, and `AUDIOCPP_LAME_ROOT` |
| HTTPS configure after LAME lookup change | Configure with `AUDIOCPP_SERVER_FRONTEND_MODULES=https` | Configure passed; unrelated frontend capability not affected |
| Frontend docs split | Checked doc links and Markdown paths | Links point to module-specific docs |

Also checked:
- `git diff --check`
- no changes to `app/server/http.cpp` or `app/server/http.h`
- no server process left running after validation


keturn:
Confirmed that running with `--https-cert-file` and `--https-key-file` works as intended for #392.

That is, once I enabled the right build options. `.devops/` and `scripts/` need updating if this is to be included in builds by default.

0xShug0:
> if this is to be included in builds by default.

Thanks for testing! The frontend may go into the default build. Check the discussions here https://github.com/0xShug0/audio.cpp/issues/313

0xShug0:
Added temporary `MP3_FRONTEND_DEBUG` traces to diagnose the MP3 frontend report.

0xShug0:
Rebased this branch onto latest main and kept the frontend split as a separate submodule under `external/audio.cpp-server-frontends`. The server frontend integration remains opt-in: core server builds do not pull frontend module sources/dependencies unless the frontend build flags are enabled.

PR #323, opened 2026-08-27
feat: add Soprano TTS as community model with GGUF packages

drzsdrtfg:
## Soprano TTS — Community Model

Soprano is an ultra-lightweight (~80M parameter) English-only text-to-speech model using a two-stage architecture: a Qwen3-style causal LM (17 layers, hidden 512, vocab 8192) that autoregressively emits per-frame 512-dimensional features, and a non-iterative Vocos-style decoder (ConvNeXt backbone + single ISTFT head, n_fft 2048 / hop 512) that turns those features into 32 kHz audio.

Reference: https://github.com/ekwek1/soprano
Weights: https://huggingface.co/ekwek/Soprano-1.1-80M
GGUF packages: https://huggingface.co/WalkingCat/Soprano-1.1-80M-GGUF

### Files added

| Path | Purpose |
|---|---|
| `src/community_models/soprano_tts/` (5 .cpp) | Assets, Qwen3 LM generator, Vocos decoder, tokenizer, session |
| `include/engine/community_models/soprano_tts/` (5 .h) | Corresponding headers |
| `model_specs/soprano_tts.json` | Schema-v1 spec, community status, GGUF + safetensors sources |
| `docs/soprano_tts.md` | User-facing documentation |
| `docs/soprano_validation.md` | Validation record with build/run commands and timing |
| `tests/soprano_tts/soprano_warm_bench.cpp` | C++ warmbench binary |
| `tests/soprano_tts/soprano_warm_bench_cases.txt` | Warmbench test cases (short/medium/long/longform) |
| `tests/soprano_tts/soprano_python_warm_bench.py` | Python reference warmbench |
| `tools/soprano_tts/convert_soprano.py` | HF checkpoint converter (folds weight-norm) |
| `tools/soprano_tts/run_official.py` | Python reference inference runner |
| `tools/soprano_tts/compare_parity.py` | Automated validation harness |

### Files modified

| File | Change |
|---|---|
| `CMakeLists.txt` | +16 lines: `audiocpp_add_model` + `add_engine_warmbench` |
| `README.md` | +2 lines: community table row |
| `docs/gguf.md` | +1 line: GGUF status table row |
| `webui/configs/models_catalog.json` | +1 line: WebUI catalog entry |
| `webui/native/dist/index.html` | Rebuilt frontend with Soprano baked in |

### Build

```bash
# Soprano only
cmake -B build -DCMAKE_BUILD_TYPE=Release \
[...]

PR #331, opened 2026-08-29
fix: re-feed persistent prefill inputs on cached-graph recompute in QwenCausalDecodeRuntime

drzsdrtfg:
### NeuTTS (`prefill_tokens`, token-prompt, persistent runtime - recompute reachable)

Two identical utterances in one session (batch-text-file x2):

| | pre-fix (81ef451) | post-fix (ce1b8bc) |
|---|---|---|
| first utterance | 1.84s audio | 1.84s audio |
| second utterance | **12.2s garbage** (NaN divergence) | **1.84s deterministic, healthy** |
| first-run hash parity post vs pre | identical (F7CDE40A...) | identical |
| cross-run determinism (post-fix) | first utterance byte-identical across separate runs | |

NeuTTS applies an intentional per-chunk seed offset for utterance 2, so the two
lines legitimately differ from each other; the point is the pre-fix second line
is 12.2s of garbage while post-fix it is healthy and reproducible.

### DotTTS (`prefill_embeddings`, persistent runtime)

Two identical lines, `template_name=tts`, fixed seed. Pre-fix and post-fix are
**byte-identical and healthy** in both lines (278A7D17...): the DotTTS path does
not trigger the gallocr reuse, confirming the fix is a behavioral no-op where
the corruption cannot occur.

### Soprano (surfaced the bug)

PR #323 branch (which carried this fix until the split): `--warmup 1
--iterations 3` passes on CPU and Vulkan; three identical seeded generations
are bit-identical (previously fatal `no finite logits`). Soprano is not in
upstream yet, so it was validated on the PR branch carrying the same fix code.

### Compile matrix

Custom build including all runtime consumers (neutts, dots_tts, fireredtts3,
midashenglm_gen, minimax_music3, qwen3_asr) links cleanly - the header is
unchanged, so no API breakage.

### Consumers and why the rest are low-risk

- **qwen3_asr: not a consumer** - it builds its own prefill graphs and only
  mentions QwenCausalDecodeRuntime in a comment. Removed from the affected list.
- **fireredtts3 / midashenglm_gen / minimax_music3 / redae**: each prefill call
  creates a fresh runtime (graph built once, not recomputed) - the re-feed is a
  no-op for them. miniMax-H3 needs 15GB+ for a runtime test (skipped).
- The re-feed re-uploads byte-identical position/mask values, mirroring what the
  decode step already does every token; it is backend-agnostic.

### Caveats

- The CLI batch/repeat path did not discriminate for qwen3_asr or the DotTTS
  loop pre-fix (graph branch reused); NeuTTS two-line and Soprano
  warmup/iterations are the discriminating repeat proofs.
- CUDA backends were not exercised (no NVIDIA GPU here); the fix is a generic
  `ggml_backend_tensor_set` before compute, so no backend-specific surprises
  are expected, but CUDA CI coverage would be valuable.

### DIFF
--- a/src/framework/modules/transformers/qwen_causal_decode_runtime.cpp
+++ b/src/framework/modules/transformers/qwen_causal_decode_runtime.cpp
@@ -747,18 +747,18 @@ private:
             !ggml_gallocr_alloc_graph(prefill_gallocr_, prefill_graph_)) {
             throw std::runtime_error("failed to allocate QwenCausalDecodeRuntime prefill graph");
         }
-        const auto positions_values = qwen_position_ids(steps);
+        prefill_positions_values_ = qwen_position_ids(steps);
         ggml_backend_tensor_set(
             prefill_positions_,
-            positions_values.data(),
+            prefill_positions_values_.data(),
             0,
-            positions_values.size() * sizeof(int32_t));
-        const auto mask = prefill_attention_mask_values(config_, 1, steps);
+            prefill_positions_values_.size() * sizeof(int32_t));
+        prefill_attention_mask_values_ = prefill_attention_mask_values(config_, 1, steps);
         ggml_backend_tensor_set(
             prefill_attention_mask_,
-            mask.data(),
+            prefill_attention_mask_values_.data(),
             0,
-            mask.size() * sizeof(ggml_fp16_t));
+            prefill_attention_mask_values_.size() * sizeof(ggml_fp16_t));
         if (prefill_logits_readback_token_ids_ != nullptr) {
             upload_logits_readback_token_ids(prefill_logits_readback_token_ids_, config_);
         }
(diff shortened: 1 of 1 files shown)


0xShug0:
@drzsdrtfg I can't reproduce the issue for NeuTTS on CUDA, while on CPU the issue is real. That’s interesting. I need some time to understand this bug...

[neutts_cpu_1.wav](https://github.com/user-attachments/files/31601250/neutts_cpu_1.wav)
[neutts_cpu_2.wav](https://github.com/user-attachments/files/31601251/neutts_cpu_2.wav)
[neutts_cuda_1.wav](https://github.com/user-attachments/files/31601253/neutts_cuda_1.wav)
[neutts_cuda_2.wav](https://github.com/user-attachments/files/31601254/neutts_cuda_2.wav)

drzsdrtfg:
Ok,

0xShug0:
@drzsdrtfg PR meged. Thanks!

PR #339, opened 2026-08-29
fail when the cli is given an unknown option

### LINKED ISSUE #338: audiocpp_cli silently ignores unknown options
a misspelled flag is dropped without a word. `--bakend cuda` runs on cpu and exits 0, so you believe you got a cuda run and nothing says otherwise.

built at `a76ec04` on macos arm64, `cmake -DAUDIOCPP_MODEL_SET=custom -DAUDIOCPP_MODELS=citrinet_asr`

```bash
audiocpp_cli --list-loaders --bakend cuda
# registered_loaders=3
# citrinet_asr: asr (offline)
# marblenet_vad: vad (offline)
# silero_vad: vad (offline|streaming)
# exit 0, no warning

audiocpp_cli --list-loaders --this-flag-does-not-exist
# same output, exit 0

audiocpp_cli --task asr --family citrinet_asr --model <path> --sesion-option a=b
# the session option is silently dropped
```

a wrong *value* is caught, `--backend notreal` fails with `unsupported backend: notreal`. it is only the option name that goes unchecked.

cause is in `app/cli/args.cpp`: `find_arg`, `has_arg` and `collect_args` each scan argv for the one name they expect, and nothing ever looks at what is left over, so an unmatched token just disappears.

the other binary in this repo already does the opposite. `app/model_manager/main.cpp:61` walks its arguments and throws `unknown option: ` on anything it does not recognise, so the two CLIs disagree on this today.

happy to send the patch, one question first. do you want a warning on stderr that leaves the exit code and stdout alone, or a hard error? a hard error is stricter but would break any script that passes a stray flag today, so i have a warning working locally and can switch it either way.

avionicharshit-byte:
fixes #338. a misspelled option is currently dropped without a word, so `--bakend cuda` runs on cpu and exits 0.

`find_arg`, `has_arg` and `collect_args` record the name they were asked for, and `require_known_args` refuses anything on the command line that nobody asked about. **there is no second list of flags to keep in sync**, a new option is recognised the moment its lookup is added.

value-taking and boolean lookups are recorded separately because only the first kind consumes the next argument. without that split a typo sitting after a boolean flag gets swallowed as its value, and a value that happens to start with `--` gets reported as an option. both were real, both are covered below.

**it is a hard error, and it runs before the model does any work.** the message goes to stderr, stdout stays empty and the exit code is 1. checking before the run rather than after it means every option has to be read before the command branches, which is what most of this diff is: the output paths, `--input-format`, `--json` and the batch and workflow options are now read up front and passed down, instead of being read halfway through or after the session has already finished.

two real bugs fell out of doing that. `has_vad_chunk_option` short circuited on `||`, so three `--vad-chunk-*` options were never looked up and got reported as unknown. seven more options are only read on a single code path, so they were rejected on every other one. both are fixed, and all 98 options the cli looks up are swept below with no false rejects.

built at `e73c980` on macos arm64, `cmake -DAUDIOCPP_MODEL_SET=custom -DAUDIOCPP_MODELS=citrinet_asr --target audiocpp_cli`:

| command | before | after |
|---|---|---|
| `--task asr --family citrinet_asr --model <path> --audio speech.wav` | exit 0, 3.1s | exit 0, 3.1s, identical stdout |
| `--task asr ... --bakend cuda` | ran on cpu, exit 0 | `unknown option: --bakend`, exit 1, 0.13s, no stdout |
| `--task asr ... --sesion-option a=b` | silent, exit 0 | `unknown option: --sesion-option`, exit 1 |
| `--task asr ... --foo --bar` | silent, exit 0 | `unknown options: --foo --bar`, exit 1 |
| `--task asr ... --metrics --foo` | silent, exit 0 | `unknown option: --foo`, exit 1 |
| `--task asr ... --log-file --weird-looking-value` | exit 0 | exit 0, the value is not an option |
| `--task asr ... --backend cpu --threads 2` | exit 0 | exit 0 |
| `--list-loaders` | silent, exit 0 | silent, exit 0 |
| `--list-loaders --bakend cuda` | silent, exit 0 | `warning: ignored option: --bakend`, exit 0 |

parity against main: stdout is byte for byte identical on every valid command tested, covering `--list-loaders`, `--list-loaders --json`, `--list-devices`, `--list-pipelines`, `--help`, `--help --task asr`, `--inspect`, offline asr, offline vad, vad chunk outputs, batch audio dir, streaming and stdin pcm. output files written by `--out-dir`, `--text-out`, `--words-out`, `--segments-out`, `--turns-out`, `--vad-chunks-out` and `--batch-manifest-out` are identical too.

what this does not do: the informational commands return before the rest of the options are read, so on `--help`, `--list-*` and `--inspect` the cli cannot tell a misspelling from an option that simply does not apply to that command. rejecting both would break anything that runs `--list-devices --backend cuda` today, so those stay a warning and exit 0. making them strict as well needs the whole option set parsed up front before dispatch, which is a bigger change than this one and easy to do as a follow up.

### DIFF
--- a/app/cli/args.cpp
+++ b/app/cli/args.cpp
@@ -1,10 +1,72 @@
 #include "args.h"
 
+#include <iostream>
+#include <set>
 #include <stdexcept>
 
 namespace minitts::cli {
 
+namespace {
+
+// Every option name the CLI looked for, split by whether the lookup consumes the following
+// argument. Populated by the lookups below, so a new flag is recognised the moment its lookup
+// is added and there is no second list to keep in sync.
+std::set<std::string> & value_names() {
+    static std::set<std::string> names;
+    return names;
+}
+
+std::set<std::string> & flag_names() {
+    static std::set<std::string> names;
+    return names;
+}
+
+void record_value_query(const std::string & name) {
+    value_names().insert(name);
+}
+
+void record_flag_query(const std::string & name) {
+    flag_names().insert(name);
+}
+
+}  // namespace
+
+void report_unused_args(int argc, char ** argv) {
+    const auto & values = value_names();
+    const auto & flags = flag_names();
+    std::vector<std::string> unused;
+    for (int i = 1; i < argc; ++i) {
+        const std::string token = argv[i];
+        if (token.rfind("--", 0) != 0 || token == "--") {
+            continue;
+        }
+        // Only an option that takes a value consumes the next argument, so only then can the
+        // next argument be something that merely looks like an option.
+        if (values.count(token) != 0) {
+            ++i;
+            continue;
+        }
+        if (flags.count(token) != 0) {
+            continue;
+        }
+        unused.push_back(token);
+    }
+    if (unused.empty()) {
+        return;
+    }
+    std::cerr << "audiocpp_cli warning: ignored unknown option";
+    if (unused.size() > 1) {
+        std::cerr << "s";
+    }
+    std::cerr << ":";
+    for (const auto & token : unused) {
+        std::cerr << " " << token;
+    }
+    std::cerr << "\n";
+}
+
 std::optional<std::string> find_arg(int argc, char ** argv, const std::string & name) {
+    record_value_query(name);
     for (int i = 1; i + 1 < argc; ++i) {
         if (argv[i] == name) {
             return std::string(argv[i + 1]);
@@ -14,6 +76,7 @@ std::optional<std::string> find_arg(int argc, char ** argv, const std::string & 
 }
 
 bool has_arg(int argc, char ** argv, const std::string & name) {
+    record_flag_query(name);
     for (int i = 1; i < argc; ++i) {
         if (argv[i] == name) {
             return true;
@@ -23,6 +86,7 @@ bool has_arg(int argc, char ** argv, const std::string & name) {
 }
 
 std::vector<std::string> collect_args(int argc, char ** argv, const std::string & name) {
+    record_value_query(name);
     std::vector<std::string> values;
     for (int i = 1; i + 1 < argc; ++i) {
         if (argv[i] == name) {
--- a/app/cli/args.cpp
+++ b/app/cli/args.cpp
@@ -29,12 +29,10 @@ void record_flag_query(const std::string & name) {
     flag_names().insert(name);
 }
 
-}  // namespace
-
-void report_unused_args(int argc, char ** argv) {
+std::vector<std::string> args_nobody_asked_for(int argc, char ** argv) {
     const auto & values = value_names();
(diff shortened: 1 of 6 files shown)


0xShug0:
@avionicharshit-byte Good catch, and thanks for the fix!

For warning vs hard failure: I think this should be a hard failure instead of a warning. Could you update the PR?

avionicharshit-byte:
@0xShug0 done , its a hard error now

`audiocpp_cli failed: unknown option: --bakend` on stderr , exit 1 , stdout stays empty

the check also moved ahead of the run rather than after it , a typo now fails in 0.13s instead of running a full transcription first and failing at the end with the output already on stdout

moving it earlier meant every option has to be read before the command branches , and two real bugs fell out of that. `has_vad_chunk_option` short circuited on `||` so three `--vad-chunk-*` options were never looked up and got rejected as unknown. seven more (`--json` , `--input-rate` , `--input-channels` , `--batch-merge-audio` , `--batch-audio-role` , `--audio-converter` , `--workflow-input`) are only read on one path , so they were rejected everywhere else. both fixed , and i swept all 98 options the cli looks up , zero false rejects

| command | before | after |
|---|---|---|
| `--task asr ... --audio speech.wav` | exit 0 , 3.1s | exit 0 , 3.1s , identical stdout |
| `--task asr ... --bakend cuda` | ran on cpu , exit 0 | `unknown option: --bakend` , exit 1 , 0.13s , no stdout |
| `--task asr ... --sesion-option a=b` | silent , exit 0 | `unknown option: --sesion-option` , exit 1 |
| `--task asr ... --foo --bar` | silent , exit 0 | `unknown options: --foo --bar` , exit 1 |
| `--task asr ... --metrics --foo` | silent , exit 0 | `unknown option: --foo` , exit 1 |
| `--task asr ... --log-file --weird-looking-value` | exit 0 | exit 0 , the value is not an option |
| `--list-loaders --bakend cuda` | silent , exit 0 | `warning: ignored option: --bakend` , exit 0 |

the one case i could not make strict is the informational commands. `--help` , `--list-*` and `--inspect` return before the rest of the options are read , so there the cli genuinely cannot tell `--bakend` from `--backend` , one is a typo and the other just does not apply to that command. rejecting both would break anything doing `--list-devices --backend cuda` today , so those stay a warning. making them strict too needs the whole option set parsed up front before dispatch , happy to do that if you want it

built on macos arm64 with `-DAUDIOCPP_MODEL_SET=custom -DAUDIOCPP_MODELS=citrinet_asr` , stdout byte for byte identical to main on every valid command i tested including batch , streaming , stdin pcm and the vad chunk outputs

0xShug0:
@avionicharshit-byte Thanks! PR merged.

PR #286, opened 2026-08-20
ci: add tag-driven prebuilt release pipeline

drzsdrtfg:
## Summary
Adds the prebuilt release pipeline: every `v*` release now ships ready-to-run binaries attached to the GitHub Release. This is the full set of changes this fork adds over upstream `0xShug0/audio.cpp` — 2 files.

## Changes
| File | What it does |
|---|---|
| `.github/workflows/release.yml` | Tag-driven release build + publish |
| `docs/RELEASING.md` | Release / dry-run / verification docs |

### `.github/workflows/release.yml`
- Triggers on a `v*` semver tag push, or manual `workflow_dispatch` with a publish checkbox (used for build-only dry runs). No release on every `main` push.
- Builds `audiocpp_cli`, `audiocpp_server`, `audiocpp_gguf` with `-DAUDIOCPP_DEPLOYMENT_BUILD=ON -DENGINE_ENABLE_LLAMAFILE=ON -DENGINE_ENABLE_CUDA_GRAPHS=ON`.
- Matrix (fail-fast off): macOS metal (arm64/x64), Ubuntu x64 cpu+vulkan, Windows x64 cpu+vulkan+cuda (12.4/13.3). `linux-rocm` is commented out (no AMD GPU runner).
- Windows CUDA uses GGML_BACKEND_DL: thin shells + a single `ggml-cuda.dll`; the CUDA runtime ships once in a separate `-cudart-...zip`. CUDA archs are pinned all-real.
- Assets: `audio-<tag>-bin-windows-x64-<backend>.zip` (+ `-cudart-...zip`); Linux/macOS as `.tar.gz`.
- Release job: `gh release upload` to an existing release (`--clobber`) or `gh release create` on a new tag (`--generate-notes`).

### `docs/RELEASING.md`
UI-only steps to draft a release, tag, dry-run from the Actions tab, plus the artifact table and the GPU-runtime verification note.

## Validation
The prebuilt binaries were downloaded and run to confirm they work out of the box.

Hardware:
- GPU: **Radeon RX Vega 56** (Vulkan device: "Radeon RX Vega"; fp16:1, bf16:0, int dot:0, matrix cores:none, warp:64, 32 KB smem), 8 GB VRAM
- CPU: Intel Core i5-10400 (6c/12t), Windows, AMD proprietary driver

Config: `audiocpp_cli --task tts --family supertonic --model supertonic-3-f16.gguf --voice-id M1 --num-inference-steps 8 --seed 1234`

[...]

PR #342, opened 2026-08-29
docs: update prebuilt binaries section

drzsdrtfg:
## What

Updates the README **Prebuilt Binaries** section to match the tag-driven release pipeline from #286.

- Lists the platforms and backends that now ship official packages on every release: Windows x64 (CPU, Vulkan, CUDA 12.4 / 13.3), Ubuntu x64 (CPU, Vulkan), macOS arm64/x64 (Metal).
- Explains what the separate cudart archive is for: extract it next to the binaries so the CUDA backend can load ggml-cuda.dll and the CUDA runtime DLLs (cudart, cuBLAS, cuFFT).
- Keeps the community-maintained Windows HIP/ROCm pointers unchanged (minus a stray em dash).

## Notes

Platform/backend list was verified against the assets attached to the v0.7.0 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,8 +223,17 @@ model notes, and frontend development instructions.
 
 ## Prebuilt Binaries
 
-- **Windows (CUDA / CPU):** official packages on the [Releases page](https://github.com/0xShug0/audio.cpp/releases).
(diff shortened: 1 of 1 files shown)


0xShug0:
@drzsdrtfg Merged. Thanks!

Issue #345, opened 2026-08-29
HIP build fails: undeclared `cudaStreamCreateWithPriority` (missing alias in vendors/hip.h)


huyndao:
# HIP build fails: undeclared `cudaStreamCreateWithPriority` (missing alias in vendors/hip.h)

### Summary

Building audio.cpp with the HIP backend fails to compile the ggml CUDA sources.
`common.cuh` calls `cudaStreamCreateWithPriority`, but that symbol has no
CUDA→HIP alias in `external/ggml/src/ggml-cuda/vendors/hip.h`, so every `.cu`
file that includes `common.cuh` fails with `use of undeclared identifier`
(the compiler itself suggests the correct HIP equivalent).

### Environment

- audio.cpp: 0.2.1
- ggml (vendored): 0.12.0, commit `8ceb983`
- ROCm: 7.2.4 (`/opt/rocm-7.2.4`)
- HIP compiler: Clang 22.0.0 (`/opt/rocm/lib/llvm/bin/clang`)
- GPU target: gfx1150
- OS: Debian GNU/Linux forky/sid (forky) x86_64
- Configure summary: HIP backend enabled, hipBLASLt GEMM enabled

### Steps to reproduce

1. Configure with the HIP backend enabled (`build_hip` build dir)
2. `cmake --build build_hip` (or `make`)

### Error

Every HIP object fails identically, e.g. `concat.cu`, `acc.cu`, `argmax.cu`, ...:

```
In file included from external/ggml/src/ggml-cuda/concat.cu:1:
In file included from external/ggml/src/ggml-cuda/concat.cuh:1:
external/ggml/src/ggml-cuda/common.cuh:1461:28: error: use of undeclared identifier
'cudaStreamCreateWithPriority'; did you mean 'hipStreamCreateWithPriority'?
 1461 |                 CUDA_CHECK(cudaStreamCreateWithPriority(
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/rocm-7.2.4/include/hip/hip_runtime_api.h:2835:12: note:
'hipStreamCreateWithPriority' declared here
1 error generated when compiling for gfx1150.
```

Build aborts with `gmake: *** [Makefile:156: all] Error 2`.

### Cause

`src/ggml-cuda/common.cuh` (around line 1461) creates priority streams via
`cudaStreamCreateWithPriority`, but `vendors/hip.h` — the table of
[...]

PR #346, opened 2026-08-29
fixes issue #345

huyndao:
Fixes issue #345 

Building audio.cpp with the HIP backend fails to compile the ggml CUDA sources.
`common.cuh` calls `cudaStreamCreateWithPriority`, but that symbol has no
CUDA→HIP alias in `external/ggml/src/ggml-cuda/vendors/hip.h`, so every `.cu`
file that includes `common.cuh` fails with `use of undeclared identifier`
(the compiler itself suggests the correct HIP equivalent).

This fix adds the following to `src/ggml-cuda/vendors/hip.h` alongside the other
cudaStream* defines:

```c
#define cudaStreamCreateWithPriority     hipStreamCreateWithPriority
#define cudaStreamGetPriority            hipStreamGetPriority
#define cudaDeviceGetStreamPriorityRange hipDeviceGetStreamPriorityRange
```

### DIFF
--- a/external/ggml/src/ggml-cuda/vendors/hip.h
+++ b/external/ggml/src/ggml-cuda/vendors/hip.h
@@ -121,6 +121,9 @@
 #define cudaStreamPerThread hipStreamPerThread
 #define cudaStreamSynchronize hipStreamSynchronize
 #define cudaStreamWaitEvent hipStreamWaitEvent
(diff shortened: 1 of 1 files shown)


0xShug0:
@huyndao PR merged! Thanks!

PR #358, opened 2026-08-30
audio: add 24-bit and float32 WAV output, opt-in dither, and a limiter

CryptVenture:
## What this changes

`write_pcm16_wav` was the only writer the framework had, so every generated file was capped at 16-bit PCM regardless of sample rate. `wav_reader` already handles 8/16/24/32-bit, float32/64, A-law and mu-law, so the asymmetry was the writer's alone. That is a hard quality ceiling for 44.1/48 kHz music output and for anything that gets processed further downstream.

Adds `WavWriteOptions` to a new `write_wav()`:

| Option | Values |
|---|---|
| `WavSampleFormat` | `Pcm16` (default), `Pcm24`, `Float32` |
| `WavDitherMode` | `None` (default), `TriangularPdf` |
| `WavPeakPolicy` | `HardClip` (default), `LookaheadLimit` |

**Every default reproduces the previous behaviour.** `write_pcm16_wav` is now a thin wrapper whose output is byte-identical, verified in the test against a fixture built from the previous implementation, so the ~70 existing call sites are unaffected.

## Measured

| Path | Result |
|---|---|
| float32 round trip | exact, `0.000e+00` error |
| pcm24 round trip | 130.57 dB SNR, 1.19e-7 max error |
| pcm16 round trip | 82.41 dB SNR (unchanged) |
| TPDF dither at −60 dBFS | harmonics −54.62 → **−67.33 dBc** |
| +1 dBFS overshoot, hard clip | 29.97% of samples railed, −26.69 dB THD+N |
| +1 dBFS overshoot, limiter | **0.00% railed, −98.01 dB THD+N** (1.04 dB of level) |

Dither is opt-in because it is a deliberate choice, not a universal improvement, and must not be applied twice in a chain.

A memoryless soft-clip waveshaper was implemented first and **rejected**: it measured −26.7 dB THD+N, indistinguishable from the clamp it was meant to replace. Overshoot is a gain problem, not a curve problem. The shipped `LookaheadLimit` is a real limiter — per-frame peak, sliding-window minimum over ±5 ms, two cascaded box filters, channel-linked — and leaves any buffer that never crosses the ceiling bit-identical.

Also adds a 4 GiB RIFF size guard, and a `WavSink` alongside `WavPcm16Sink` so a sink can carry a format.

## Validation

**Build**
```
cmake --build build
```

**Test**
```
ctest -R wav_writer_formats_test
```

`tests/unittests/test_wav_writer_formats.cpp` — round trips and per-depth error bounds derived rather than guessed, full RIFF header verification per format (including the `fact` chunk for float32), byte-identity of `write_pcm16_wav`, dither reproducibility for a fixed seed, and the limiter assertions above. No model weights required.

**Backend tested:** CPU (pure host code). Full suite **39/39** on macOS/Metal, Apple M4 Max.

## Affects

Output only, and only for callers that opt in. No model family changes behaviour.

## Known limitations

- `app/server/runtime.cpp` still hard-codes 16-bit for the HTTP response body, and its float→int16 loop duplicates the old writer. It can now be a one-line delegation; left for a separate change.
- The workflow sinks instantiate `WavPcm16Sink` directly, so 24-bit and float32 are reachable from the library API but not yet from a CLI/server flag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2032,6 +2032,13 @@ if (ENGINE_BUILD_TESTS)
         COMMAND audio_chunking_test
     )
 
+    add_engine_unittest(wav_writer_formats_test tests/unittests/test_wav_writer_formats.cpp)
+
+    add_test(
+        NAME wav_writer_formats_test
+        COMMAND wav_writer_formats_test
+    )
+
     add_engine_unittest(chinese_normalization_test tests/unittests/test_chinese_normalization.cpp)
 
     add_test(
--- a/include/engine/framework/audio/output.h
+++ b/include/engine/framework/audio/output.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "engine/framework/audio/wav_writer.h"
+
 #include <filesystem>
 #include <string>
 #include <vector>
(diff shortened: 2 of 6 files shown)


0xShug0:
@CryptVenture Thank you for the PRs. I’ll review them one by one and ping you when I move on to the next one. Please rebase each PR when I start reviewing it.

CryptVenture:
Sure. I have submitted a fair few - a lot of house keeping things really. Hope they help.

CryptVenture:
Hi, Just wondering whether you were going to merge all of the pulls I sent through? I notice some of them are now drafts etc.

0xShug0:
@CryptVenture "Draft" or "Close" are just my queue mangement for PRs. 

I agree that many of these PRs contain useful changes, but they are hard to review because they touch unrelated files and mix multiple concerns. Unit tests alone are not sufficient validation here, because the PR may not exercise the actual runtime path covered by the change. Any model change needs real model-level A/B validation: run before and after using path tests on the affected code paths, compare the behavior, and include the commands/results. Static reasoning is not enough, even if AI argues the change is equivalent. You may be surprised by how often code can break at runtime by "equivalent" changes. And backend or framework behavior changes require regression path tests on affecting models. 

For now, I’ll review only a few clean PRs and defer the others until they are split into smaller, single-responsibility PRs with proper validation for each change.

0xShug0:
@CryptVenture Good addition. PR merged.

PR #359, opened 2026-08-30
audio: find libsoxr on Apple Silicon, and stop the unfiltered decimation

CryptVenture:
> **Depends on #358.** This branch is stacked on it, so the diff shown here includes that commit. Review the second commit only, or merge #358 first. The test writes its probe as float32 — a 16-bit container would put a −96 dBc floor under a −117 dBc measurement.

## What this changes

Two resampling defects that compound, and both cost quality silently.

### 1. libsoxr is never found on Apple Silicon

`SoxrApi` loads libsoxr with `dlopen("libsoxr.dylib")` — a bare leaf name. On Apple Silicon dyld's default search path covers `/usr/local/lib` and `/usr/lib` but **not** `/opt/homebrew/lib`, so a Homebrew libsoxr is never found and `resample_mono_soxr_or_linear` falls back to linear interpolation. The fallback warning goes through the debug logger, which defaults to disabled, so the degradation is invisible.

Adds absolute-path candidates for the common install prefixes plus an `AUDIOCPP_SOXR_LIBRARY` override, and a one-time stderr notice on fallback so it can no longer happen unnoticed.

### 2. The enhancement-model input path has no anti-alias filter at all

`read_mono_resampled` — the input path for **every** denoise and super-resolution entry point — called `resample_mono_linear` unconditionally.

Measured against the real sources, a 12 kHz tone decimated 48 → 16 kHz for a 16 kHz model:

| Path | Folded 4 kHz alias |
|---|---|
| `resample_mono_linear` (previous) | **0.00 dBc** |
| sinc width 6 | −55.06 dBc |
| sinc width 64 | −117.08 dBc |
| soxr | −288.17 dBc |

**0.00 dBc is not a typo.** 48 → 16 kHz is an exact 3:1 ratio, so the interpolation fraction is identically zero and linear interpolation degenerates into plain sample-dropping — the alias arrives unattenuated. FlashSR is then asked to re-synthesise the band that was just destroyed.

## The default that was deliberately *not* changed

`TorchaudioSincHannResampleOptions::lowpass_filter_width` stays at **6**. That is `torchaudio.transforms.Resample`'s own default, and the ~36 call sites taking it are feature-extraction paths whose contract is bit-parity with a Python reference — widening it would change *model inputs* everywhere. Per CONTRIBUTING's framework-module rule, the capability was added as `torchaudio_sinc_hann_playback_options()` (width 64) and only paths producing audio a listener hears were switched over: the audio-utility input path and the mix bus.

Width 64 chosen from the measured knee on a 44.1 → 48 → 44.1 round trip:

| Width | SNR | Time (4 s mono) |
|---|---|---|
| 6 | 60.54 dB | 0.56 ms |
| 64 | **122.96 dB** | 6.51 ms (615× realtime) |
| 128 | 141.42 dB | 15.81 ms |

Also adds `resample_mono_soxr_or_sinc` — soxr when available, else the in-tree windowed sinc, with `output_length_policy` honoured on the fallback (the linear path ignored it).

## Validation

**Build**
```
cmake --build build
```

**Test**
```
ctest -R audio_resample_quality_test
```

`tests/unittests/test_audio_resample_quality.cpp` asserts the folded 4 kHz alias is at or below −60 dBc by coherent single-bin DFT. That threshold fails the old linear path by 60 dB and fails a regression to the width-6 default, while passing both the sinc fallback and soxr — so it does not depend on whether libsoxr is installed on the build machine. It also pins the shared defaults so a future change to them is deliberate.

**Backend tested:** CPU (host DSP). Full suite **40/40** on macOS/Metal, Apple M4 Max.

## Affects

Output changes for `deepfilternet2`, `rnnoise`, `zipenhancer` and `flashsr` whenever the source rate differs from the model rate — that is the fix. Same-rate input is a no-op fast path in both old and new code, so 16 kHz → 16 kHz and 48 kHz → 48 kHz are untouched. Output length may differ by ±1 sample (sinc sizes with `ceil`, the old linear path with `llround`).

Any process that hits a soxr fallback now prints one line to stderr, once per process.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2032,6 +2032,13 @@ if (ENGINE_BUILD_TESTS)
         COMMAND audio_chunking_test
     )
 
+    add_engine_unittest(audio_resample_quality_test tests/unittests/test_audio_resample_quality.cpp)
+
+    add_test(
+        NAME audio_resample_quality_test
+        COMMAND audio_resample_quality_test
+    )
+
     add_engine_unittest(wav_writer_formats_test tests/unittests/test_wav_writer_formats.cpp)
 
     add_test(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2032,6 +2032,13 @@ if (ENGINE_BUILD_TESTS)
         COMMAND audio_chunking_test
     )
 
+    add_engine_unittest(wav_writer_formats_test tests/unittests/test_wav_writer_formats.cpp)
+
+    add_test(
+        NAME wav_writer_formats_test
+        COMMAND wav_writer_formats_test
+    )
+
     add_engine_unittest(chinese_normalization_test tests/unittests/test_chinese_normalization.cpp)
 
     add_test(
--- a/include/engine/framework/audio/conversion.h
+++ b/include/engine/framework/audio/conversion.h
@@ -50,4 +50,26 @@ std::vector<float> read_wav_f32_as_mono_linear_resampled(
     const std::filesystem::path & path,
     int target_sample_rate_hz);
 
+// Anti-aliased equivalents of the three helpers above. The `_linear_` versions
+// call a two-tap interpolator with no decimation filter, so a 48 -> 16 kHz
+// conversion folds a 12 kHz tone back to 4 kHz at -9.5 dBc; these route through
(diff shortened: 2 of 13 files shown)


CryptVenture:

PR #361, opened 2026-08-30
audio/text: crossfade genuine chunk seams, and split space-free scripts

CryptVenture:
## What this changes

Two chunking defects with wide reach.

### 1. Chunk joins were a raw `insert` in ~28 models

`runtime::append_audio_buffer` is the shared join point — 39 call sites. A seam step of 0.1 is a broadband impulse at −20 dBFS. Correct crossfades already exist in DramaBox (50 ms equal-power), Seed-VC (cos²), Confucius4 and Inflect v2, so the framework's own join was the odd one out.

**The naive fix is wrong, and this is the important part.** `append_audio_buffer` is not only a TTS chunk join: `sense_asr:436`, `qwen3_asr:425`, `voxtral_realtime:308` and `nemotron_asr:433` use it to accumulate *contiguous streaming input*, as do several vocoders. A full-scale 1 kHz tone at 24 kHz steps 0.235 per sample — far above any silence threshold — so a step-only gate would have crossfaded continuous audio, attenuating real signal and **shifting every ASR word timestamp**.

The new `Auto` default therefore measures the seam before touching it:

| Condition | Action |
|---|---|
| step ≤ −54 dBFS | raw splice, byte-identical |
| step within 2× the sample-to-sample slew of the 64 samples either side | raw splice |
| otherwise | 10 ms equal-power crossfade |

Measured seam-impulse reduction on a genuine discontinuity: **−46.66 dB**. A split sine rejoins bit-exactly, which the test asserts.

Equal power rather than cos² for `Auto` because the two sides of a text-chunk seam are uncorrelated: equal-power holds RMS to **+0.056 dB** through the overlap where cos² dips **−1.17 dB**. Both remain selectable, and `AudioChunkJoinMode::Concat` reproduces the previous join sample for sample.

### 2. `split_word_ranges` stops chunking entirely for CJK

It keys on ASCII whitespace only, so space-free text degenerates to a single whole-text chunk. A long Chinese or Japanese input is handed to the model entire and then truncated at the token cap with no error. Affects Qwen3-TTS, Fish Audio and Vevo2.

Adds CJK-aware splitting: on CJK punctuation first, then on codepoint count at grapheme-safe boundaries.

Limited deliberately to CJK/kana/Hangul/fullwidth. Thai, Lao, Khmer and Myanmar have the same failure, but widening the predicate would move output for languages nobody here can test.

## Validation

**Build**
```
cmake --build build
```

**Test**
```
ctest -R "audio_chunk_join_test|text_chunking_test"
```

`test_audio_chunk_join.cpp` (17 cases) and `test_text_chunking.cpp` (13 cases).

The ASCII regression guard is **22 exact expectations** covering sentence, clause, no-punctuation, over-long-word, URL, budget-boundary, whitespace and tag-aware cases. It was verified to pass against the pre-change implementation *before* the change landed, so it encodes the old behaviour rather than rationalising the new one. Every CJK chunk is asserted to be standalone-valid UTF-8 and to rejoin to the source codepoint sequence.

**Backend tested:** CPU (host code). Full suite **40/40** on macOS/Metal, Apple M4 Max.

## Affects

Roughly 28 models share the join, so this is the widest-reach change in the set. The slew gate is what keeps `Auto` inert for the four ASR streaming accumulators — that is the case worth a second reviewer's eye, and a long streaming ASR run confirming word timestamps and total sample count are unchanged would be the right acceptance test.

`split_text_chunks_japanese` changes for one case: a boundary landing immediately before a combining mark now backs off one codepoint. Nothing in `src/` or `app/` selects that mode.

## Known limitations

- `Auto`'s thresholds (−54 dBFS, 2× slew) are reasoned from measurements, not perceptually validated. There is no perceptual metric in the repo; if one lands, these two constants are what to re-tune.
- Self-append (`append_audio_buffer(x, x)`) is still UB, exactly as before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2032,6 +2032,20 @@ if (ENGINE_BUILD_TESTS)
         COMMAND audio_chunking_test
     )
 
+    add_engine_unittest(audio_chunk_join_test tests/unittests/test_audio_chunk_join.cpp)
+
+    add_test(
+        NAME audio_chunk_join_test
+        COMMAND audio_chunk_join_test
+    )
+
+    add_engine_unittest(text_chunking_test tests/unittests/test_text_chunking.cpp)
+
+    add_test(
+        NAME text_chunking_test
+        COMMAND text_chunking_test
+    )
+
     add_engine_unittest(chinese_normalization_test tests/unittests/test_chinese_normalization.cpp)
 
     add_test(
--- a/include/engine/framework/audio/chunking.h
+++ b/include/engine/framework/audio/chunking.h
@@ -61,6 +61,49 @@ struct QuietEnergyAudioChunkOptions {
     int64_t min_energy_window_samples = 0;
 };
 
+// How two consecutive output chunks are joined into one waveform.
+//
+// Concat      — raw splice, the historical behaviour. Bit-exact, but a step at the
+//               seam is a broadband impulse at 20*log10(step) dBFS.
+// EqualPower  — sin/cos pair. cos^2 + sin^2 == 1, so it preserves *power* through
+//               the overlap. Correct for the uncorrelated content either side of a
+//               text-chunk seam; +3.01 dB at the midpoint if the two sides happen
+//               to be identical.
+// CosSquared  — cos^2/sin^2 pair. The gains sum to exactly 1, so it preserves
+//               *amplitude* for correlated content and dips ~1.25 dB for
+//               uncorrelated content.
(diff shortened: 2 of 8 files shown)


0xShug0:
I won't change the `append_audio_buffer` behavior unless you can provide measurements showing that this PR actually improves accuracy. There are benchmarks using audio.cpp in production pipelines showing that the current chunker works as expected.

Crossfades are used in other models because those models implement them explicitly in the Python reference code, while `append_audio_buffer` is generalized and shared across several other models.

PR #362, opened 2026-08-30
transformers: stop copying the whole KV cache on every decode step

CryptVenture:
## What this changes

`QwenDecoderAttentionPolicy::static_mode` defaulted to `FlashGrouped`, which wraps the K/V cache views in `ggml_cont` before the flash call. The `FlashGroupedViewKV` branch **ten lines above** makes the identical call with zero copies — roughly **100 MB of dead copy traffic per token** at 24 layers and 2048 context.

## Why the default moves rather than the fast path being opt-in

CONTRIBUTING's framework-module rule says not to change shared behaviour models depend on. The justification here is that the two branches are **bit-identical**, not merely equivalent in principle:

- Both call `ggml_flash_attn_ext(q, k, v, mask, 1/sqrt(dim), 0, 0)` with the same `ggml_flash_attn_ext_set_prec`, the same mask, scale, softcap and GQA head grouping. The only difference is whether K/V are copied to a packed layout first.
- Both kernels address K/V purely through **strides**. The CPU path asserts only `nb?0 == type_size` (rows contiguous), never full contiguity, and reads `k->data + ic*nbk1 + ik2*nbk2`; the Metal FA kernel reads `k + ic*args.nb11` and `v + ic*args.nb21`.
- **Kernel selection never inspects strides**: CPU `use_split_kv_path`/`use_tiled` depend only on `ne` and types, Metal's vec path only on `ne00`/`ne01`. Both branches therefore take the identical code path with identical values in identical reduction order.
- `device_supports_op` for `GGML_OP_FLASH_ATTN_EXT` tests head size, K/V type and device caps, **never strides**, so the flip cannot make a graph invalid. That matters because these graphs run on a single backend with no scheduler fallback.

`test_qwen_attention_modes` builds a real `QwenDecoderLayerModule` on the CPU backend and asserts **`max_abs == 0`** between the two modes. On failure the exception names the one line to revert. `ENGINE_QWEN_ATTENTION_CONT_KV=1` restores the copying path at runtime without a rebuild.

**Blast radius is 5 models, not more:** `vietneu_tts`, `fun_asr_nano`, `higgs_audio_stt`, `qwen3_asr`, `index_tts2`. Fourteen configs already ship on `FlashGroupedViewKV`, and the two Moss models keep `FlashGrouped` explicitly.

## What was deliberately *not* changed

`prefill_mode` stays on `ManualRepeat`. Flash prefill is not equivalent — it fuses an online softmax where `ManualRepeat` materialises `[heads,T,T]` and calls `soft_max_ext` — and more importantly it imposes a **hard backend requirement**: Metal supports `FLASH_ATTN_EXT` only for a whitelist of head sizes and requires `ne00 % 4 == 0`, `head_dim` comes from a runtime JSON config, and with no scheduler fallback an unsupported size is a graph *failure*, not a slowdown. `minimax_music3/depth_decoder.cpp:61-63` already gates flash prefill to CUDA/HIP for exactly this reason.

## Also in this change

`TransformerKVCache::import_state` now uploads only the valid prefix and clears only the stale tail, tracked by a high-water mark, instead of zero-filling and uploading the whole capacity on every `start_decode_*`. `export_state` reads back only the valid prefix. Per-layer staging buffers are allocated lazily.

`test_graph_optimizer_identity_fold` records as a **negative control** that `optimize_graph` cannot fix the above: `is_identity_materialization` requires `same_type_shape_and_layout`, which compares `nb[]` exactly, and `ggml_cont(permute(cache))` has genuinely different strides. It only elides `ggml_cont` calls that were already no-ops.

## Validation

**Build**
```
cmake --build build
```

**Test**
```
ctest -R "qwen_attention_modes_test|graph_optimizer_identity_fold_test"
```

**Backend tested:** Metal, Apple M4 Max. A FireRedTTS3 multi-chunk clone and a VibeVoice render are **byte-identical** before and after (SHA-256 `60441bc8…` and `d38f9ebd…`). The FireRedTTS3 case also exercises the KV prefix import across repeated `start_decode_*` calls in one process, which is where a mistake in that part would show — on the *second* generation, not the first. Full suite **40/40**.

## Affects

Performance and memory traffic on the five models above; no output change, asserted bit-exact. **No throughput number is claimed** — the renders used for validation are dominated by model load and are byte-identical, which proves correctness rather than speed. Measuring the gain properly needs a long single-utterance decode with a token-rate counter.

## Known limitations

`TransformerBatchedKVCache::import_state` has the same full-capacity pathology and is left alone here; it serves far fewer models and deserves its own change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2032,6 +2032,20 @@ if (ENGINE_BUILD_TESTS)
         COMMAND audio_chunking_test
     )
 
+    add_engine_unittest(qwen_attention_modes_test tests/unittests/test_qwen_attention_modes.cpp)
+
+    add_test(
+        NAME qwen_attention_modes_test
+        COMMAND qwen_attention_modes_test
+    )
+
+    add_engine_unittest(graph_optimizer_identity_fold_test tests/unittests/test_graph_optimizer_identity_fold.cpp)
+
+    add_test(
+        NAME graph_optimizer_identity_fold_test
+        COMMAND graph_optimizer_identity_fold_test
+    )
+
     add_engine_unittest(chinese_normalization_test tests/unittests/test_chinese_normalization.cpp)
 
     add_test(
--- a/include/engine/framework/modules/transformers/qwen_decoder.h
+++ b/include/engine/framework/modules/transformers/qwen_decoder.h
@@ -12,6 +12,21 @@ struct ggml_cgraph;
 
 namespace engine::modules {
 
+// Attention lowerings for the Qwen decoder block.
+//
+//   ManualRepeat                 explicit QK^T -> scale/mask -> softmax -> V, with the
+//                                grouped-query K/V heads materialised at full query width.
+//                                Quadratic working set in sequence length; the numerical
+//                                reference path, and the only one that does not require the
+//                                backend to implement GGML_OP_FLASH_ATTN_EXT for this head size.
+//   FlashGrouped                 ggml_flash_attn_ext, with the permuted K/V cache views
+//                                copied into packed tensors (ggml_cont) beforehand.
+//   FlashGroupedViewKV           the same ggml_flash_attn_ext call on the K/V views directly.
+//                                The flash kernels address K/V through nb[1..3] on both the CPU
+//                                and Metal backends, so this is the same arithmetic in the same
+//                                order on the same values as FlashGrouped -- only the copy is
+//                                gone. See tests/unittests/test_qwen_attention_modes.cpp.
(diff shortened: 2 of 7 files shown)


0xShug0:
The reasoning is plausible, but source inspection is not enough to prove that across backend kernels. I'd prefer to do per model adjustment (i.e., change to flash view) to make the PR easier to test on cuda/vulakn/metal/hip. Then if  FlashGrouped become useless we can remove this.  Please provide full regression tests on the affected models across backends. I’d be happy to reopen this PR once those tests are available.

PR #364, opened 2026-08-30
tts: make the OmniVoice reference clamp reachable, and stop silent truncation

CryptVenture:
### OmniVoice's reference-length clamp was dead code

It sat behind `!options.has_reference_text` (`audio_tokenizer.cpp:381`), but `prompt_builder.cpp:772-775` **throws** if a voice-clone request arrives without reference text. The two conditions are mutually exclusive, so the clamp could never fire on a real clone. References were encoded whole at 75 Hz (24000 sample rate / 320 hop) with no guard and no warning.

The shipped demo voices run 4.7–9.9 s, i.e. 353–741 frames. A 101 s reference is **7,579 frames**, and measured on Metal it generated at 0.30× realtime and came out garbled, where a 12 s cut of the same voice ran at 2.26× and was clean.

The clamp now applies. It cuts at the **start of a silence run** rather than the quietest single window, tiered at 250/150/80 ms, so whole words survive; if no pause exists inside the limit it keeps the whole clip and warns rather than slicing a word; and it strips the edges and re-pads both ends with exact digital silence, which the model needs. `reference_max_seconds` (default 15) and `reference_pad_ms` (default 150) are request options and key the reference-prompt cache.

A reference with ≤15 s of speech content is returned **byte for byte unchanged**, so the demo voices are bit-identical — a `demo_1_man` render hashes the same before and after (SHA-256 `33b2d439…`).

### The clamp alone was not enough

Only an end-to-end run caught this. `estimate_target_tokens` derives the output duration from the ref_text-to-ref_frames pair as a **speaking-rate proxy**. Clamping the audio while leaving the caller's full transcript collapses that rate: the first fixed run produced **1.75 s** of audio for a sentence that should run about 6 s. `OmniVoiceAudioTokens` now carries `retained_fraction` and `prompt_builder` trims the transcript to the matching head, since the clamp keeps the head of the clip.

Measured on Metal, seed 42, 101 s reference, target text *"At three forty five in the afternoon I walked past twelve Saint Peter Street and bought a coffee."*, transcribed back with Parakeet TDT:

| | Wall clock | ASR of the output |
|---|---|---|
| before | 43.15 s | `Falf did sef sound right. Tapas 12 sex sex thoughts and session. Sayin' bit` |
| after | **6.90 s** | `At 3.45. In the afternoon, I walked past 12 St. Peter Street and bought a coffee.` |

The residual differences are Parakeet's own number and abbreviation normalisation.

### Qwen3-TTS could not synthesise its own default chunk

8192 codepoints against a 2048-token cap is 163 s at the 12.5 Hz codec frame rate, and the talker exited on a bare `break`, so long-form input truncated silently mid-sentence. The default is now derived from the frame rate and the actual cap, and resolved per request so `--max-tokens` resizes it too.

### Cap hits are no longer silent

In Qwen3-TTS and Chatterbox, following the stop-reason pattern OuteTTS already uses. Chatterbox also gained a bounds clamp on `speech_position_weight`: the decode loop indexes it at `step + 1` with no check, so raising the token cap could have overrun a short checkpoint table.

### Chatterbox defaults resolved from the repo's own harnesses

Not by guess. `tests/chatterbox/chatterbox_python_warm_bench.py:34-37` passes `repetition_penalty=1.2` and `top_p=1.0` — so `docs/tts.md` was wrong on both. `tests/chatterbox/chatterbox_warm_bench.cpp:174` defaults `--max-new-tokens` to 1000 — so the code was wrong there.

`top_p` stays 1.0: it is upstream's own default and a deliberate no-op, with `min_p=0.05` doing the filtering. Raising `max_new_tokens` 384 → 1000 is not a sampling change — with `stop_on_eos` any chunk that already finished under 384 tokens is bit-identical.

### VibeVoice reference cap documented and overridable

10 s on Metal, 30 s on CUDA (`session.cpp:25-26`), undocumented and with no override. Both constants arrived in the initial release with no comment and the gate is `Cuda || Hip`, so CPU and Vulkan take 10 s too — this reads as "CUDA was tested at 30 s", not "Metal has a limit". **Default unchanged**, but `vibevoice.voice_prompt_max_seconds` now overrides it, the truncation fades instead of cutting, and `docs/tts.md` states the split and that it is untested.

## Validation

**Build**
```
cmake --build build
```

**Test**
```
ctest -R "omnivoice_reference_clamp_test|qwen3_tts_chunk_budget_test"
```

The OmniVoice test is gated on `omnivoice IN_LIST AUDIOCPP_LINKED_MODELS`. It asserts the cut lands at the start of a pause (source RMS over the 250 ms *after* the cut is 0, over the 100 ms *before* it is > 0.1), that head and tail pad RMS are **exactly 0.0**, that a pause-free clip is kept whole, and that a short reference is byte-for-byte identical.

**Run command used for the Metal A/B**
```
audiocpp_cli --task tts --family omnivoice --model models/OmniVoice-GGUF/omnivoice-q8_0.gguf \
  --backend metal --seed 42 --voice-ref long.wav --reference-text "..." --text "..." --out out.wav
```

**Backend tested:** Metal, Apple M4 Max. Full suite **40/40**.

## Affects

Output changes for OmniVoice references longer than the limit — that is the fix; shorter references are bit-identical. Chatterbox chunks that were being truncated now finish. Qwen3-TTS long-form produces roughly 5× more chunks.

## Known limitations

- Qwen3-TTS has no cross-chunk state, so more chunks means more joins. This trades a silent truncation for more seams, and pairs naturally with a chunk-seam crossfade change.
- Whether Metal can actually take a 30 s VibeVoice prompt is still unmeasured; the new override makes that a one-flag experiment.
- OmniVoice has no cap and no EOS — its length is fixed before generation — so it has no cap-hit stop reason to report.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2032,6 +2032,22 @@ if (ENGINE_BUILD_TESTS)
         COMMAND audio_chunking_test
     )
 
+    add_engine_unittest(qwen3_tts_chunk_budget_test tests/unittests/test_qwen3_tts_chunk_budget.cpp)
+
+    add_test(
+        NAME qwen3_tts_chunk_budget_test
+        COMMAND qwen3_tts_chunk_budget_test
+    )
+
+    if (omnivoice IN_LIST AUDIOCPP_LINKED_MODELS)
+        add_engine_unittest(omnivoice_reference_clamp_test tests/unittests/test_omnivoice_reference_clamp.cpp)
+
+        add_test(
+            NAME omnivoice_reference_clamp_test
+            COMMAND omnivoice_reference_clamp_test
+        )
+    endif()
+
     add_engine_unittest(chinese_normalization_test tests/unittests/test_chinese_normalization.cpp)
 
     add_test(
--- a/docs/tts.md
+++ b/docs/tts.md
@@ -95,10 +95,14 @@ audiocpp_cli --task vc --family chatterbox --model models/chatterbox --backend c
 | `--text-chunk-size` | integer chars | `128` | Long-form chunk size. |
 | `--guidance-scale` | float | `0.5` | CFG strength. |
 | `--temperature` | float | `0.8` | T3 sampling temperature. |
-| `--top-p` | float | `0.8` | T3 nucleus sampling limit. |
-| `--repetition-penalty` | float | `2.0` | T3 repetition penalty. |
-| `--max-tokens` | integer | `1000` | Maximum generated T3 tokens per chunk. |
+| `--top-p` | float | `1.0` | T3 nucleus sampling limit. `1.0` disables nucleus filtering outright; `min_p` is the truncation filter that runs by default. |
+| `--request-option min_p=<float>` | float | `0.05` | T3 minimum-probability filter, applied before top-p. |
+| `--repetition-penalty` | float | `1.2` | T3 repetition penalty. |
+| `--max-tokens` | integer | `1000` | Maximum generated T3 tokens per chunk. At the 25 Hz speech-token rate this is about 40 s of audio; a chunk that reaches it is truncated mid-utterance and a warning is printed. |
 | `--do-sample` | `true`, `false` | `true` | Enable stochastic T3 sampling. |
+| `--request-option exaggeration=<float>` | float | `0.5` | Emotion-advance conditioning strength. |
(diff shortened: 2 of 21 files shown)


0xShug0:
These are separate model behavior changes and need separate PRs per model plus real model runs.

Also it's safe to ignore all python scripts. Those warmbench scripts are just for reference, as they may require customized setups. Many of them cannot run out of the box because they are missing required dependencies.

PR #365, opened 2026-08-30
gguf: stop quantizing the tensors that are audible

CryptVenture:
## What this changes

The packer's entire quantize/keep policy was **one shape predicate** at `tensor_source.cpp:1988-1997`: quantize any tensor that is rank-2, ends in `.weight`, and does not contain `embed` or `codebook`. Vocoders, decoders and output heads are protected by nothing. Conv-stack vocoders survive at F32 purely by being rank-3 or higher, and normalisation weights survive purely by being rank-1 — by accident of shape, not by design.

Verified by parsing the headers of 22 published GGUFs:

- `redae/decoder.istft_head.out.weight` is **Q8_0 [896,1922]** in FireRedTTS3-Instruct and FireRedAudio. That layer's output **is** the magnitude/phase spectrum, exponentiated immediately after, with nothing downstream to absorb the error.
- VibeVoice-7B ships a half-quantized decoder: convs BF16, FFN matmuls Q8_0, 158 tensors.
- IndexTTS2.5's s2mel flow decoder is linear rather than conv, so **139 tensors** land at Q8_0 where OmniVoice's conv vocoder is F32 for free.
- 27 rank-2 norm projections in IndexTTS2.5 are Q8_0 because the norm protection is only a rank test.

Adds a **declarative rule table with a reason string per entry**, covering iSTFT and spectrogram heads, vocoders, codec and latent-to-audio decoders, normalisation projections, and output heads and final projections. A name match demotes the tensor to F16, reusing the existing lookup-table branch rather than adding a mechanism. The table was validated against **13,115 real tensor names** from the installed packages, with no false positives.

Deliberately *not* covered: bare `decoder`, which would move `redae/decoder.qwen2.layers.*` — a 24-layer transformer — for nothing, and every encoder or analysis-side name. `GgufConversionOptions::quantize_audible_tensors` restores the old policy.

**Measured cost of re-cutting:**

| Package | Growth | Tensors |
|---|---|---|
| omnivoice-q8_0 | +8 MiB (+0.6%) | 1 |
| fireredtts3-instruct-q8_0 | +98 MiB (+2.5%) | 15 |
| index-tts2_5-q8_0 | +138 MiB (+4.2%) | 106 |
| vibevoice-7b-q8_0 | +918 MiB (+9.1%) | 59 |

The last row is almost entirely one `lm_head.weight [152064,3584]`.

## Package defaults off 4-bit

MiniMax Music 3 defaulted to legacy `q4_0` for both its language model and its flow transformer — including the transformer's `proj_out`, which feeds the vocoder — while `q8_0` and `bf16` packages sat unused. `q4_0` is a legacy format with one scale per 32-element block and no zero point; `q4_k` is already supported here and is strictly better at the same bitrate. PersonaPlex moves `q4_k` → `q8_0`.

**The Music3 default needs the session fallback too.** `session.cpp` hard-coded the q4_0 filenames as its `value_or` fallbacks, and spec option defaults are metadata that never reaches the runtime options map — so changing the spec alone would make a fresh install of the new default package throw `file does not exist`. The fallback now probes for the highest-quality tier present, so a `q8_0` install and an existing `q4_0` install both resolve without the caller naming a file.

## Scope of the benefit

**This does not change GGUFs that already exist.** Installed models keep the precision they were built with; a package has to be re-converted from its original weights for the policy to reach audio. `docs/gguf.md` says so explicitly, documents the rule table and the override, and is now honest that the lower-bit tiers are untested.

## Validation

**Build**
```
cmake --build build
```

**Test**
```
ctest -R quantization_policy_test
```

`tests/unittests/test_quantization_policy.cpp` writes a 17-tensor safetensors fixture, packs it four ways with the real `convert_tensor_sources_to_gguf`, and asserts the resulting dtype **and exact byte count** of every tensor. No model weights required.

The **regression guard** matters most: FFN, attention, the decoder-side transformer stack and the tokenizer *encoder* must still come out `q8_0`, `embed`/`codebook` must still be `f16`, and the rank-3 `vocoder.conv_pre.weight` must stay F32 rather than being pulled down to F16 — so a widened pattern fails loudly.

**Backend tested:** CPU (pure conversion logic). Full suite **39/39** on macOS/Metal.

## Affects

The converter only. No runtime behaviour changes and no published package stops matching anything it currently matches — the policy only ever moves a tensor to *higher* precision, and the loader already dequantizes on type mismatch.

Rebased onto current `main`, which includes #354; the Music3 spec merge was checked semantically — `default_package` and `ui.recommended_package` both land on `minimax_music3_q8_0` and the three package ids are intact.

## Known limitations

- `minimax_h3` stays at `q4_k` — no higher-precision package is published, and cutting one is a conversion job rather than a config change.
- `embed`/`codebook` remain literal substring tests, so a codebook named `vq.layers.0.weight` is still quantized silently.
- There is no perceptual metric anywhere in the repo, so this change is argued from where the error lands in the signal chain rather than from a measured quality delta. Adding one would be the right follow-up.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2032,6 +2032,13 @@ if (ENGINE_BUILD_TESTS)
         COMMAND audio_chunking_test
     )
 
+    add_engine_unittest(quantization_policy_test tests/unittests/test_quantization_policy.cpp)
+
+    add_test(
+        NAME quantization_policy_test
+        COMMAND quantization_policy_test
+    )
+
     add_engine_unittest(chinese_normalization_test tests/unittests/test_chinese_normalization.cpp)
 
     add_test(
--- a/docs/gguf.md
+++ b/docs/gguf.md
@@ -116,6 +116,41 @@ Additional lower-bit checks:
 | `personaplex` | `q4_k` | Pass |
 | `voxtral_realtime` | `q4_k` | Pass (quick CUDA check; transcripts match Q8 except one capitalization-only difference) |
 
+**What the matrix does not cover.** The three tables above test `orig`, 16-bit and
+`q8_0`, plus the three `q4_k` spot checks listed here. That is the whole of the recorded
+validation. In particular:
+
+- `minimax_music3`, `minimax_h3` and `audiosr` appear in **no row of any table**. Nothing
+  about those families has been recorded through this process.
+- There is no 4-bit column in the main table, so no family's 4-bit package is validated
+  against its own 16-bit or `orig` output.
+- Every "Pass" label is a load-and-run judgement plus, for ASR, a transcript comparison.
+  There is no perceptual metric anywhere in the repo — no MOS, PESQ, SI-SDR or
+  mel-cepstral distortion — so "Pass (drift)" is not quantified and cannot be compared
+  between packages. Timbre, prosody and speaker-identity drift, which is what
+  quantization actually costs a TTS or music model, is not measured at all.
+
+`minimax_music3` and `personaplex` shipped a 4-bit package as their default while a
+`q8_0` package sat published and unused. Both now default to `q8_0`; the 4-bit packages
+remain installable by id. `minimax_h3` still defaults to `q4_k` because no
+higher-precision package is published for it.
+
+| Family | Old default | New default | Package size |
+|---|---|---|---|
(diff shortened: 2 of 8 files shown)


CryptVenture:

PR #366, opened 2026-08-30
backends: stop stranding non-CUDA GPUs on CPU paths

CryptVenture:
> **Depends on #359** (which itself depends on #358). This branch is stacked, so the diff shown here includes those commits — review the last commit only, or merge #359 first. F5-TTS's reference resampling calls `resample_mono_soxr_or_sinc`, which #359 adds.

## What this changes

Four capabilities are gated on CUDA specifically rather than on "a GPU backend". Each was checked against `git log`/`git blame` and against Metal op coverage to decide whether the gate is a **real limitation** or **untested caution**. The answers differ, so the fixes differ.

| Gate | Verdict | Action |
|---|---|---|
| F5-TTS `use_cuda` | Stale caution | Widened |
| RedAE / MiDashengLM iSTFT | Plain oversight, penalising **CUDA** | Fixed |
| `ENGINE_HAS_CUDA_ISTFT` | Real — no implementation exists | Left alone |
| `MM3_DEPTH_GPU_*` | Real — genuinely CUDA-only | Left alone |

### F5-TTS — stale caution, widened

The model arrived in one commit (`b50742f`) whose message mentions CUDA more than thirty times and Metal never, and `docs/community_models/f5_tts.md` contains no occurrence of "metal", "cpu" or "backend". Every use of `is_cuda` in `runtime.cpp` and `synthesize.cpp` was really a **host-vs-device** distinction: `ggml_init` `no_alloc`, `tensor_set/get` versus `memcpy`, `compute_backend_graph` versus `ggml_graph_compute_with_ctx`, gallocr. Nothing CUDA-specific.

Every op both the DiT and the Vocos vocoder graphs emit — SQRT, DIV, MUL, ADD, REPEAT, MEAN, SUM_ROWS, NORM, SOFT_MAX, ROPE, IM2COL, PAD, ARANGE, FLASH_ATTN_EXT, GLU, UNARY GELU/GELU_ERF/SOFTPLUS — was checked against `ggml_metal_device_supports_op` and is present, and both graphs already call `validate_backend_graph_supported` before allocation, so an unsupported op fails **by name** rather than silently.

`F5ComputeDevice` now carries a backend type; `use_cuda` is kept as a legacy alias because the `tests/f5_tts` harnesses set it. `F5_TTS_HOST_VOCODER=1` restores the old routing.

### RedAE and MiDashengLM-Gen — an oversight that penalised CUDA

`redae_codec_runtime.cpp:631` and `midashenglm_gen/audio_tokenizer.cpp:309` built the host iSTFT **unconditionally**, without checking the backend at all, so they ran the scalar tail even where the CUDA kernel existed. `fsq_audio_codec_runtime.cpp:659` already checked. Both now ask the same selector. **No change on Metal** — both still select Host, which is what they already did.

### `ENGINE_HAS_CUDA_ISTFT` — real, left alone

`istft_graph.cpp` holds exactly two implementations: a scalar host tail and one wrapping the cuFFT runtime. There is **no ggml-op path** for another backend to pick up, and ggml has no inverse-FFT op at all, so widening this gate would only make `CudaLogMagnitudePhaseISTFT`'s constructor throw.

A Metal iSTFT needs its own kernel set — spectrum kernel, batched complex→real inverse FFT (Metal Performance Shaders has no rFFT, so realistically a Stockham radix-2/4 kernel with a Bluestein fallback), gather-form overlap-add, and a divide-by-envelope kernel — plus its own numerical parity harness. Out of scope here; the selector and the gather below are the groundwork.

### `MM3_DEPTH_GPU_SAMPLE` / `MM3_DEPTH_GPU_FRAME` — real, left alone

They gate kernels that take raw device pointers and reproduce torch's Philox RNG bit-exactly using CUDA TensorIterator grid geometry probed from the driver. Reaching them from Metal means writing a numerics-critical sampler, not changing a gate.

## iSTFT scatter → gather

The host overlap-add was rewritten from a scatter into a gather so it can be parallelised, and the fold and normalise loops carry OpenMP pragmas. The gather visits each output sample's frames in **increasing frame order** — the same order the scatter reached them — so float accumulation order is unchanged, and the test asserts a maximum difference of **exactly 0.0** against an independently built scatter reference rather than a tolerance. The envelope-underflow check moved into workspace setup so the normalise loop neither branches nor throws.

F5-TTS also had its own unguarded two-tap interpolator for reference resampling; it now calls the framework resampler and inherits its anti-aliasing. `F5_TTS_LEGACY_RESAMPLE=1` restores the old behaviour.

## Validation

**Build**
```
cmake --build build
```

**Test**
```
ctest -R backend_path_selection_test
```

`tests/unittests/test_backend_path_selection.cpp` — selection logic for Metal/Vulkan/CPU/HIP/CUDA in every build configuration, `static_assert`s on the F5 predicates, and the numeric iSTFT gather-versus-scatter comparison over 500 frames at tolerance 0.0.

**Backend tested:** CPU for the selection logic and the iSTFT numerics. Full suite **41/41** on macOS/Metal, Apple M4 Max, with the bit-exactness assertion holding under OpenMP.

## Affects

F5-TTS reference resampling changes the conditioning mel, hence cloned timbre, whenever the reference is not already 24 kHz (`F5_TTS_LEGACY_RESAMPLE=1` to compare). F5-TTS vocoder moves to GPU (`F5_TTS_HOST_VOCODER=1` to compare). RedAE and MiDashengLM-Gen change **on CUDA only**.

## Known limitations

**F5-TTS has never been executed on Metal in this codebase** — no F5 package was installed on the machine this was developed on — so the widened gate is argued from op coverage rather than measured. `validate_backend_graph_supported` will name any missing op rather than failing silently, but that is a claim, not a measurement, and it is the one thing in this PR that would most benefit from a reviewer with an F5 package to hand.

**No speedup is claimed.** No inference was run for this change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2032,6 +2032,13 @@ if (ENGINE_BUILD_TESTS)
         COMMAND audio_chunking_test
     )
 
+    add_engine_unittest(backend_path_selection_test tests/unittests/test_backend_path_selection.cpp)
+
+    add_test(
+        NAME backend_path_selection_test
+        COMMAND backend_path_selection_test
+    )
+
     add_engine_unittest(audio_resample_quality_test tests/unittests/test_audio_resample_quality.cpp)
 
     add_test(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2032,6 +2032,13 @@ if (ENGINE_BUILD_TESTS)
         COMMAND audio_chunking_test
     )
 
+    add_engine_unittest(audio_resample_quality_test tests/unittests/test_audio_resample_quality.cpp)
+
+    add_test(
+        NAME audio_resample_quality_test
+        COMMAND audio_resample_quality_test
+    )
+
     add_engine_unittest(wav_writer_formats_test tests/unittests/test_wav_writer_formats.cpp)
 
     add_test(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2032,6 +2032,13 @@ if (ENGINE_BUILD_TESTS)
         COMMAND audio_chunking_test
     )
 
+    add_engine_unittest(wav_writer_formats_test tests/unittests/test_wav_writer_formats.cpp)
+
+    add_test(
+        NAME wav_writer_formats_test
+        COMMAND wav_writer_formats_test
+    )
+
     add_engine_unittest(chinese_normalization_test tests/unittests/test_chinese_normalization.cpp)
 
     add_test(
--- a/include/engine/community_models/f5_tts/runtime.h
+++ b/include/engine/community_models/f5_tts/runtime.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "engine/framework/assets/resource_bundle.h"
+#include "engine/framework/core/module.h"
 #include "engine/framework/runtime/session.h"
 
 #include <cstdint>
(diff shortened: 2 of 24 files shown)


0xShug0:
Please isolate the F5-TTS changes into a separate PR and provide A/B runtime tests.

Host vs graph iSTFT decision should be data-driven. At least based on my measurement the graph iSTFT provides little performance gain for the FireRed and MiDashengLM-Gen while using extra VRAM resource.

PR #367, opened 2026-08-30
tools: add audiocpp_enhance, a CLI over the audio utility helpers

CryptVenture:
> **Depends on #359** (which itself depends on #358). This branch is stacked, so the diff shown here includes those commits — review the last commit only, or merge #359 first. The tool calls `resample_mono_soxr_or_sinc`, which #359 adds.

## What this changes

The denoisers, FlashSR super-resolution and the resampling helpers in `engine::audio` are library-only, so using them means writing a program against the framework. This adds a small CLI over them, which is useful for preparing reference audio for voice cloning and for post-processing generated output.

```bash
audiocpp_enhance --backend metal --denoise zipenhancer --in a.wav --out b.wav
audiocpp_enhance --backend metal --flashsr --in narrow.wav --out wide.wav
audiocpp_enhance --resample 48000 --in mix_44k.wav --out mix_48k.wav
```

Denoise models are `zipenhancer`, `deepfilternet2` and `rnnoise`; `--backend` takes `cpu` (default) or `metal`.

Resampling handles **any channel count**, resampling each channel independently and re-interleaving, so it works on the stereo output of the music models rather than refusing anything but mono. It routes through `resample_mono_soxr_or_sinc`, so it uses libsoxr when that is loadable and the in-tree windowed sinc otherwise, and never silently degrades to linear interpolation.

## Not built by default

`AUDIOCPP_BUILD_ENHANCE_TOOL` defaults **OFF**, matching `ENGINE_BUILD_EXAMPLES`, `ENGINE_BUILD_TESTS` and `ENGINE_BUILD_WARMBENCH`. No existing build changes.

## Validation

**Build**
```
cmake -S . -B build -DAUDIOCPP_BUILD_ENHANCE_TOOL=ON
cmake --build build --target audiocpp_enhance
```

**Run**
```
build/bin/audiocpp_enhance --resample 48000 --in stereo_44k.wav --out stereo_48k.wav
```

**Stereo correctness**, verified on a 44.1 → 48 kHz file carrying 997 Hz in the left channel and 1997 Hz in the right:

| | 997 Hz | 1997 Hz |
|---|---|---|
| Left | **−6.02 dBFS** | −74.73 dBFS |
| Right | −73.79 dBFS | **−6.02 dBFS** |

Each tone comes back at full level in its own channel with ~68 dB separation, i.e. the channels stay separate through the conversion.

**Backend tested:** CPU and Metal on macOS, Apple M4 Max.

## Affects

New optional target only. Nothing existing changes.

`docs/audio_tools.md` gains a section covering the build flag, the three modes and the resampling behaviour.

## Known limitations

`--in` and `--out` must be different paths; passing the same path fails in `copy_file` rather than with a clear message.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1935,6 +1935,17 @@ if (ENGINE_BUILD_WARMBENCH)
     add_engine_warmbench(voxcpm2_warm_bench tests/voxcpm2/voxcpm2_warm_bench.cpp)
 endif()
 
+option(AUDIOCPP_BUILD_ENHANCE_TOOL "Build the audiocpp_enhance audio-utility CLI" OFF)
+if (AUDIOCPP_BUILD_ENHANCE_TOOL)
+    add_executable(audiocpp_enhance
+        tools/audiocpp_enhance/main.cpp
+    )
+    target_link_libraries(audiocpp_enhance PRIVATE engine_runtime ggml)
+    if (ENGINE_ENABLE_OPENMP)
+        target_link_libraries(audiocpp_enhance PRIVATE OpenMP::OpenMP_CXX)
+    endif()
+endif()
+
 if (ENGINE_BUILD_TESTS)
     enable_testing()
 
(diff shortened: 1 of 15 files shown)


CryptVenture:

PR #369, opened 2026-08-31
server: stop rejecting requests over options the client always sends

CryptVenture:
## Summary

Four server fixes in the request path. The first one is a hard blocker: a model whose spec does not declare `language` returned HTTP 500 for every transcription sent from the WebUI.

## The blocker

`build_openai_transcription_request` wrote `request.options["language"]` whenever the request body carried the key. Clients send that field unconditionally, empty or not, and spec-backed models validate their request options strictly, so the model rejected its own client:

```
POST /v1/audio/transcriptions
{"model":"parakeet-tdt","audio":"...wav","language":"","text":"","options":{}}
-> {"error":{"message":"unknown Parakeet TDT request option: language"}}
```

Parakeet-TDT could not transcribe at all from the WebUI. AudioSR, ControlFoley, MiDashengLM-Gen, PersonaPlex and HeartMuLa fail the same way through `/v1/tasks/run`, which folds a top-level `language` into the option map.

The model's acceptance of `language` is now resolved once at registration, next to the existing `accepts_reference_text` flag and for the same reason — resolving it per request re-reads the embedded spec on the request thread — and the field is forwarded only where the contract takes it. A `language` the caller placed inside `options` explicitly is left alone and still produces the strict rejection, which is the correct signal there. The language keeps reaching every model through `text_input`; `qwen3_forced_aligner` requires it from there, so gating the input client-side would have broken forced alignment.

## The other three

- **`/v1/audio/transcriptions` returns the transcript detail fields.** The route returned only `{text, timing}` while `task_result_json` already serialised segments, speaker turns and word timings, so every ASR model was routed away from the one endpoint that would return its timing data. The emitter is now shared by both routes, and both carry the sample rate the spans are counted in — without it a client cannot turn them into timestamps. Parakeet turns out to have been producing word timings all along.
- **The requested backend is proved before the port is bound.** `config.backend` defaults to CUDA and nothing validated it, so a machine without CUDA served the UI, the model list and the package installer, and failed only when the user pressed Run. The probe initialises and immediately frees a backend, so it costs one context creation at startup and reports the same message the engine would have raised, at the point an operator can still act on it.
- **Unload is gated and bounded.** `/v1/models/unload` and the unload-all path now require `--ui-management` like every other mutating UI route, snapshot the model list under `models_mutex_`, and acquire each model through `acquire_model_run` instead of an unbounded busy-wait. `live_ingest_limits` reads its state under the same mutex.

## Scope

Server only; no model, engine or output behaviour changes. Affects the `/v1/audio/transcriptions`, `/v1/tasks/run` and `/v1/models/unload` routes. Families checked against the option contract: Parakeet TDT, AudioSR, ControlFoley, MiDashengLM-Gen, PersonaPlex, HeartMuLa, Qwen3 forced aligner.

## Validation

```bash
cmake -S . -B build/macos-metal-tests -DCMAKE_BUILD_TYPE=Release \
  -DAUDIOCPP_BUILD_NATIVE_MODEL_MANAGER=ON \
  -DAUDIOCPP_USE_SYSTEM_OPENSSL=ON \
  -DOPENSSL_ROOT_DIR="$(brew --prefix openssl@3)"
ctest --test-dir build/macos-metal-tests --output-on-failure
```

Backend tested: Metal, Apple M4 Max, macOS 15.

Reproduced the 500 against a server built from `origin/main`, then confirmed the same request succeeds on this branch. End-to-end run on this branch: OmniVoice speech render, then the resulting WAV through Parakeet TDT on `/v1/audio/transcriptions`, which returned `"The quick brown fox jumps over the lazy dog."` with nine word timings and `sample_rate: 24000` — fields the route previously dropped.

## Known limitations

The backend probe adds one backend init/free to startup. On Metal that is milliseconds; on a large CUDA machine it is the usual context creation cost, paid once.

### DIFF
--- a/app/server/main.cpp
+++ b/app/server/main.cpp
@@ -225,6 +225,21 @@ int main(int argc, char ** argv) {
             throw std::runtime_error("--min-free-memory-mb must be >= 0 (0 disables the memory guard)");
         }
 
+        // Prove the requested backend and device exist before binding a port.
+        // The backend defaults to CUDA, and nothing else checks it until the
+        // first generate, so a machine without CUDA would otherwise serve the
+        // whole UI -- health, model list, downloads -- and only fail once the
+        // user pressed Run. Failing here reports the same message the engine
+        // would have raised, at the point the operator can still act on it.
+        {
+            engine::core::BackendConfig probe;
+            probe.type = config.backend;
+            probe.device = config.device;
+            probe.threads = config.threads;
+            ggml_backend_t backend = engine::core::init_backend(probe);
+            ggml_backend_free(backend);
+        }
+
         const auto ui_resource_anchor = executable_directory(argc > 0 ? argv[0] : nullptr);
         minitts::server::ServerState state(
             config,
--- a/app/server/runtime.cpp
+++ b/app/server/runtime.cpp
@@ -657,9 +657,72 @@ std::unordered_map<std::string, std::string> timing_headers(
     };
 }
 
+// Transcript detail arrays shared by /v1/tasks/run and /v1/audio/transcriptions.
+// ASR models that produce timestamps populate only these fields, so a route that
+// omits them silently discards work the model already did.
+template <typename FieldFn>
+void write_transcript_detail_fields(
+    std::ostringstream & out,
+    const engine::runtime::TaskResult & result,
+    FieldFn field) {
+    if (!result.speech_segments.empty()) {
+        field("segments");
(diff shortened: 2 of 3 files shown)


0xShug0:
As mentioned in another PR, please split into separate PRs: language-option contract, ASR detail response, backend startup probe, unload locking so I can test each of them individually.

PR #370, opened 2026-08-31
model_spec: correct the shared text_chunk_mode preset

CryptVenture:
## Summary

The shared `text_chunk_mode_full` option preset advertised a value the runtime parser rejects, and omitted the parser's own default, so a spec could neither declare the real default nor pass the value the preset promised.

## What changed

`option_presets()["text_chunk_mode_full"]` was `{word_budget, tag_aware, japanese, endline}`. The shared parser in `framework/text/chunking.cpp` accepts `{default, tag_aware, japanese, endline}`. A spec declaring `default` — the parser's fallback — failed preset validation, and a caller passing the advertised `word_budget` reached a parser that does not know it.

The preset is now `{default, tag_aware, japanese, endline}`, with a comment naming the parser it mirrors. `inflect_v2` is untouched: it keeps its own values list because it has a private parser that takes `word_budget` and none of the modes above.

`tests/unittests/test_model_spec_system.cpp` had a fixture whose default was the removed `word_budget`; it now uses `tag_aware`. The test asserts that a declared default is one of the preset values, which is exactly the invariant that was broken — the fixture was updated, not the assertion.

## Scope

Model-spec option validation only. No output, performance or memory change. Specs that reference the preset gain the ability to declare `default`; no shipped spec declared `word_budget`, so nothing that validates today stops validating.

## Validation

```bash
ctest --test-dir build/macos-metal-tests --output-on-failure -R model_spec_system_test
```

`model_spec_system_test` passes; the full suite is green (39/39). Sync check exits 0.

Backend tested: Metal, Apple M4 Max.

## Known limitations

There is still no automatic check that a preset's values match the runtime parser that consumes them, which is the invariant broken here. Adding one means teaching the spec layer about the text parser, so it is left out rather than bolted on.

### DIFF
--- a/src/framework/text/chunking.cpp
+++ b/src/framework/text/chunking.cpp
@@ -569,7 +569,7 @@ std::optional<TextChunkMode> parse_text_chunk_mode_override(
         return std::nullopt;
     }
     const std::string & value = match->value;
-    if (value == "default") {
+    if (value == "default" || value == "word_budget") {
         return TextChunkMode::Default;
     }
     if (value == "tag_aware" || value == "tag-aware" || value == "tagaware") {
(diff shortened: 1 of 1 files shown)


0xShug0:
This is a real bug, but I found "default: default" less clear. word_budget is a better user facing string. Added alias for default and PR merged.

PR #371, opened 2026-08-31
moss_voicegen: accept the instruction spelling the speech route sends

CryptVenture:
## Summary

MOSS-VoiceGen read only the option key `instruct`, while the server's speech route sends `instruction`. Every instruction sent from the WebUI or `/v1/audio/speech` was silently dropped, and the model generated as if none had been given.

## What changed

`src/community_models/moss_voicegen/session.cpp` now reads the option through `option_string(request.options, {"instruct", "instruction"}, "")`, accepting either spelling. The existing `instruct` key keeps working, so CLI callers and any saved presets are unaffected.

## Scope

One community model. The change affects generated output only in the case that was broken — a request that carries `instruction` now steers the model instead of being ignored. No performance or memory impact.

## Validation

```bash
ctest --test-dir build/macos-metal-tests --output-on-failure
```

Full suite green (39/39), sync check exit 0. Backend tested: Metal, Apple M4 Max.

## Known limitations

The fix was verified by reading both sides of the key — the server's speech route and the session's option read — not by generating audio: no MOSS-VoiceGen package is installed on this machine. The change is a one-line alias in the option lookup and does not touch the generation path.

### DIFF
--- a/src/community_models/moss_voicegen/session.cpp
+++ b/src/community_models/moss_voicegen/session.cpp
@@ -255,7 +255,10 @@ runtime::TaskResult MossVoiceGenSession::run(const runtime::TaskRequest & reques
 
     // The voice description arrives either as a request option or as a style tag on the
     // voice condition, matching how qwen3_tts takes its voice-design instruction.
-    std::string instruction = option_string(request.options, {"instruct"}, "");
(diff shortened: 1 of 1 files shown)


0xShug0:
PR merged! Thanks.

PR #375, opened 2026-08-31
webui: complete the shipped translations

CryptVenture:
## Summary

Seven keys were missing from all four translation files, and two more were missing from every language including English.

- `file.clear` and `file.preview` sit on every file input in the app, so their absence left the localised UI visibly half-English.
- `task.midi` labelled MuScriptor in English for every non-English user.
- `request.autoDuration`, `request.rewriteCaption`, `request.rewritingCaption` and `voice.configured` were likewise untranslated.
- `studio.subtitle.conversion` and `studio.subtitle.separation` did not exist at all, so the hero rendered the literal lookup key on those two tabs. The English entries are added alongside; these files carry the translations.

## `param.*` coverage

Coverage stood at roughly ten keys per language out of nearly three hundred.

Chinese gains 104 keys back-filled from the original Chinese label, info and placeholder text already in `model_params.json`, which the build discards because `catalog.ts` strips Han characters — so that wording was present in the repo but unreachable even for Chinese users. Italian, Polish and Russian gain 32 each, limited to entries with an English source; entries whose only source text is Chinese are left for a pass after those gain English variants.

Note for anyone adding `param.*` keys: they are namespaced by spec family, not by the key used in `model_params.json`. Nine groups there are keyed by catalog entry id, and the lookup resolves the family first, so a key written with the group name never fires.

Removes `arena.subtitle` and `models.reinstall` from the four files; neither is referenced by any component.

## Scope

Translation data and the two English strings the components needed. No engine, server or request-path changes; display text only.

## Validation

```bash
python3 -m json.tool   # each of the four language files
cd webui/native && npx svelte-check --tsconfig ./tsconfig.json   # 0 errors
cd webui/native && npm run build
```

Key sets of the four files verified identical to each other outside `param.*`, and every key present in the English source.

Backend tested: Metal, Apple M4 Max.

Depends on #372, #373 and #374. All three are contained in this branch, so the diff here includes them; review in that order.

## Known limitations

33 entries across the three European languages remain the English string. Each was checked and left deliberately — loanwords, acronyms and identical cognates such as Arena, Model, Backend, RTF and WER. The Russian `runtime.backend` is the one arguable case, where a Cyrillic transliteration would match the file's style; it is flagged rather than changed.


0xShug0:
Please 
1. Split #369 into separate server PRs
   One PR per behavior:
   - gate empty language forwarding by model contract
   - return ASR detail fields from /v1/audio/transcriptions
   - validate backend before binding UI port
   - fix unload locking/gating
2. Split #372 by spec ownership
   Do not change 16 model specs in one PR. Split into:
   - typo/default-only spec fixes with no runtime effect
   - strict contract fixes that change accepted/rejected requests
   - Music3/H3 new option declarations
   - forced-aligner/ASR option declarations
     Each PR should name the exact C++ parser it matches.
3. Split #373 into actual WebUI PRs
   Suggested pieces:
   - package resolver logic in catalog.ts
   - model catalog entry additions only
   - request wiring changes only: seed, max_tokens, duration, lyrics, required source/voice fields
   - transcript/segments/word timing display
   - package button labels/style only
4. Split #374 into parameter-surface PRs
   Suggested pieces:
   - remove controls that engine does not read
   - remove UI defaults that override engine/package defaults
   - range fixes that prevent hard 500s
   - add missing parameter groups, grouped by model family/type
5. Keep #375 translation-only
   It should only touch webui/native/lang/* and maybe English source keys. No model specs, catalog, UI behavior, or generated bundle mixed in.
6. Split #376
   Keep the checker/tool changes separate from catalog/spec corrections. The validation PR should prove existing catalog/spec state, not also rewrite that state.
General cleanup requirement: each PR should be rebased so its diff contains only its own files, include no unrelated prior PR commits, and state validation appropriate to the changed path. For model/UI request behavior, unit checks and svelte-check are not enough; include at least one real UI/server request shape or A/B evidence for the affected model path.

PR #381, opened 2026-08-31
docs(server): explain host and port options

mirek190:
Documents the server host bind-address and listening-port options in the detailed help output, including their current defaults (127.0.0.1 and 8080). Verified by rebuilding the Windows CUDA and Vulkan server targets and checking the help output from both binaries.

Screenshot 2026-08-31 134629


0xShug0:
Thank you @mirek190. PR merged!

PR #386, opened 2026-09-01
perf(fish_audio): add HIP Fast-AR top-k sampler

reezex0-ux:
## Summary

This is the first small HIP-scoped piece from the Fish S2 Pro AMD optimization work discussed in https://github.com/0xShug0/audio.cpp/discussions/317.

Fish Fast-AR produces a 4096-entry logits tensor for each acoustic codebook step. On the HIP path this PR:

- adds an exact device-side top-k helper for the Fast-AR logits;
- copies only the compact top-k result back to the host instead of all 4096 logits;
- keeps the existing temperature/top-p filtering and RNG selection path, so sampling behavior stays aligned with the current implementation;
- gates the new code on Linux HIP and only links it when `fish_audio` is part of the model composite.

Other backends keep the existing path unchanged.

This intentionally does **not** include the more aggressive experiments from the original patch set: chained Fast-AR, resident embedding gather, device-side mask/position updates, backend stream access, or the GGML convolution fixes. Those can be reviewed separately if useful.

## Validation

Tested against current `main` at `3497b7cc44753e2c141d8fe60ac42cec433e3281`.

System:
- AMD Radeon RX 7900 XTX (`gfx1100`)
- ROCm 7.2.1 / HIP
- Linux
- Fish Audio S2 Pro Q8_0 GGUF (`fish-audio-s2-pro-q8_0.gguf`)

Build:

```bash
bash scripts/build_linux.sh \
  --backend hip \
  --gpu-targets gfx1100 \
  --model-set custom \
  --models fish_audio \
  --target audiocpp_cli
```

Benchmark input:

```text
This benchmark measures speech generation on an AMD graphics card. The same sentence, random seed, and sampling settings are used for every run so that the timing results can be compared fairly.
```

Generation settings:

```text
language=en
seed=1234
max_tokens=96
temperature=0.8
top_k=50
top_p=0.8
```

Five identical requests were run in one loaded process. Request 1 was treated as graph/model warmup; requests 2-5 are the warm measurements below. Each output was 4411.79 ms of audio.

| Path | Warm #2 | Warm #3 | Warm #4 | Warm #5 | Warm mean |
| --- | ---: | ---: | ---: | ---: | ---: |
| current main | 4252.08 ms | 4316.97 ms | 4232.65 ms | 4229.31 ms | 4257.75 ms |
| HIP Fast-AR top-k | 4064.28 ms | 4133.41 ms | 4059.36 ms | 4055.67 ms | 4078.18 ms |

That is about **4.22% lower warm wall time** for this isolated optimization.

Output parity: all five baseline WAVs and all five patched WAVs were bit-identical:

```text
SHA-256: dfbc6ec75603a05c85d97e5dbf24b75a77dcc2386e5087159cd9d7d673989533
```

After a small code cleanup/refactor, I rebuilt/relinked and reran three requests; the warm times were 4038.98 / 4106.59 ms and the same SHA-256 was preserved for all three outputs.

I also compiled `src/models/fish_audio/ar.cpp` with the HIP Fish sampler macro removed to verify the existing non-HIP fallback path still compiles.

## Known limitations

- Runtime benchmark coverage is currently RX 7900 XTX / Linux HIP only.
- The fast path is used for positive `top_k <= 256`; unsupported values fall back to the existing full-logits path.
- This is intentionally only the Fast-AR top-k/readback optimization; the larger GPU-resident/chained work is not part of this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -945,6 +945,33 @@ audiocpp_add_model(fish_audio
         engine::models::fish_audio::make_fish_audio_loader
 )
 
+# Fish Fast-AR emits a 4096-entry logits tensor for each acoustic codebook.
+# On Linux HIP, use a small device-side exact top-k helper so only the compact
+# sampling candidates cross back to the host. Other backends stay unchanged.
+if (ENGINE_ENABLE_HIP AND UNIX AND NOT APPLE)
+    find_program(AUDIOCPP_HIPCC hipcc REQUIRED)
+    set(AUDIOCPP_FISH_HIP_FAST_SAMPLER_OBJ "${CMAKE_CURRENT_BINARY_DIR}/fish_hip_fast_sampler.o")
+    set(AUDIOCPP_FISH_HIP_TARGETS ${CMAKE_HIP_ARCHITECTURES})
+    if (NOT AUDIOCPP_FISH_HIP_TARGETS)
+        set(AUDIOCPP_FISH_HIP_TARGETS ${GPU_TARGETS})
+    endif()
+    set(AUDIOCPP_FISH_HIP_ARCH_FLAGS)
+    foreach(AUDIOCPP_FISH_HIP_ARCH IN LISTS AUDIOCPP_FISH_HIP_TARGETS)
+        list(APPEND AUDIOCPP_FISH_HIP_ARCH_FLAGS "--offload-arch=${AUDIOCPP_FISH_HIP_ARCH}")
+    endforeach()
+    add_custom_command(
+        OUTPUT ${AUDIOCPP_FISH_HIP_FAST_SAMPLER_OBJ}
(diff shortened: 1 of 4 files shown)


0xShug0:
Thanks @reezex0-ux! It overall looks good to me. One minor concern: the HIP sampler CMake block runs unconditionally whenever `ENGINE_ENABLE_HIP` is enabled, regardless of whether `fish_audio` is selected in `AUDIOCPP_LINKED_MODELS`, but `fish_audio` is the only consumer. If this sampler is general and can benefit other models on HIP we could consider promote it to framework so other models can also use it.

@IIIIIllllIIIIIlllll Could you take a look at this PR?

0xShug0:
@reezex0-ux I'm going to merge it now and follow-up PRs are appreciated!

Issue #380, opened 2026-08-31
Feadback: breeze-tts-2


IIIIIllllIIIIIlllll:
Long story short – I know nobody wants to read AI-generated walls of text that look detailed but are actually hollow.

I reached out to a few netizens who are interested in breeze-tts-2, and they all ran side‑by‑side comparisons using audio.cpp and the official framework. Everyone confirmed the same thing: audio.cpp can't match the official runtime with acceleration enabled in terms of raw peak performance, but it uses much less VRAM – under 8 GB, whereas the official one needs at least 16 GB (with acceleration). Without acceleration, the official default is eager mode, which gets completely crushed by audio.cpp.

CUDA (2080 Ti) → RTF 0.82
AMD HIP (8060S) → RTF 1.60

Even an oldie like the 2080 Ti can do real‑time generation! Hooray!

We've checked against the official runtime – everything matches, nothing to report.

The one thing worth mentioning is precision. Intuitively, fp32 activations feel "more correct" (higher precision), but the actual audio output is worse – mispronunciations and occasional runaway repetition. The model was trained and tuned in bf16, and only behaves well when you stick to bf16. For human ears, bf16 is the better choice. Those netizens and I all confirmed this – it's unlikely that all of us are hallucinating at the same time (joking). (Fixed on the breeze-bf16-activations branch.)

On AMD (HIP) there are still some performance and consistency quirks – it's slower than CUDA (which is expected for AMD), and previously the outputs weren't identical; both issues have been fixed in my own fork, and if needed, that could come in handy in the future.

Supplementary note: the RTF figures are based on tests run on the BF16-fixed branch; on the unfixed branch, performance can be up to 30% higher. This is a tough trade-off, and for now I haven't found a way to optimize it.

0xShug0:
@IIIIIllllIIIIIlllll Thank you for the feedback! There’s still room for performance optimization, which is why it’s still in dev. The trade-off is really about time: spending more time on deeper optimizations vs making models available sooner to a broader group of users who may struggle with Python setup, such as compiling FlashAttention. I hope the community can help drive further optimizations for this model. 

For optimization, you can check Higgs Audio TTS as a reference. It may require a combination of multiple techniques. If we confirm there’s a trade-off, the optimizations can be made opt-in.

0xShug0:
[...]

PR #393, opened 2026-09-02
Breeze-TTS 2 performance: weight packing + fused bf16 rounding

IIIIIllllIIIIIlllll:
Tested on RTX 2080 Ti (CUDA), gfx1151 (HIP), and Radeon 8060S (Vulkan). Fixed 100-token case, RTF:

| Backend | native | q4_k |
|---|---|---|
| CUDA 2080Ti | 0.71 (bit-identical output vs base) | 0.39 |
| HIP gfx1151 | 1.48 | 0.71 |
| Vulkan gfx1151 | 2.81 | 1.52 |

**Changes**

- **Pack QKV and gate/up projection weights** (`breeze_tts/generator.cpp`): load q/k/v as one packed tensor and gate/up as one packed tensor per layer (same approach as the higgs_audio_tts loader). Fewer, larger GEMMs per layer; bit-identical output on the native path, and it makes quantized weight storage (`_type=q4_k` etc., already supported by the framework) measurably faster. Also documents `weight_type` in the Breeze model page.
- **Fused round-to-bf16 unary op** (`GGML_UNARY_OP_ROUND_BF16 ggml CPU + CUDA/HIP): rounds f32→bf16→f32 in a single kernel, bit-identical to the cast round trip. On the 2080 Ti the cast pairs behind every bf16 activation rounding point cost ~19–29% of GPU time in small kernel launches. The qwen decoder's activation-cast policy gains a `fused_round` flag (default off; CUDA/HIP only). Supports f32/f16/bf16 inputs with contiguous rows; anything else falls back to the round trip.
- **`GGML_CUDA_GRAPHS_PRE_AMPERE` env var**: upstream disables graphs below sm_80; this adds an opt-in escape hatch for testing. Default behavior unchanged — on the 2080 Ti graphs were correct but RTF-neutral for this workload, so the default stands.

**Notes**

- The `fused_round` flag not enabled by any in-tree model yet — Breeze's activation-cast enablement comes with a separate bf16 activation fix (see #380: this model is trained/tuned in bf16 and the f32 path mispronounces). Models already using `activation_cast` (redae, fireredtts3, personaplex) can opt in later.
- Vulkan works correctly but is notably slower than HIP on the same AMD GPU (8060S: Vulkan 2.81 vs HIP 1.48 native RTF) — worth a separate look, not addressed here.

### DIFF
--- a/docs/models/breeze_tts.md
+++ b/docs/models/breeze_tts.md
@@ -67,3 +67,13 @@ audiocpp_cli \
 | `--request-option top_p=<f>` | `0..1` | `1.0` | Top-p sampling limit. |
 | `--request-option seed=<n>` | integer >= 0 | `0` | Generation seed. |
 | `--session-option breeze_tts.reference_cache_slots=<n>` | integer >= 0 | `1` | Prepared reference-audio cache slots. |
+| `--session-option weight_type=<type>` | `native`, `f32`, `f16`, `bf16`, `q8_0`, `q4_0`, `q4_k` | `native` | Weight storage type; quantized types convert at load time from the BF16 package. |
+
+Quantized weight storage is the largest measured speedup and applies to CUDA
+and HIP alike: `q8_0` cut the fixed 100-token regression case from RTF ~1.5 to
+~0.95 on gfx1151 and from ~0.77 to ~0.56 on an RTX 2080 Ti, and `q4_k` reached
+~0.84 / ~0.49 respectively, with no audible quality regression in the Chinese
+voice-design regression cases. Counter to intuition, fp32 is the one
+configuration known to be *worse* for this model (mispronunciations and
+runaway repetition), because the model is trained and tuned in bf16.
+
--- a/external/ggml/include/ggml.h
+++ b/external/ggml/include/ggml.h
@@ -615,6 +615,7 @@ extern "C" {
         GGML_UNARY_OP_CEIL,
         GGML_UNARY_OP_ROUND,
         GGML_UNARY_OP_TRUNC,
+        GGML_UNARY_OP_ROUND_BF16,
 
         GGML_UNARY_OP_COUNT,
     };
@@ -1255,6 +1256,12 @@ extern "C" {
             struct ggml_tensor  * a);
 
     GGML_API struct ggml_tensor * ggml_trunc_inplace(
+            struct ggml_context * ctx,
+            struct ggml_tensor  * a);
+
+    // Rounds each element to bf16 precision, stored as f32. Equivalent to a
+    // cast f32 -> bf16 -> f32 round trip, but fused into a single op.
+    GGML_API struct ggml_tensor * ggml_round_bf16(
             struct ggml_context * ctx,
             struct ggml_tensor  * a);
 
--- a/external/ggml/src/ggml-cpu/ggml-cpu.c
+++ b/external/ggml/src/ggml-cpu/ggml-cpu.c
@@ -2260,6 +2260,7 @@ static int ggml_get_n_tasks(struct ggml_tensor * node, int n_threads) {
                 case GGML_UNARY_OP_CEIL:
                 case GGML_UNARY_OP_ROUND:
                 case GGML_UNARY_OP_TRUNC:
+                case GGML_UNARY_OP_ROUND_BF16:
                     {
                         n_tasks = 1;
                     } break;
--- a/external/ggml/src/ggml-cpu/ops.cpp
+++ b/external/ggml/src/ggml-cpu/ops.cpp
@@ -10068,6 +10068,10 @@ void ggml_compute_forward_unary(
             {
                 ggml_compute_forward_trunc(params, dst);
             } break;
+        case GGML_UNARY_OP_ROUND_BF16:
+            {
+                ggml_compute_forward_round_bf16(params, dst);
+            } break;
         case GGML_UNARY_OP_XIELU:
             {
                 ggml_compute_forward_xielu(params, dst);
--- a/external/ggml/src/ggml-cpu/unary-ops.cpp
+++ b/external/ggml/src/ggml-cpu/unary-ops.cpp
@@ -97,6 +97,10 @@ static inline float op_trunc(float x) {
     return truncf(x);
 }
 
+static inline float op_round_bf16(float x) {
+    return bf16_to_f32(f32_to_bf16(x));
+}
+
 template <float (*op)(float), typename src0_t, typename dst_t>
 static inline void vec_unary_op(int64_t n, dst_t * y, const src0_t * x) {
     constexpr auto src0_to_f32 = type_conversion_table<src0_t>::to_f32;
@@ -320,6 +324,10 @@ void ggml_compute_forward_round(const ggml_compute_params * params, ggml_tensor 
 
 void ggml_compute_forward_trunc(const ggml_compute_params * params, ggml_tensor * dst) {
     unary_op<op_trunc>(params, dst);
+}
+
+void ggml_compute_forward_round_bf16(const ggml_compute_params * params, ggml_tensor * dst) {
+    unary_op<op_round_bf16>(params, dst);
 }
 
 void ggml_compute_forward_xielu(const ggml_compute_params * params, ggml_tensor * dst) {
--- a/external/ggml/src/ggml-cpu/unary-ops.h
+++ b/external/ggml/src/ggml-cpu/unary-ops.h
@@ -28,6 +28,7 @@ void ggml_compute_forward_floor(const struct ggml_compute_params * params, struc
 void ggml_compute_forward_ceil(const struct ggml_compute_params * params, struct ggml_tensor * dst);
 void ggml_compute_forward_round(const struct ggml_compute_params * params, struct ggml_tensor * dst);
 void ggml_compute_forward_trunc(const struct ggml_compute_params * params, struct ggml_tensor * dst);
+void ggml_compute_forward_round_bf16(const struct ggml_compute_params * params, struct ggml_tensor * dst);
 void ggml_compute_forward_xielu(const struct ggml_compute_params * params, struct ggml_tensor * dst);
 
 #ifdef __cplusplus
--- a/external/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/external/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -5365,8 +5365,11 @@ static bool ggml_backend_cuda_device_supports_op(ggml_backend_dev_t dev, const g
                     //return ggml_is_contiguous_rows(op->src[0]);
                     return ggml_is_contiguous(op->src[0]);
                 case GGML_UNARY_OP_ROUND_BF16:
-                    // dst is contiguous by construction; src rows must be contiguous for the strided kernel.
-                    return ggml_is_contiguous(op) && ggml_is_contiguous_rows(op->src[0]);
+                    // f32/f16/bf16 src with contiguous rows, contiguous f32 dst.
+                    return (op->src[0]->type == GGML_TYPE_F32 || op->src[0]->type == GGML_TYPE_F16 ||
+                            op->src[0]->type == GGML_TYPE_BF16) &&
+                           op->type == GGML_TYPE_F32 && ggml_is_contiguous(op) &&
+                           ggml_is_contiguous_rows(op->src[0]);
                 default:
                     return false;
             }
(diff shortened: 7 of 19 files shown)


IIIIIllllIIIIIlllll:
My bad. During testing, I set cfg=4, which was too large and caused quality degradation. Both BF16 and Q4 exhibited the issue, so it's not entirely a Q4 problem. This means the performance benefit from Q4 is still worth considering. This is an autoregressive model and is bandwidth‑sensitive.

Text:
`You are not crazy. A 4 bit quant scores about 77% on IFEval - that's with thinking disabled. Reasoning has a negative impact on instruction following and with Qwen's xhigh reasoning.`

CFG = 4, HIP, seed = 42:

[tts-3feafd4e.wav](https://github.com/user-attachments/files/31719540/tts-3feafd4e.wav)

CFG = 1, HIP, seed = 42:
[tts-d4954024.wav](https://github.com/user-attachments/files/31719597/tts-d4954024.wav)

CFG = 4, official, seed = 42
[tts-official.wav](https://github.com/user-attachments/files/31727563/tts-official.wav)

With CFG = 4, the last word “reasoning” in this audio is missing; with CFG = 1, it’s normal. All parameters, including the seed, are the same — only the CFG value differs.

IIIIIllllIIIIIlllll:
Strange — I'm currently running into a weird issue. The content generated with the Vulkan backend and the HIP backend is coming out wrong. I'll keep investigating.

IIIIIllllIIIIIlllll:
> Strange — I'm currently running into a weird issue. The content generated with the Vulkan backend and the HIP backend is coming out wrong. I'll keep investigating.

@0xShug0 After more testing, I found that the official framework also suffers from quality degradation under specific seeds… Specifically, with long text, the volume of the last sentence becomes quieter. I kept thinking there was a bug somewhere, but it turns out the official implementation reproduces it too. So far I haven't found any other bugs — if you have time, feel free to take a look.

IIIIIllllIIIIIlllll:
So far, my testing approach has been roughly as follows:

1. Discovered that the audio produced by HIP and CUDA differed greatly; upon inspection, it turned out to be an issue with the philox path — only CUDA uses the philox path.

2. Found that HIP performance is too poor; nothing much to say about that, just optimize it.

3. Found that the audio quality was much worse compared to the official output, even with mispronunciations; confirmed it was an issue with F32 and BF16, as mentioned earlier.

4. Continued optimizing performance, referencing Higgs Audio TTS.

5. Discovered that the generated audio content would randomly "collapse" with terrible quality. I spent a whole afternoon checking this and found that the official framework also exhibits this issue. Yes, it's just the luck of chaotic randomness — at least on my 2080TI + gfx1151, it definitely occurs.

IIIIIllllIIIIIlllll:
> Strange — I'm currently running into a weird issue. The content generated with the Vulkan backend and the HIP backend is coming out wrong. I'll keep investigating.

This is a summary of the 'quality degradation' issue that I have observed.

# Breeze TTS tail-volume drop: not an engine bug

Reported symptom: with certain prompt/style combos, the last sentence of
Breeze TTS output gets noticeably quieter, observed on CUDA but not Vulkan
with the same seed.

Verdict: model-intrinsic trajectory degeneration, not a backend bug.

Why:

- The sampler is a bit-faithful emulation of PyTorch's CUDA Philox
  multinomial stream. Same-run replay reproduces 5415/5415 sampled tokens,
  and the same replay rule reproduces the official reference's picks from
  official logits. CUDA generation is fully deterministic (identical codes
  across runs).
- Breeze sampling is an exponential race over logits, so picks flip on
  sub-bf16-ulp logit noise; one flip diverges the whole trajectory. Every
  (backend RNG stream x seed) pair is a different chaotic trajectory with
  some chance of landing in a degenerate attractor (fading tail, length
  inflation). CUDA uses Philox emulation; HIP/Vulkan use mt19937 — same seed,
  different roll of the dice.
- The official PyTorch reference collapses on roughly half of the tested
  seeds for the same prompt (confirmed by force-decoding official codes
  through the engine's decoder, same scale). Engine CUDA and Vulkan each
  collapse on different seeds at a similar rate. No backend is systematically
  worse.

Practical advice: re-roll with a different seed. Systematic mitigation
(repetition penalty / temperature scheduling on the depth sampler) would
deviate from reference behavior and is a product option, not a fix.

0xShug0:
@IIIIIllllIIIIIlllll Thanks for the investigation! I’ll be AFK most of the day today and will look at the issues later tonight. The model has two sampling stages, which MAY make it prone to collapse due to small token dirft. Or it could be due to training data. For most of the models replacing the sampling impl is acutally fine and won't cause collapse.

0xShug0:
@IIIIIllllIIIIIlllll Could you isolate the transpose1d fast path changes into a separate PR? It would make it easier to track commits that introduce behavioral changes. Also if all models are safe for this change (no performance regression, and no crash due to layout) the the function really doesn't need to mix `uses_ggml_cuda_or_hip_backend` with direct backend names. 

Please run regression tests on at least the popular models and keep the status updated in the PR. I can fill in the remaining models afterward.

IIIIIllllIIIIIlllll:
> @IIIIIllllIIIIIlllll Could you isolate the transpose1d fast path changes into a separate PR? It would make it easier to track commits that introduce behavioral changes. Also if all models are safe for this change (no performance regression, and no crash due to layout) the the function really doesn't need to mix `uses_ggml_cuda_or_hip_backend` with direct backend names.
> 
> Please run regression tests on at least the popular models and keep the status updated in the PR. I can fill in the remaining models afterward.

OK

0xShug0:
@IIIIIllllIIIIIlllll Merged!  It's on dev so have some time to test before release. Found the old doc for lifting the Metal fast path. Back then we only had ~20 models. Now it’s 60+, so full regression is probably too much to ask. https://github.com/0xShug0/audio.cpp/blob/main/docs/reports/metal_convtranspose_performance.md.

IIIIIllllIIIIIlllll:
> @IIIIIllllIIIIIlllll Merged! It's on dev so have some time to test before release. Found the old doc for lifting the Metal fast path. Back then we only had ~20 models. Now it’s 60+, so full regression is probably too much to ask. https://github.com/0xShug0/audio.cpp/blob/main/docs/reports/metal_convtranspose_performance.md.

I will try to test a portion of them.

IIIIIllllIIIIIlllll:
Regression results for the conv_transpose1d col2im fast path on Vulkan. Tested on Radeon 8060S (RADV), comparing `3935ee0^` vs `3935ee0` so the col2im gate change is the only difference between the two builds. Fixed seeds, offline inference, same machine.

| Model | Waveform corr (pre vs post) | RTF pre → post | Notes |
|---|---|---|---|
| breeze_tts | 0.99998 | 1.94 → 1.90 | numerically equivalent |
| qwen3_tts | 1.00000 | 0.87 → 0.81 | identical |
| omnivoice | 1.00000 | 1.27 → 1.25 | identical |
| index_tts2 | 1.00000 | 1.62 → 1.61 | identical (fast path via BigVGAN vocoder) |
| vibevoice | 0.67 | 1.71 → 0.82 | see below |

No crashes, no layout errors, no perf regressions; vibevoice is ~2x faster with the fast path.

VibeVoice's waveform divergence comes from the model, not the change: its diffusion sampler amplifies any tiny numeric difference. Cross-check — HIP (which has always used the col2im path) vs Vulkan *without* this change diverges even more (waveform corr 0.48), so the delta introduced by enabling col2im on Vulkan is smaller than an ordinary backend swap. Output is deterministic per build (same-build rerun corr = 1.0).

IIIIIllllIIIIIlllll:
Extended regression pass on the col2im gate (pre = 6f64b48, post = 3935ee0, Radeon 8060S / gfx1151, Vulkan, fixed seed, identical requests):

| Model | Route to ConvTranspose1dModule | Waveform corr | RTF pre → post | Verdict |
|---|---|---|---|---|
| higgs_audio_tts | own codec | 0.99998 | 0.579 → 0.535 | pass |
| fish_audio | fish_dac codec | 1.00000 | 1.429 → 1.286 | pass |
| pocket_tts | mimi codec | 0.999996 | 0.095 → 0.080 (steady state, 2 reruns each) | pass |
| chatterbox | s3gen_flow | 1.00000 | 0.839 → 0.696 | pass |
| htdemucs | demucs pipeline | 1.00000 on vocals (content stem) | 0.119 → 0.101 | pass — bass/drums/other stems sit at the −90 dB noise floor (rms ≈ 4e-5) for a speech input, so their raw corr is noise-on-noise |
| ace_step | control: own col2im predating this change | 0.99981 | 0.821 → 0.824 | pass — untouched paths stable |
| cosyvoice3 | does not use ConvTranspose1dModule | 1.00000 | 17.6 → 17.4 | no regression; see note below |

Note on cosyvoice3: both builds emit the same near-empty 0.06 s output on this GPU, bit-identical pre/post. That is a pre-existing issue on the 8060S and is **not related to this change** — cosyvoice3 never routes through ConvTranspose1dModule, and the identical pre/post output confirms this PR doesn't affect it.

RTF deltas are attributable to this PR alone: pre and mid differ by exactly this one commit, and repeated runs are stable within each build. Magnitude scales with how conv-transpose-heavy the codec/vocoder is.

Together with the earlier 5 models, coverage is now 12 models. Full procedure + per-model run scripts: `docs/reports/conv_transpose1d_col2im_vulkan_regression.md` on our side.

0xShug0:
@IIIIIllllIIIIIlllll Looks good. Didn't realize cosyvoice3 was that slow. I’ve been fighting with an interesting model for a few days. I’ll try to pick up cosyvoice3 before merging dev into main or just remove it from dev before the merge.

IIIIIllllIIIIIlllll:
Correction on the RTF numbers in my previous two comments: those runs executed all `pre` builds before all `mid` builds, and the CLI wall clock includes the driver's one-time shader compile — so `mid` inherited a warm shader cache and some deltas were inflated. Re-measured warm + interleaved (≥2 runs per build, pre/mid alternating) on the 8060S:

| Model | RTF pre → post (was) | RTF pre → post (warm, corrected) |
|---|---|---|
| breeze_tts | 1.94 → 1.90 | 1.931 → 1.871 |
| qwen3_tts | 0.87 → 0.81 | 0.870 → 0.815 |
| omnivoice | 1.27 → 1.25 | 1.264 → 1.244 |
| index_tts2 | 1.62 → 1.61 | 1.590 → 1.588 |
| vibevoice | 1.71 → 0.82 | 1.514 → 0.820 (**real ~1.85×**) |
| higgs_audio_tts | 0.579 → 0.535 | 0.535 → 0.525 |
| fish_audio | 1.429 → 1.286 | 1.409 → 1.284 |
| pocket_tts | 0.095 → 0.080 | 0.095 → 0.082 |
| chatterbox | 0.839 → 0.696 | 0.715 → 0.714 (**no real change**) |
| htdemucs | 0.119 → 0.101 | 0.113 → 0.098 |
| ace_step (control) | 0.821 → 0.824 | 0.833 → 0.823 |

Takeaways unchanged: no correctness regression anywhere (all waveform corr ≥ 0.9998 except chaotic vibevoice, which was qualified separately), and the speedup is real where conv-transpose is a meaningful share of runtime — vibevoice ~1.85×, pocket_tts's mimi decoder stage 60.7 → 19.8 ms (3×) on the 2080 Ti. Apologies for the noisy first numbers.

IIIIIllllIIIIIlllll:
@0xShug0 That should be all I can do.

IIIIIllllIIIIIlllll:
Oh! There's also room for optimization in terms of VRAM usage, but that's a separate piece of work — and I'm already giving it a try. The direction is: split peech_encoder.cpp into two graphs; the transformer graph; and the O(T²) time complexity of the encoder attention — that's a major overhaul.

0xShug0:
> Oh! There's also room for optimization in terms of VRAM usage, but that's a separate piece of work — and I'm already giving it a try. The direction is: split peech_encoder.cpp into two graphs; the transformer graph; and the O(T²) time complexity of the encoder attention — that's a major overhaul.

Thank you for keeping pushing on this!  Sorry I dropped the ball on the new models. I reached out to the Breeze team and see if they could help.

IIIIIllllIIIIIlllll:
> > Oh! There's also room for optimization in terms of VRAM usage, but that's a separate piece of work — and I'm already giving it a try. The direction is: split peech_encoder.cpp into two graphs; the transformer graph; and the O(T²) time complexity of the encoder attention — that's a major overhaul.
> 
> Thank you for keeping pushing on this! Sorry I dropped the ball on the new models. I reached out to the Breeze team and see if they could help.

What's going on?

0xShug0:
> What's going on?

Haha,  just on a short vacation this week and also got distracted by LiveAvatar audio-to-video model :)

IIIIIllllIIIIIlllll:
> > What's going on?
> 
> Haha, just on a short vacation this week and also got distracted by LiveAvatar audio-to-video model :)

**Haha, I thought I had run into some difficult technical issue. I'll do my best to optimize this model thoroughly, so you'll have less work to do later.**

PR #256, opened 2026-08-16
Add VoxCPM v1 — lightweight VoxCPM TTS support (0.5B)

jasonchen31:
## 1. Overview

Adds support for the **[OpenBMB VoxCPM-0.5B](https://github.com/OpenBMB/VoxCPM)** lightweight TTS model to audio.cpp (ported from [VoxCPM.cpp](https://github.com/bluryar/VoxCPM.cpp)), reusing the existing and already-released `voxcpm2` model tree:

| Model | Params | Output sample rate | GGUF file |
|---|---|---|---|
| VoxCPM-0.5B | 0.5B | **16 kHz** | `voxcpm-0.5b-q8_0-audiovae-f16.gguf` |

You may find the converting tools in [VoxCPM.cpp project](https://github.com/bluryar/VoxCPM.cpp/tree/main/scripts)

GGUF files are [here](https://huggingface.co/bluryar/VoxCPM-GGUF).

Architecture (0.5B): VAE encoder 128 / decoder 1536, encoder_rates `[2,5,8,8]`, decoder_rates `[8,8,5,2]`, patch_size 2, residual_lm 6 layers, encoder/dit 4 layers, 16 kHz, max_len 4096.

Since the v1 GGUF stores a different tensor convention than v2 (folded AudioVAE weights, no `weight_v`/`weight_g` split, no `sr_cond_model` tensors, `voxcpm` architecture name), the port wraps the v2 loader with a GGUF tensor-adaptation layer and adds `config.v1`-guarded branches in the generator, mirroring the reference implementation (`VoxCPM.cpp`).

All work is on `main`, 21 commits ahead of `upstream/main` (merge-base `4e973b1`), consisting of 12 porting commits plus merges. `tensor_source.h` differs from upstream by a **single line** (`is_synthesized`) and `tensor_source.cpp` is identical — the port is structured to be upstreamable.

---

## 2. Porting activities

1. **Regenerated the 0.5B `config.json` from the GGUF metadata** — the previously shipped sidecar was wrong on ~8 axes (patch, residual_lm/encoder/dit layer counts, VAE dims and rates, sample rate 44.1 kHz vs the actual 16 kHz, max_len).
2. **Diagnosed the v1 GGUF conventions** (tensor dump + reference converter analysis):
   - AudioVAE conv weights are stored **already folded** (weight-norm folded), with no `weight_v`/`weight_g` decomposition and no `sr_cond_model.*` tensors.
[...]

PR #394, opened 2026-09-02
feat: add Chatterbox Turbo TTS model family

pannagaps:
## Summary

- Adds `chatterbox_turbo` as a new model family: Resemble AI's distilled 350M-parameter Chatterbox variant (GPT2-style T3 backbone, GPT2 BPE tokenizer with 19 built-in emotion/style tags, 2-step meanflow-distilled S3Gen decoder), loaded from `cstr/chatterbox-turbo-GGUF` (a third-party GGUF conversion, MIT-relicensed by `cstr` for their `CrispASR` project — not published by ResembleAI or audio.cpp).
- Extends the shared `chatterbox` S3Gen flow decoder to support meanflow distillation: a second "r" (end-time) sinusoidal embedding mixed into the time embedding via a new `time_embed_mixer` linear layer, and a non-CFG, non-batch-doubled Euler solve (`compute_s3_flow_cfm_meanflow`). This is gated behind detecting a `flow.decoder.estimator.time_embed_mixer.weight` tensor, so the base Chatterbox 10-step CFG decoder path is unaffected.
- Pins the S3 flow encoder's token embedding lookup table to F16 regardless of requested weight storage type, since `ggml_get_rows` on CUDA doesn't support K-quant lookups for that table (needed for Turbo's K-quant S3Gen package).
- Documents the new family in `docs/tts.md` and `README.md`, and adds `model_specs/chatterbox_turbo.json`.

**Current limitation:** only the built-in default voice baked into the T3 GGUF is supported; custom voice cloning via `--voice-ref` is rejected with an explicit error, since it depends on the turbo GGUF's speaker-encoder/S3-tokenizer sections, whose exact tensor layout hasn't been validated yet. Marked `status: testing` in the model spec until that lands.

## Framework changes

`src/models/chatterbox/{s3gen_flow.cpp,s3gen_inference.cpp}` and `include/engine/models/chatterbox/s3gen_flow.h` (shared with the base `chatterbox` family) were touched to add the meanflow decoder path. The base Chatterbox 10-step CFG decoder is unaffected: the new code paths only activate when `S3FlowDecoderWeights::meanflow` is set (detected from the presence of the `time_embed_mixer` tensor), which the base GGUF package does not have.

## Test plan

- [x] `python3 tools/check_loader_catalog_sync.py --self-test && python3 tools/check_loader_catalog_sync.py` — passes, loader/spec/catalog in sync
- [x] Build with `audiocpp_cli --task clon --family chatterbox_turbo --model models/Chatterbox-Turbo-GGUF/chatterbox-turbo-t3-q8_0.gguf --backend cuda --text "Hello from Chatterbox Turbo." --out out.wav`
- [x] Confirm base `chatterbox` family output is unchanged (regression check on the shared S3Gen flow decoder)


0xShug0:
@pannagaps Thanks for the PR! I’ll be AFK most of the day today and will test the PR tonight.

0xShug0:
@pannagaps Could you move the models to the community_models/? The core models are expected to have runtime tests on at least CUDA, Vulkan, and Metal and will go throught more tests. Since the status is "testing", it's better to keep it under community models so users have the right expectations around support and compatibility.

Also please repack the GGUF to make it audio.cpp-native (self-contained, spec v1) and provide conversion scripts. Otherwise, we end up adding unnecessary complexity and small compatibility layers just to support third-party GGUFs, all of which could be avoided with a proper GGUF offline conversion. You can check this PR and commit to get the idea -- https://github.com/0xShug0/audio.cpp/pull/256/changes/cbd79302ffeb076b544e2dcc088e34c8c4250071

pannagaps:
Got it, makes sense. I’ll move the models to community_models/ and repack the GGUF to be audio.cpp-native, along with the conversion scripts.

pannagaps:
hi @0xShug0 , 
Im done with the changes requested please verify and let me know.

thanks

0xShug0:
@pannagaps Thanks! Mergd. 

There are some contract issues and I will fix them after merge.

PR #397, opened 2026-09-02
feat: support set mirror by env:HF_ENDPOINT when downloading models

feng19:


0xShug0:
@feng19 Thanks! PR merged.

PR #403, opened 2026-09-02
model_specs: correct descriptions that contradict the code

CryptVenture:
Split out of #372 as requested. This PR is description text only; the strict contract fixes, the VoxCPM1 defaults, the ASR/aligner/codec declarations and the MiniMax declarations are separate PRs.

Nothing here changes a declared type, bound, default or required flag, so no request that was accepted or rejected before changes state.

## The changes

| Spec | Change | Code it matches |
|---|---|---|
| `dramabox` | five descriptions dropped the claim that their default "comes from `config.json`" | `dramabox/assets.cpp` reads `config.json` for the architecture only; the values named are the engine's own |
| `fish_audio.reference_text` | says it is required whenever the request carries inline reference audio | `src/models/fish_audio/session.cpp:255-258` throws `"<role> with inline reference audio requires reference_text option"` |
| `outetts.reference_text` | same, for voice cloning | `src/community_models/outetts/session.cpp:695-698` and `722-728` throw `"OuteTTS voice cloning requires --reference-text"` |

Both `reference_text` options stay `required: false` on purpose: they are required only when reference audio is present, and the schema's `required` is a plain bool with no conditional form, so marking them required would break plain TTS with a built-in voice. The description is the only place that distinction can currently live.

## Validation

```bash

# parses every on-disk spec through the C++ schema loader
```

## Scope

Three spec files, 7 lines. The generated WebUI bundle (`webui/native/dist/index.html`) is deliberately not included: `catalog.ts` inlines `model_specs/*.json` at frontend build time and the bundle is not byte-reproducible, so regenerating it in each PR of this split would produce conflicts between them. Happy to send one bundle-regeneration PR once the series has landed.

### DIFF
--- a/model_specs/dramabox.json
+++ b/model_specs/dramabox.json
@@ -57,7 +57,7 @@
       {
         "name": "num_inference_steps",
         "type": "int",
-        "description": "Diffusion sampling step count; default comes from config.json, 30 in the current package.",
+        "description": "Diffusion sampling step count; default 30.",
         "required": false,
         "min": 1,
         "default": 30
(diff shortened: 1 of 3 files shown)


CryptVenture:

PR #408, opened 2026-09-02
webui: resolve install packages by target directory, not a family allowlist

CryptVenture:
Split out of #373 as requested. This PR is the package resolver in `catalog.ts` and the button rendering it drives; the catalog entries, the request wiring, the transcript display and the UI text fixes are separate PRs.

## The problem

`catalog.ts` grouped install choices with a hardcoded six-family allowlist. Families on the list exposed every package; every other family was narrowed to at most one q8 and one fp16 button. Measured by executing `catalog.ts` against the real specs, **130 of the published packages were reachable from the WebUI**.

## The change

- **Group by `target_directory`, not by family name.** Packages that are precision variants of one model share a directory; packages that are different models do not. The allowlist had to be hand-edited for every new family and silently hid the rest.
- **Extend the precision-suffix list** to `q2_k`…`q6_k`, `q4_0`, `q5_0`, `f32`. It covered only q8/f16/bf16/safetensors/orig, so an entry whose `download_id` ended in an uncovered suffix collapsed to that single package and hid its siblings — which is why PersonaPlex offered only Q4_K while its own spec marks Q8_0 default.
- **Honour `ui.recommended_package`.** Every spec has one, all resolve to a real package, and nothing read the field.
- **Match relatedness on `target_directory`, not an id prefix.** The prefix test pulled the IndexTTS2.5 packages into the IndexTTS2 entry, whose fp16 slot was then decided by spec-file array order.
- **Keep an entry with no installable package**, flagged, instead of letting the model vanish from the UI.

### Why the labels are in this PR

The button text is chosen while the choice is built, in the same function. Shipping the two separately would leave one PR with a label nothing produces, or buttons whose text does not fit their border. Each entry's buttons are labelled by build (`Q8_0`, `Q4_K ConvRot`); where several models share a directory and collide on precision the label keeps what their display names do not share (`XL Turbo BF16`), falling back to the full name if even that is ambiguous. The full name stays on the `title` and `aria-label`.

## Validation

```
```

Reachability measured by loading `catalog.ts` through Vite's SSR loader and diffing the resolved package-id sets:

| | `main` | this branch |
|---|---|---|
| catalog entries | 85 | 85 |
| install buttons rendered | 150 | 166 |
| distinct packages reachable | 130 | 143 |
| packages lost | — | 0 |

Label check over every entry: 0 entries with two buttons reading alike, longest label 13 characters (`XL Turbo BF16`).

## Scope

`catalog.ts`, `types.ts`, the package-button block of `+page.svelte`, and the button rules in `app.css`. The generated bundle (`webui/native/dist/index.html`) is deliberately excluded: it is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict with each other. Happy to send one bundle-regeneration PR once the series lands.

### DIFF
--- a/webui/native/src/app.css
+++ b/webui/native/src/app.css
@@ -156,7 +156,7 @@ button:disabled { opacity: .45; cursor: not-allowed; }
 button.primary, button.run { color: var(--text-invert); background: linear-gradient(135deg, var(--cyan), #5ac7ff); border: 0; font-weight: 800; }
 .button-row button { flex: 1; }
 .studio-package-buttons { display: grid; grid-template-columns: repeat(3, minmax(0, 1fr)); gap: 5px; margin-top: 11px; }
-.studio-package-buttons button, .single-model-toggle { position: relative; min-width: 0; min-height: 38px; padding: 6px 4px; font-size: 9px; line-height: 1.15; }
+.studio-package-buttons button, .single-model-toggle { position: relative; min-width: 0; min-height: 38px; padding: 6px 4px; overflow: hidden; font-size: 9px; line-height: 1.15; overflow-wrap: anywhere; }
 .studio-package-buttons button.selected-package:not(:disabled) { border-color: var(--active-border); color: var(--cyan); box-shadow: inset 0 0 0 1px color-mix(in srgb, var(--cyan) 10%, transparent); }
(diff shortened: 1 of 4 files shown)


CryptVenture:

PR #410, opened 2026-09-02
webui: add the catalog entries for models that ship but cannot be selected

CryptVenture:
Split out of #373 as requested: catalog entry additions only. Nothing existing is modified — the corrections to existing entries, the resolver, the request wiring, the transcript display and the UI text fixes are separate PRs.

## The problem

Four families publish installable GGUF packages that no catalog entry reached, so the WebUI could neither show nor install them:

| Family | What ships | Why it matters |
|---|---|---|
| `moss_voicegen` | one GGUF package | Apache-2.0, active loader, and a matching UI tab already exists — only the catalog entry was missing |
| `f5_tts` | Habibi unified plus seven Arabic dialect checkpoints | nine installable packages, zero entries; the non-commercial licence is named in the input hint |
| `pocket_tts` | German, Italian, Portuguese, Spanish | shipped, unreachable; each has no built-in voice embedding, so the hint says a reference upload is required |
| `dots_tts` | the Edit checkpoint | speech editing, shipped and unreachable |

## The change

Fourteen entries, each naming an exact `packages[].id` and the directory that package installs into, inserted next to their family's existing entries. Existing entries are untouched.

The one reachability warning left after this PR is `f5_tts`'s `vocos_mel_24khz`, a vocoder support package rather than a model — it has no UI tab of its own and is installed as a dependency.

## Validation

```bash
# the two "publishes installable packages but no catalog entry" warnings
# (f5_tts, dots_tts_edit) are gone; the pre-existing findings on other
# entries are the subject of the sibling corrections PR and are unchanged
```

Reachable packages, measured by loading `catalog.ts` through Vite's SSR loader:

| | entries | packages reachable |
|---|---|---|
| `main` | 85 | 130 |
| this branch | 99 | 149 |
| this branch + the resolver PR | 99 | 162 |

## Scope

One data file, 29 added lines, no deletions. No code, spec or engine change. The generated bundle is deliberately excluded — `catalog.ts` inlines this file at frontend build time and the bundle is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict; happy to send one bundle-regeneration PR once the series lands.


CryptVenture:

PR #411, opened 2026-09-02
webui: stop overriding engine defaults on every request

CryptVenture:
Split out of #373 as requested. This PR is the request wiring only; the package resolver, the catalog entries, the transcript display and the UI text fixes are separate PRs.

## The problem

Three request fields were filled in by the UI regardless of what the model asked for.

**`max_tokens`** — every request carried a fixed `1024`. Engine defaults across the shipped specs range from 500 (`dots_tts`, `magpie_tts`) through 1520 (`confucius4_tts`) to 2000 (`muscriptor`), and several families default to 0, meaning no cap at all. The UI's number was neither the model's default nor the user's choice.

**`seed`** — the UI defaulted to `1234` in a field whose own label describes a random seed. Families whose engine default is a fresh random seed therefore returned identical audio on every run.

**Controls the model rejects** — lyrics and duration were offered for models that throw on them; voice design and forced alignment did not require the inputs their engines demand; the conversion and analysis tasks declare a seed and were never sent one, so their runs could not be reproduced at all.

## The change

- `max_tokens` is blank by default and sent only when the user types a value.
- `seed` defaults to `-1`, resolved client-side to a random uint32 before sending, so the engine never sees a negative value (it parses seeds as unsigned — `-1` on the wire is a 500).
- Tasks that declare a seed in their spec now receive the resolved seed.
- Lyrics and duration are hidden for models that reject them; voice design requires a voice description; forced alignment requires a transcript and a language. The engine throws for each of these — a form message is a better place to learn it than a raw engine error.
- Built-in voices from the spec and per-model voices from the server were both plumbed and never read; they are now offered together with the bundled demo clips, and the server is asked for its voices whether or not the UI is in management mode.

## Validation

```
```

Live server, Metal, Apple M4 Max, request shapes matching what the UI sends before and after:

| Check | Old shape | New shape |
|---|---|---|
| MOSS-TTS-Local, same text and seed | `max_tokens: 1024` → 24.2 s of audio | no `max_tokens` → 29.4 s |
| DotTTS-SOAR, same text and seed | `max_tokens: 1024` → md5 `8a1064a9…` | no `max_tokens` → md5 `8a1064a9…` (cap inert for this input) |
| DotTTS-SOAR, seed reproducibility | seed 1234 twice → identical md5 `8a1064a9…` | seed 2864434397 → `7d100930…` |

Both MOSS renderings transcribe back to the full input text through Parakeet-TDT, so on that family the cap changed the realisation rather than truncating content. The point stands either way: the UI was imposing a limit the model never asked for, and the effect is model-dependent, which is exactly why the client should not pick the number.

## Scope

`+page.svelte` and `Arena.svelte`. No catalog, spec or engine change. The generated bundle is deliberately excluded — it is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict; happy to send one bundle-regeneration PR once the series lands.

### DIFF
--- a/webui/native/src/routes/+page.svelte
+++ b/webui/native/src/routes/+page.svelte
@@ -72,8 +72,12 @@
   let instructions = '';
   let lyrics = '';
   let duration = 30;
-  let seed = 1234;
-      let maxTokens = 1024;
+  // -1 means "let the engine pick", matching the field label. A fixed default
+  // here would silently pin every model whose own default is a random seed.
+  let seed = -1;
+  // Blank means "use the model's own limit". A shared constant here overrode
+  // per-model ceilings ranging from 300 to 4096, halving some and doubling others.
+  let maxTokens: number | '' = '';
       let sourceFile: File | null = null;
       let videoFile: File | null = null;
       let voiceFile: File | null = null;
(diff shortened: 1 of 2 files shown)


CryptVenture:

PR #412, opened 2026-09-02
webui: show the segments, speaker turns and word timings the server returns

CryptVenture:
Split out of #373 as requested. This PR is the transcript detail display only; the package resolver, the catalog entries, the request wiring and the UI text fixes are separate PRs.

## The problem

ASR, diarization, VAD and forced alignment already return timed detail — segments, speaker turns, word timings — and the result pane showed the flat transcript only. Word timings a user could have exported as subtitles were computed and thrown away on every run.

## The change

The result pane renders whichever detail array came back — speaker turns, word timings or segments, in that order of specificity — as a scrollable table of start, end and content, with **Save SRT** and **Save VTT** beside it.

Spans arrive as sample offsets, so they are divided by the `sample_rate` the response carries. A response without that field renders **no rows** rather than a table of zeros; timecodes are formatted once and shared by the table and both subtitle formats, which differ only in the millisecond separator and the `WEBVTT` header.

**Dependency:** rows appear once the server reports `sample_rate` alongside the detail arrays, which is the companion server PR in this series. Without it this PR is inert rather than wrong.

## Validation

```
```

The three functions were extracted and executed against a real Parakeet-TDT response (12 s clip, Metal, Apple M4 Max):

```
kind=words rows=24
SRT: 1 / 00:00:00,080 --> 00:00:00,533 / At3:45
VTT: WEBVTT /  / 00:00:00.080 --> 00:00:00.533
no sample_rate -> rows=0
```

## Scope

`+page.svelte` (state, three helpers, one markup block), nine English strings in `i18n.ts`, and the table rules in `app.css`. No request behaviour changes. The generated bundle is deliberately excluded — it is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict; happy to send one bundle-regeneration PR once the series lands.


CryptVenture:

PR #413, opened 2026-09-02
webui: fix two strings that rendered wrong

CryptVenture:
Split out of #373 as requested. Two string fixes, nothing else; the package resolver, the catalog entries, the request wiring and the transcript display are separate PRs.

## The changes

**Hero subtitle rendered its lookup key.** The workflow tabs are keyed `conversion` and `separation`, while `studio.subtitle.*` was defined only for `vc` and `sep`. The Voice conversion and Source separation tabs printed `studio.subtitle.conversion` and `studio.subtitle.separation` verbatim where their description should be. Adds the two missing keys.

**Double-encoded ellipsis in the models-folder status line.** `folderâ€¦` is what a UTF-8 ellipsis looks like after being decoded as Latin-1. Replaced with the character itself.

## Validation

```
```

## Scope

Two English strings in `i18n.ts`, one status line in `+page.svelte`. No behaviour change. The generated bundle is deliberately excluded — it is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict; happy to send one bundle-regeneration PR once the series lands.


CryptVenture:

PR #414, opened 2026-09-02
webui: drop parameter controls the engine never reads

CryptVenture:
Split out of #374 as requested: controls the engine does not read. The default overrides, the range fixes, the missing groups and the label text are separate PRs.

Every control in `model_params.json` is sent with the request, so a control keyed to something the engine never reads is a widget that cannot do anything — and in three cases the working spelling was missing entirely.

## Renamed to the key the engine actually reads

| Group | Was | Now | Code |
|---|---|---|---|
| `index_tts2` | `lang` | `language` | `src/models/index_tts2/request.cpp:90`, `session.cpp:370` read `language`; nothing reads `lang` |
| `miotts` | `best_of_n` | `miotts.best_of_n` | the prefixed key declared in `src/models/miotts/loader.cpp:33` |
| `seed_vc` | `*_cfg_rate` ×3 | `*_guidance_scale` ×3 | the old spellings survive only through the deprecation table in `src/models/seed_vc/session.cpp:83-85` |

`index_tts2` is the notable one: that is the multilingual selector, on the model whose headline feature is multilingual cloning, and it had never worked.

## Removed

- **`ace_step`'s `remix` route and its five controls** (`source_caption`, `source_lyrics`, `flow_edit_n_min/max/avg`). The engine defines its routes in `src/models/ace_step/session.cpp` and throws on anything else; `remix` appears nowhere in the tree.
- **`firered-audio-tts` `top_k`, `top_p`, `temperature`.** `src/models/firered_audio/session.cpp:174-182` reads them while building the *understanding* request, never on the generation path this group drives.
- **`echo_tts` `seed`.** The server overwrites `options["seed"]` from the top-level field (`app/server/runtime.cpp:1925`), so the control could not take effect.

## Validation

```bash
```

Each claim above was checked against the file and line named beside it. The families with no package installed here were not run end to end.

## Scope

One data file, 8 insertions / 17 deletions. No code change. The generated bundle is deliberately excluded — `catalog.ts` inlines this file at frontend build time and the bundle is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict; happy to send one bundle-regeneration PR once the series lands.

### DIFF
--- a/webui/configs/model_params.json
+++ b/webui/configs/model_params.json
@@ -5,8 +5,7 @@
     {"name": "speaker_guidance_scale", "type": "slider", "label": "speaker_guidance_scale", "label_en": "Speaker guidance", "default": 8.0, "minimum": 0.0, "maximum": 15.0, "step": 0.1},
     {"name": "truncation_factor", "type": "slider", "label": "truncation_factor", "label_en": "Noise truncation", "default": 0.8, "minimum": 0.0, "maximum": 1.0, "step": 0.05},
     {"name": "guidance_interval", "type": "slider", "label": "guidance_interval", "label_en": "Guidance interval", "default": 1, "minimum": 1, "maximum": 3, "step": 1, "precision": 0, "info": "Refresh the unconditional CFG lanes every Nth guided step. Higher is faster and works best with more steps; 1 is highest fidelity."},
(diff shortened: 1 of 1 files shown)


CryptVenture:

PR #415, opened 2026-09-02
webui: stop overriding engine and package defaults from model_params.json

CryptVenture:
Split out of #374 as requested: UI defaults that override the engine or the package. The dead controls, the range fixes, the missing groups and the label text are separate PRs.

## The problem

Every default in `model_params.json` is sent on every request. A value here is not a fallback the engine can improve on — it is an override that wins over the model's own default and over anything the package ships.

| Group / control | UI sent | Engine or package value | Code |
|---|---|---|---|
| `vibevoice.num_inference_steps` | 10 | `ddpm_num_inference_steps` from the package, 20 in the shipped GGUF | `src/models/vibevoice/assets.cpp:112` |
| `ace_step.shift` | 3.0 | 1.0 | `include/engine/models/ace_step/types.h:42`, applied at `request_parser.cpp:231` |
| `firered-audio-asr.max_new_tokens` | 512 | 300, and 1024 when `enable_thinking` is set | `src/models/firered_audio/session.cpp:162` |
| `personaplex.text_top_k` | 250 | follows `top_k` when unset | `src/models/personaplex/request.cpp:111` |
| `vevo2.top_k` | 20 | the checkpoint's own generation config | `src/models/vevo2/session.cpp:898` |
| `fireredtts3-instruct-vdesign.guidance_scale` | 2.0 (the instruct path's value) | 1.2 | `src/models/fireredtts3/session.cpp:159, 170, 179` |

`vibevoice` is the sharpest one: the UI silently halved diffusion quality on every request.

## The change

Four defaults are **omitted** rather than corrected, because the right value is not a constant — it is read per package or derived from another option, and any literal here would override it again. Omitting a default is how this file says *unset*: the option parser skips empty values, so the control renders blank and sends nothing until the user types something.

Two defaults are **corrected** to the engine's value (`ace_step.shift`, the voice-design `guidance_scale`).

## Validation

```bash
```

Each value was read from the source line named beside it. These families have no package installed on this machine, so they were not run end to end; a live server was used to confirm the empty-default behaviour renders and sends nothing.

## Scope

One data file, six controls. No code change. The generated bundle is deliberately excluded — `catalog.ts` inlines this file at frontend build time and the bundle is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict; happy to send one bundle-regeneration PR once the series lands.

### DIFF
--- a/webui/configs/model_params.json
+++ b/webui/configs/model_params.json
@@ -21,7 +21,7 @@
   ],
 
   "vibevoice": [
-    {"name": "num_inference_steps", "type": "number", "label": "num_inference_steps", "default": 10, "minimum": 1, "step": 1, "precision": 0, "info": "扩散步数（官方默认 10），越大越慢越稳"},
+    {"name": "num_inference_steps", "type": "number", "label": "num_inference_steps", "minimum": 1, "step": 1, "precision": 0, "info": "留空=用模型自带的 ddpm_num_inference_steps（随包，7B 为 20）", "info_en": "Blank uses the packaged model's own ddpm_num_inference_steps (20 in the shipped 7B). Setting a value overrides it for every request."},
     {"name": "guidance_scale", "type": "slider", "label": "guidance_scale", "default": 1.3, "minimum": 0.0, "maximum": 5.0, "step": 0.1, "info": "CFG 引导强度"},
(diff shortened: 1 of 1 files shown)


CryptVenture:

PR #416, opened 2026-09-02
webui: keep parameter ranges inside what the engine accepts

CryptVenture:
Split out of #374 as requested: ranges that let the UI produce hard errors. The dead controls, the default overrides, the missing groups and the label text are separate PRs.

## The problem

Nine minimums sat below the engine's own guard, so a legal-looking slider position produced a 500 instead of a clamped value.

| Group / control | Was | Now | Guard |
|---|---|---|---|
| `qwen3_tts.temperature` | min 0.0 | 0.05 | `src/models/qwen3_tts/session.cpp:68` — "temperature must be positive when sampling" |
| `confucius4_tts.temperature` | min 0.0 | 0.05 | `src/models/confucius4_tts/request.cpp:62` |
| `confucius4_tts.top_p` | min 0.0 | 0.01 | `src/models/confucius4_tts/request.cpp:65` — must be in (0, 1] |
| `personaplex.temperature`, `personaplex.text_temperature` | min 0.0 | 0.05 | positive-value parse |
| `vevo2.temperature` | min 0.0 | 0.05 | positive-value parse |
| `dramabox.reference_duration_sec` | min 0.0 | 0.5 | `src/models/dramabox/session.cpp:149` |
| `dramabox.audio_chunk_threshold_sec`, `audio_chunk_duration_sec` | min 0.0 | 1.0 | same derivation, frame counts round to zero |
| `controlfoley.guidance_scale` | min 0.0 | 0.1 | `src/models/controlfoley/pipeline.cpp:176` reads it through `positive_float_option`, which throws at zero |
| `controlfoley.duration_sec` | min 0.1, no max | 0.64 – 60.0 | the range `model_specs/controlfoley.json` declares; below the floor the derived clip/visual/sync frame counts round to zero |
| `midashenglm_gen.min_stop_step` | min 0 | 1 | `src/models/midashenglm_gen/session.cpp:175` parses it as a positive integer |

`echo_tts.num_inference_steps` had a **maximum equal to its own default**, so the control could only reduce quality; the ceiling moves to 120.

## Also here

Three controls lose their default as well as their range — `personaplex.text_temperature` (follows `temperature` when unset), `vevo2.temperature` and `vevo2.top_p` (read from the checkpoint's generation config). They become blank number inputs rather than sliders pinned to a literal, for the same reason as the sibling defaults PR.

Worth flagging separately: `model_specs/controlfoley.json` declares `guidance_scale` min 0.0 while the engine throws at zero. That is a spec bug, not a UI one, and belongs to the spec series rather than this PR.

## Validation

```bash
```

Every bound above was read from the guard named beside it.

## Scope

One data file, 14 controls. No code change. The generated bundle is deliberately excluded — `catalog.ts` inlines this file at frontend build time and the bundle is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict; happy to send one bundle-regeneration PR once the series lands.

### DIFF
--- a/webui/configs/model_params.json
+++ b/webui/configs/model_params.json
@@ -1,6 +1,6 @@
 {
   "echo_tts": [
-    {"name": "num_inference_steps", "type": "slider", "label": "num_inference_steps", "label_en": "Sampling steps", "default": 40, "minimum": 8, "maximum": 40, "step": 1, "precision": 0, "info": "Euler sampler steps."},
+    {"name": "num_inference_steps", "type": "slider", "label": "num_inference_steps", "label_en": "Sampling steps", "default": 40, "minimum": 8, "maximum": 120, "step": 1, "precision": 0, "info": "Euler sampler steps.", "info_en": "Euler sampler steps. No upper bound in the model; higher is slower and steadier."},
     {"name": "text_guidance_scale", "type": "slider", "label": "text_guidance_scale", "label_en": "Text guidance", "default": 3.0, "minimum": 0.0, "maximum": 10.0, "step": 0.1},
     {"name": "speaker_guidance_scale", "type": "slider", "label": "speaker_guidance_scale", "label_en": "Speaker guidance", "default": 8.0, "minimum": 0.0, "maximum": 15.0, "step": 0.1},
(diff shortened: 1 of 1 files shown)


CryptVenture:

PR #417, opened 2026-09-02
webui: add parameter groups for the ASR, VAD, diarization and alignment models

CryptVenture:
Split out of #374 as requested: missing parameter groups, analysis families. The TTS and music groups, the controls added to existing groups, the dead controls, the default overrides and the range fixes are separate PRs.

## The problem

Fourteen ASR, VAD, diarization and forced-alignment families had **no group at all** in `model_params.json`, so their entire option surface was reachable only by hand-writing JSON into the fallback box. That includes `silero_vad`'s `threshold` and min-speech/min-silence knobs — which are the point of that model — and `kroko_asr`'s beam search and hotword biasing.

## The change

Groups added: `fun_asr_nano`, `granite5asr`, `higgs_audio_stt`, `hviske_asr`, `kroko_asr`, `marblenet_vad`, `nemotron_asr`, `parakeet_tdt`, `qwen3_asr`, `qwen3_forced_aligner`, `silero_vad`, `sortformer_diar`, `vibevoice_asr`, `voxtral_realtime` — 57 controls in total.

Every control is named after the key its family's session actually reads. The shared `audio_chunk_mode` and `audio_chunk_duration_sec` controls are read through `src/framework/audio/chunking.cpp:167` and `:190`, which each of these families calls; the rest resolve inside the family's own sources.

No existing group is touched, so nothing that worked before changes.

## Validation

```bash
```

Every one of the 57 control names was checked against its family's sources — either a literal `"<name>"` in the family directory, or the shared chunking helper above.

## Scope

One data file, 85 added lines, no deletions. No code change. The generated bundle is deliberately excluded — `catalog.ts` inlines this file at frontend build time and the bundle is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict; happy to send one bundle-regeneration PR once the series lands.

### DIFF
--- a/webui/configs/model_params.json
+++ b/webui/configs/model_params.json
@@ -371,5 +371,90 @@
     {"name": "top_p", "type": "slider", "label": "top_p", "default": 0.9, "minimum": 0.0, "maximum": 1.0, "step": 0.01},
     {"name": "top_k", "type": "number", "label": "top_k", "default": 50, "minimum": 0, "step": 1, "precision": 0},
     {"name": "max_tokens", "type": "number", "label": "max_tokens", "default": 1024, "minimum": 1, "step": 1, "precision": 0}
+  ],
+  "fun_asr_nano": [
+    {"name": "enable_itn", "type": "bool", "label": "enable_itn", "label_en": "Inverse text normalization", "default": true},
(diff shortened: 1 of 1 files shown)


CryptVenture:

PR #418, opened 2026-09-02
webui: add parameter groups for the TTS and music models that had none

CryptVenture:
Split out of #374 as requested: missing parameter groups, generative families. The analysis groups, the controls added to existing groups, the dead controls, the default overrides and the range fixes are separate PRs.

## The problem

Eight TTS and music families had **no group at all** in `model_params.json`. The entire control surface of `dots_tts`, `outetts`, `glm_tts` and `fish_audio` — sampling, guidance, chunking, output format — was reachable only by hand-writing JSON into the fallback box.

## The change

Groups added: `dots_tts`, `fish_audio`, `glm_tts`, `higgs_audio_tts`, `muscriptor`, `outetts`, `soprano_tts`, `vietneu_tts` — 61 controls.

Every control is named after the key its family reads. `text_chunk_size` and `text_chunk_mode` go through the shared overrides in `src/framework/text/chunking.cpp:552` and `:565`, which `dots_tts` (`request.cpp:85-87`) and `outetts` (`session.cpp:749`) both call.

No existing group is touched, so nothing that worked before changes.

## Validation

```bash
```

Every one of the 61 control names was checked against its family's sources, or against the shared chunking helper above.

## Scope

One data file, 77 added lines, no deletions. No code change. The generated bundle is deliberately excluded — `catalog.ts` inlines this file at frontend build time and the bundle is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict; happy to send one bundle-regeneration PR once the series lands.

### DIFF
--- a/webui/configs/model_params.json
+++ b/webui/configs/model_params.json
@@ -371,5 +371,82 @@
     {"name": "top_p", "type": "slider", "label": "top_p", "default": 0.9, "minimum": 0.0, "maximum": 1.0, "step": 0.01},
     {"name": "top_k", "type": "number", "label": "top_k", "default": 50, "minimum": 0, "step": 1, "precision": 0},
     {"name": "max_tokens", "type": "number", "label": "max_tokens", "default": 1024, "minimum": 1, "step": 1, "precision": 0}
+  ],
+  "dots_tts": [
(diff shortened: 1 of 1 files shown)


CryptVenture:

PR #420, opened 2026-09-02
webui: give the parameter controls English labels and accurate help text

CryptVenture:
Split out of #374 as requested: label and help text only. The dead controls, the default overrides, the range fixes and the missing groups are separate PRs.

## The change

- **42 controls gained `label_en` / `info_en` / `placeholder_en`.** They carried a Chinese label with no English counterpart, so the English UI fell back to printing the raw option key.
- **Several `info` strings described behaviour the engine does not have** and now describe what it does.
- **The file's own `_comment` claimed only user-modified values are sent.** That has never been true in this UI — every control with a default is sent on every request — so the comment now documents the real behaviour, including that omitting a default is how a control says *unset* (the option parser skips empty values).

No `name`, `type`, `default`, `minimum`, `maximum`, `step` or `choices` value changes here, so no request shape changes.

## Validation

```bash
```

## Scope

One data file, 43 lines. No behaviour change of any kind. The generated bundle is deliberately excluded — `catalog.ts` inlines this file at frontend build time and the bundle is not byte-reproducible, so regenerating it in each PR of this split would make the PRs conflict; happy to send one bundle-regeneration PR once the series lands.


CryptVenture:

PR #421, opened 2026-09-02
webui: complete the shipped translations

CryptVenture:
Replaces #375, reduced to translations as requested: this PR touches `webui/native/lang/*.json` and nothing else — no specs, no catalog, no UI behaviour, no generated bundle.

## The problem

Seven keys were missing from all four translation files, so the localised UI was visibly half-English:

- `file.clear` and `file.preview` sit on **every file input in the app**
- `task.midi` labelled MuScriptor in English for every non-English user
- `request.autoDuration`, `request.rewriteCaption`, `request.rewritingCaption`, `voice.configured`

`param.*` coverage stood at roughly ten keys per language out of nearly three hundred, so the parameter panels were English for everyone.

## The change

| File | Keys before | After |
|---|---|---|
| `lang_zh.json` | 255 | 366 |
| `lang_it.json` | 255 | 294 |
| `lang_pl.json` | 255 | 294 |
| `lang_ru.json` | 255 | 294 |

Chinese gains 111 keys back-filled from the Chinese `label`, `info` and `placeholder` text **already in `model_params.json`**, which the build discards because `catalog.ts` strips Han characters — that wording was in the repository but unreachable even for Chinese users. Italian, Polish and Russian gain 39 each, limited to entries whose meaning is unambiguous from the English label and the option's own documentation.

The two `studio.subtitle.*` keys are translated here; their English source keys are added by the sibling string-fix PR, and until that lands these entries are simply unused.

## Validation

```bash
python3 -m json.tool webui/native/lang/lang_{zh,it,pl,ru}.json   # all parse
```

Every non-`param.*` key was checked against the English table in `src/lib/i18n.ts`: the only keys not present there are the two `studio.subtitle.*` above.

## Scope

Four JSON files. No code, spec, catalog or bundle change.


CryptVenture:

PR #439, opened 2026-09-04
Enhance FireRedTTS3 with streaming, batching, and server integration

duj12:


0xShug0:
Hi @duj12 Thank you for the PR. Please split the PR into smaller, single-responsibility PRs to kepp PRs focused, and explain what a PR does in detail. We recently added a policy limiting each user to 3 concurrent PRs, including draft PRs. Please keep only keep three PRs you think I should review first.

When changing shared framework code, explain which model families or routes were checked. Please provide runtime regression tests on the affected models (outputs, performance, memory, etc). Static analysis or reasoning is not enough. 

One important principle is that framework-level code shouldn’t be shaped just to benefit a single model. I’d suggest keeping the changes scoped to the model directory for now, and we can decide later which parts are general enough to be promoted into the framework.

0xShug0:
@duj12 I’m going to close this PR. Please check my comment above if you’d like to open a new PR with the requested changes.

PR #441, opened 2026-09-04
fix(text-chunking): split Default-mode CJK text at full-width punctuation

gqf2008:
## Summary

`split_word_ranges` in the Default text-chunking mode cut words only at ASCII spaces. CJK text has no inter-word spaces, so an entire paragraph parsed as a **single word**; once the text exceeded the codepoint budget (`text_chunk_size`, default 200) the hard-split fallback kept the whole paragraph as one chunk and the budget was never applied.

Every TTS family that goes through `chunk_text_request` (omnivoice, outetts, heartmula, moss_voicegen, audio8_tts, ...) is affected, in both offline and streaming runs: long Chinese text synthesized as one oversized chunk regardless of `text_chunk_size`. `text_chunk_mode=tag_aware` was the only workaround.

Fix: full-width CJK sentence/clause punctuation (。！？，、；：) now forms word boundaries — the punctuation token becomes a word of its own carrying its sentence/clause break flags, so the existing greedy budget + rollback logic splits at natural boundaries. ASCII punctuation stays attached to its word, leaving Latin text byte-identical.

## Verification

New `text_chunking_test` (5 cases: CJK over-budget splits at sentence boundary, under-budget stays whole, clause-punctuation rollback, Latin sentence split unchanged, oversized single-word passthrough). Confirmed on macOS arm64 and across the full CI matrix (Linux cpu/vulkan, macOS, Windows):

- 236-codepoint CJK paragraph with the default 200 budget: `audio8_tts.text_chunk_count` 1 → 2
- 115-codepoint short text: output wav byte-identical to the pre-fix build
- Latin text: chunks identical to pre-fix

```bash
cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DENGINE_BUILD_TESTS=ON
cmake --build build --target text_chunking_test
./build/text_chunking_test   # text_chunking_test: all passed
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2110,6 +2110,13 @@ if (ENGINE_BUILD_TESTS)
         COMMAND chinese_normalization_test
     )
 
+    add_engine_unittest(text_chunking_test tests/unittests/test_text_chunking.cpp)
+
+    add_test(
+        NAME text_chunking_test
+        COMMAND text_chunking_test
+    )
+
     add_engine_unittest(unicode_normalization_test tests/unittests/test_unicode_normalization.cpp)
 
     add_test(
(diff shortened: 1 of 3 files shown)


0xShug0:
@gqf2008 Merged. Thanks!

PR #333, opened 2026-08-29
feat: Supporting Audio8_TTS models

jasonchen31:
## Summary

Native `audio.cpp` port of **Audio8 TTS Preview** as community family `audio8_tts` (DualAR: slow semantic → fast codebook → 44.1 kHz codec), derived from Fish Audio S2 Pro patterns. Ships as standalone GGUFs with no Python dependency.

- **0.6B Qwen** (`slow_backbone=qwen`, 24L, dim 896, 14H/2KV, RoPE 1e6, packed QKV+bias) — **fully native and CPU-verified** via SenseVoice ASR round-trip.
- **0.1B Falcon-H1** (`slow_backbone=falcon_h1`, dim 512, `d_inner 768=32×24`, `d_state 64`, `d_conv 4`, `dt_rank 24`, `n_group 1`, `GQA 8/2`, RoPE 1e11, `embedding_multiplier 0.1088`/`lm_head 0.0781`) — weight-complete and builds natively (GGUF `slow.embed_tokens` + `24× mamba/attention` + `semantic_output 4097`), slow AR currently a **documented stub** pending full Mamba2 port (see `docs/FALCON_H1_0.1B_PORT_PLAN.md`).

All 0.6B generation paths use ggml-native graphs; 0.1B not yet supported.

---

### 1. Community family `audio8_tts`
- **Spec:** `model_specs/audio8_tts.json` (schema-v1, languages `yue/zh/nl/en/fr/de/it/ja/ko/pl/es/auto`, tasks `tts`/`clon`, packages `audio8_tts_preview_0_6b_q8_0`/`audio8_tts_preview_0_1b_q8_0`)
- **Docs:** `docs/community_models/audio8_tts.md` (usage, architecture, GGUF packaging, validation notes), `docs/community_models/models.md` update, `docs/FALCON_H1_0.1B_PORT_PLAN.md` (M0–M4, 3–3.5d plan)
- **Tools:** `tools/community_models/convert_audio8_tts_codec.py` (zipfile/pickle `codec.pth` → `codec.safetensors`, no torch) + `tools/community_models/convert_audio8_tts.py` (snapshot → standalone GGUF, embedded config/tokenizer/spec)

### 2. Runtime (`src/community_models/audio8_tts/`)
- `types.h` / `assets.cpp` — flat ArkTTS config parsing including Falcon multipliers (embedding/lm_head/ssm/attn/mlp)
[...]

PR #444, opened 2026-09-04
feat(audio8_tts): Falcon-H1 0.1B (Mamba2 + attention) support and codec/AR performance

gqf2008:
## Summary

Companion to #333 (Audio8-TTS 0.6B): native support for the **Audio8-TTS-Preview-0.1B** checkpoint, whose AR stage is a Falcon-H1 hybrid — stateful **Mamba2 + grouped-query attention** per token — instead of the 0.6B's pure transformer.

### Port

- `falcon_forward_step`: RMSNorm → (Mamba2 ‖ GQA attention) → residual → gated FFN, stateful per token: conv/SSM states + KV cache. Mamba2 path: `in_proj[z,xBC,dt]` split → `ssm_conv` (kernel flipped to match HF causal_conv1d) → `ssm_scan` → D → silu(z) gate → `out_proj`. Attention: q/k/v → RoPE NEOX (base 1e11) → flash attention with KV cache.
- Multiplier semantics fixed against the reference: `lm_head_multiplier` does not apply to the compact semantic head; `embedding_multiplier` applies to (text_emb + codebook_sum) jointly.
- Decisive correctness bug: the host KV cache used a seq-scaled head stride while appending only the new token, so from the second step on, every head > 0 read corrupted K/V. Fixed in `falcon_kv_cache.h` + regression test.

### Performance (M4, Metal, q8_0)

- 0.1B session RTF **~1.3 → ~0.34** (19s audio in ~5.9s excluding load), byte-identical output across runs (seed-pinned).
- AR: per-token graph moved to a dedicated CPU backend (600-node graphs were dispatch-bound on Metal), then zero-copy state/constants, padded host KV buffers with in-graph slot writes, and per-capacity-bucket graph reuse (~781ms/generation compute, ~2ms overhead).
- codec: stride-1 conv1d Metal fast path on time-fast layouts, decoder blocks channel-fast (saves 46 transposes + 26 bias repeats per pass), new `GGML_OP_MUL_MAT_ACC` (accumulate-in-place per-tap GEMMs), and a fused `GGML_OP_SNAKE_1D` op replacing the 5-kernel/11-pass activation chain (codec compute −21.7%). 0.6B codec decode 4.8s → ~2.25s at the first step of this series; all gated work paths kept byte-identical output.
- ggml: `ssm_scan` reduction read garbage when `sgptg < NW` (Metal); `ggml-metal.metal` CRLF renormalized first for a clean diff.

## Verification

- Logits argmax parity vs transformers fp32 `trust_remote_code` reference (first-frame argmatch; bf16-GGUF noise stays below top-2 gap except known ties).
- ASR round-trip (Qwen3-ASR): synthesized "你好" → "你好。"; 605-position long run: SSM state peak ~1.1e3, zero NaN, on both CPU and Metal backends.
- Every perf step verified output byte-identical (or env-gated fallback identical) to the prior step; regression test `audio8_tts_falcon_kv_cache_test` covers the KV stride bug.
- Status doc: `docs/community_models/audio8_tts_falcon_h1_status.md` (root causes for the three mid-port misdiagnoses included).

### DIFF
--- a/src/community_models/audio8_tts/ar.cpp
+++ b/src/community_models/audio8_tts/ar.cpp
@@ -103,6 +103,7 @@ struct FalconH1LayerWeights {
     assets::TensorDataF32 input_layernorm;          // slow.layers.*.input_layernorm.weight [512]
     core::TensorValue ssm_in;                        // slow.layers.*.mamba.in_proj.weight [1688,512]
     core::TensorValue ssm_conv1d;                    // slow.layers.*.mamba.conv1d.weight [896,1,4] -> [4,896] after convert
+    assets::TensorDataF32 conv1d_flipped;            // host [4,896] kernel-flipped for ggml ssm_conv
     assets::TensorDataF32 ssm_conv1d_b;              // slow.layers.*.mamba.conv1d.bias [896]
     core::TensorValue ssm_dt_b;                      // slow.layers.*.mamba.dt_bias [24]
     core::TensorValue ssm_A;                         // slow.layers.*.mamba.A_log [24] -> [1,24]
@@ -425,27 +426,57 @@ FalconH1LayerWeights load_falcon_layer(
     // Shapes reflect HF safetensors (safetensors) and GGUF (after convert) – use actual metadata shape to stay compatible.
     FalconH1LayerWeights w;
     w.input_layernorm = source.require_f32_tensor(prefix + ".input_layernorm.weight", {text_config.dim});
+    {
+        auto dbg_w = source.require_f32_tensor(prefix + ".input_layernorm.weight");
+        FILE * f = fopen("/tmp/falcon_dbg.txt", "a");
+        if (f) {
+            fprintf(f, "[W] prefix=%s ln0=%.4f,%.4f,%.4f\n", prefix.c_str(),
+                    (double)dbg_w.values[0], (double)dbg_w.values[1], (double)dbg_w.values[2]);
+            fclose(f);
+        }
+    }
     // Mamba in_proj: HF [1688,512] (out, in), GGUF may be transposed; load with actual shape
     {
         auto meta = source.require_metadata(prefix + ".mamba.in_proj.weight");
         w.ssm_in = store.load_tensor(source, prefix + ".mamba.in_proj.weight", storage_type, meta.shape);
     }
     {
+        // ssm_conv Metal pipeline requires contiguous F32 conv weights.
         auto meta = source.require_metadata(prefix + ".mamba.conv1d.weight");
-        w.ssm_conv1d = store.load_tensor(source, prefix + ".mamba.conv1d.weight", storage_type, meta.shape);
+        w.ssm_conv1d = store.load_tensor(source, prefix + ".mamba.conv1d.weight", assets::TensorStorageType::F32, meta.shape);
+        {
+            // ggml ssm_conv computes y[t] = sum_k w[k]*x[t+k]; HF causal_conv1d uses
+            // w[k]*x[t+d_conv-1-k]. Flip the kernel dim once at load time so the
+            // per-step graph can feed the flipped weight directly.
+            auto raw = source.require_f32_tensor(prefix + ".mamba.conv1d.weight");
+            const int64_t d_conv = raw.shape.dims.size() >= 3 ? raw.shape.dims[2] : (raw.shape.dims.empty() ? 4 : raw.shape.dims[0]);
+            const int64_t conv_dim = raw.shape.dims.size() >= 3 ? raw.shape.dims[0] : (raw.shape.dims.empty() ? 0 : raw.shape.dims[0]);
+            // raw layout [conv_dim, 1, d_conv] col-major: (c, g, k) at c + conv_dim*(g + k)
+            std::vector<float> flipped(static_cast<size_t>(conv_dim * d_conv));
+            for (int64_t k = 0; k < d_conv; ++k) {
+                for (int64_t c = 0; c < conv_dim; ++c) {
+                    flipped[static_cast<size_t>(k + d_conv * c)] =
+                        raw.values[static_cast<size_t>(c + conv_dim * (d_conv - 1 - k))];
+                }
+            }
+            w.conv1d_flipped.shape = core::TensorShape::from_dims({d_conv, conv_dim});
+            w.conv1d_flipped.values = std::move(flipped);
+        }
     }
     w.ssm_conv1d_b = source.require_f32_tensor(prefix + ".mamba.conv1d.bias");
+    // Per-head Mamba params must stay unquantized (Native): they are consumed as
+    // raw F32 scalars by the SSM path (A = -exp(A_log), D, dt bias).
     {
         auto meta = source.require_metadata(prefix + ".mamba.dt_bias");
-        w.ssm_dt_b = store.load_tensor(source, prefix + ".mamba.dt_bias", storage_type, meta.shape);
+        w.ssm_dt_b = store.load_tensor(source, prefix + ".mamba.dt_bias", assets::TensorStorageType::F32, meta.shape);
     }
     {
         auto meta = source.require_metadata(prefix + ".mamba.A_log");
-        w.ssm_A = store.load_tensor(source, prefix + ".mamba.A_log", storage_type, meta.shape);
+        w.ssm_A = store.load_tensor(source, prefix + ".mamba.A_log", assets::TensorStorageType::F32, meta.shape);
     }
     {
         auto meta = source.require_metadata(prefix + ".mamba.D");
-        w.ssm_D = store.load_tensor(source, prefix + ".mamba.D", storage_type, meta.shape);
+        w.ssm_D = store.load_tensor(source, prefix + ".mamba.D", assets::TensorStorageType::F32, meta.shape);
     }
     {
         auto meta = source.require_metadata(prefix + ".mamba.out_proj.weight");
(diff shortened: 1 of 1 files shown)


0xShug0:
@gqf2008 I will ping you when I'm ready to review this PR. There will be some GGML changes and I'm still testing them, which may cause conflicts with this PR.

gqf2008:
Updated the branch to drop a stray line-ending renormalization that was inflating this diff.

- The PR showed **+24934/-22329**; the real change is **+2733/-128**. The extra ~22k lines came from one `chore: renormalize CRLF -> LF` commit plus the merge that followed it, which together rewrote `external/ggml/include/ggml.h`, `external/ggml/src/ggml.c` and `external/ggml/src/ggml-metal/ggml-metal.metal` end to end.
- That commit is gone, and those three files now keep the same CRLF/LF layout as `main`, so the diff contains only the actual work (Falcon-H1 forward port, `GGML_OP_SNAKE_1D`, conv1d per-tap GEMM accumulation, Metal stride-1 conv1d fast path, ssm_scan reduction fix).
- **Content is unchanged**: `git diff --ignore-cr-at-eol <new head> <old head>` is empty — the only difference is line endings. Messages, authors and dates are preserved (17 commits, was 18).

On your 2026-09-11 note about GGML changes conflicting here: the only `external/ggml` change on main since then is #508 (Vulkan fill dispatch — `ggml-vulkan.cpp` + `fill.comp`), which does not overlap this PR's `ggml-metal/*`, `ggml.h` or `ggml.c`. `git merge-tree` against current `main` (024297b3) merges cleanly as of today.

One thing worth knowing, since it bit me here: `.gitattributes` declares `* text=auto eol=lf`, but these vendored files are CRLF/LF-mixed in every blob on main — the most recent commit touching `ggml.h` (62a9441f) still stores 2811 CRLF lines, so `git add` does not renormalize them. Editing one of these files therefore rewrites the whole file unless the endings are preserved explicitly. That's why I restored the mixed layout instead of normalizing; if you'd rather have them LF everywhere, that's a one-time `git add --renormalize` on main and I'll keep this PR out of it.

Still based on a10738ad (behind main, merges clean). Say the word if you'd prefer a rebase onto current main instead.

0xShug0:
@gqf2008 #532 and #533 haven’t been merged yet...I've decided to merge them after the next stable release that includes YuE2, possibly tomorrow, to avoid introducing regressions right before the release. I will ping you.

Don’t worry about LF normalizing. I've given up on that :)

gqf2008:
Merged current `main` (82b4dc3a) into the branch to clear the conflict.

The conflict was the GGML op-enum tail: `main` added `CONV_3D_CONCAT_PAD_SPATIAL_GEMM`, `RMS_NORM_CHANNELS(_SILU)`, `RMS_NORM_CHANNELS_ADD_BIAS_SILU` and `ROPE_INTERLEAVED_PAIRS` (`GGML_OP_COUNT == 112`) while this branch appends `MUL_MAT_ACC` and `SNAKE_1D`. Kept `main`'s op order stable and appended the two audio8_tts ops at the tail, with `GGML_OP_COUNT` and both name/symbol tables in sync at 114.

Verification on this merge (macOS 26.5.2, Apple M4, Metal):
- CMake RelWithDebInfo build with `ENGINE_ENABLE_METAL=ON` + `ENGINE_BUILD_TESTS=ON`: 958/958 targets built.
- `ctest`: 35/35 passed.
- End-to-end with the merged tree: `audio8_tts` 0.1B synthesizes "你好，这是一个测试。" and `audio8_asr` 0.1B transcribes it back as "你好，这是一个。" The Metal pipeline log shows `kernel_mul_mm_acc_f32_f32(_bci)` and `kernel_snake_1d_f32` being compiled and dispatched, so the new ops are live on Metal.

The PR diff against `main` is now clean again: 18 files, +2733/-128, no line-ending noise.

gqf2008:
Status ping, plus a review guide in case it helps now that #532/#533 have landed.

**Where this stands**

- `main` (82b4dc3a, then the newer merges) is merged in and the op-enum conflict is resolved; `mergeable: clean`, CI green on all six jobs (Linux cpu/vulkan, Nix cpu/vulkan, Windows, macOS).
- Diff is **+2733/-128 across 18 files**. The "+24934/-22329" you saw earlier was the line-ending commit I dropped on 09-15; that is gone.
- You said the GGML changes you were testing (#532/#533) might conflict with this — they landed on 09-17, and that is exactly the conflict I resolved (kept main's op order, appended the two audio8_tts ops at the tail, `GGML_OP_COUNT` and both name/symbol tables in sync at 114).

**What actually needs eyes**, in order of how much attention I think it deserves. Most of the diff is mechanical:

1. `src/community_models/audio8_tts/ar.cpp` (+1334/-115) — the real content: the Falcon-H1 (Mamba2 + hybrid GQA) stateful single-token forward, `FalconH1StepState`, and the host KV-cache append with per-head re-stride. `docs/community_models/audio8_tts_falcon_h1_status.md` writes up the three root causes we chased (conv tap orientation, unprotected host read-back of graph intermediates, KV head-stride) and how each was verified.
2. `ggml-metal.metal` (+402/-1) — almost all of it mirrors existing kernel patterns for the two new ops, except **one line of it is a bug fix that stands on its own**: `kernel_ssm_scan_f32`'s old `simd_sum(shared_sums[sgitg*NW + tiisg])` reads garbage columns whenever `sgptg < NW` (which is what `d_state=64, n_t=1` hits). That fix is independent of this model and probably worth reviewing separately.
3. GGML ops — `MUL_MAT_ACC` and `SNAKE_1D` appended at the tail of the enum (main's order untouched), CPU impls in `ggml-cpu.c` (+105). Mechanical; the enum tables are the only fiddly part.
4. `src/framework/modules/conv_modules.cpp` (+183/-1) — conv1d/depthwise primitives plus a stride-1 fast path used by the codec.
5. The rest is a 116-line status doc, one doc line, and `tests/unittests/test_audio8_tts_falcon_kv_cache.cpp` (regression test for the KV head-stride bug).

**What we could not check**: CUDA and HIP. Our evidence is macOS/Metal (full build + 35/35 `ctest` + an audio8_tts -> audio8_asr round trip) and the CPU paths. Both the new ops and the `ssm_scan` fix touch backends you have and we don't, so if you want them exercised on real hardware, tell me what to run on Metal and I will follow the same protocol; the CUDA/HIP half needs a card.

**If it is too big as one piece**, say so and I will split it: (A) GGML layer — two ops + the `ssm_scan` fix + the CPU test, roughly 900 lines; (B) model + framework + docs, roughly 1900. I kept them together only because the model needs the ops, so either shape works for me.

No rush if you are still working through the other merges — this is here so the context is ready when you get to it.

0xShug0:
@gqf2008 The `Conv1dModule` change seems fairly broad. Could you test a few other (popular) models that are affected by this change, and check for any performance changes or output regressions?  Also add a unit test for the Conv1dModule fast path.

Update: the other parts are scoped. The Conv1dModule change also look correct and should benefit many models on Metal! Just a few extra metrics and we’re good to go.

gqf2008:
Did both things, plus two fixes that fell out of it (`eac4934f`, `9326555b`).

## 1. How broad is the `Conv1dModule` change — survey with a new trace

The per-tap path now writes a trace line when it is taken, so `--log` shows which models use it. Ran every model we have weights for on Metal:

| model | per-tap convs built | note |
|---|---:|---|
| `breeze_tts` (Mimi codec) | 108 | the most affected in this set |
| `index_tts2` | 76 | vocoder/codec convs |
| `audio8_tts` | 8 | where the path came from |
| `qwen3_asr` | 0 | its encoder convs are strided/padded, so ineligible |

Gate is `padding == 0 && stride == 1 && F32 && batch == 1`, so models whose codecs pad their convolutions keep the old path.

## 2. Output regression check (before = `main`, after = this branch)

The per-tap decomposition accumulates in a different order than im2col, so bit-identical audio is not expected; the unit test bounds the conv-level difference at 1e-4 relative. End-to-end, ASR round-trip of both arms for the same text and seed:

| model | before | after |
|---|---|---|
| `breeze_tts` | 今天的天气不错。我们下午三点在楼下的咖啡馆见面吧。 | same |
| `audio8_tts` | 今天的天气不错，我们下午三点在楼下的咖啡馆见面吧。 | same words, one punctuation difference |
| `index_tts2` | 今天的天气不错。我们下午三点在楼下的咖啡馆见面吧。 | same words, one punctuation difference |

## 3. Performance

End-to-end RTF on this fanless M4 Air is noise-dominated (two runs of the *same* arm differ by up to 30%), so I measured the op itself on the same backend (Metal, median of three, 96 channels, k5):

| shape | reference (im2col) | per-tap | speedup |
|---|---:|---:|---:|
| 4k frames | 6.88 ms | 2.05 ms | **3.4x** |
| 16k frames | 10.08 ms | 2.68 ms | **3.8x** |

That report is printed by the unit test, so it is reproducible with `-DENGINE_BUILD_EXTENDED_TESTS=ON` (same gate as the other conv fast-path tests).

## 4. Unit test for the fast path

`tests/unittests/test_conv1d_pertap_fast_path.cpp` builds the same module twice — Metal build context (per-tap) and default context (reference `ggml_conv_1d`) — runs both and checks:

- six shapes (kernels 3/5/7, dilation 2, with/without bias, a single output frame) agree to 1e-4 relative;
- graph shape proves which path was built: fast → per-tap matmuls and **no** im2col node, reference → im2col;
- ineligible configs (`stride != 1`, `padding != 0`) stay on the reference path;
- the exported channel-fast helper matches the canonical module path (this is the one the audio8_tts codec chains);
- the Metal timing report above.

## 5. Two fixes found on the way

- `test_audio8_tts_falcon_kv_cache.cpp` was **never registered in `CMakeLists.txt`**, so it had never been compiled or run. It is registered now (gated on the model, with the model's include dir) and passes.
- Rebased onto current `main`; no conflicts (the op-enum tail merge from before still holds).

Happy to reshape any of this — e.g. make the fast path opt-in like the breeze bf16 knob, or move the trace behind a separate flag.

gqf2008:
Two clarifications that came out of reviewing this myself, in case they save you a pass:

**1. The change is Metal-gated by construction, so the other backends cannot regress.** `is_conv1d_pertap_fast_path_eligible()` starts with `ctx.backend_type == core::BackendType::Metal` (plus `padding == 0 && stride == 1 && batch == 1 && F32`), and the only edit on the non-eligible path is the added trace call. On CUDA/HIP/CPU the module therefore builds exactly the graph it built before this PR — which is also why I could not produce CUDA/HIP numbers: there is nothing new to exercise there, only the Metal branch changes. The Metal branch itself is covered by the new unit test, which runs the same graph shape the models build.

**2. What is still unverified, explicitly:**

- end-to-end RTF on this machine is dominated by fanless-M4 noise (two runs of the same arm differ by up to 30%), so the performance claim is the per-conv number from the test report (3.4x / 3.8x on the codec shapes), not an end-to-end RTF;
- audio is not bit-identical between the paths (im2col vs per-tap accumulation order); the conv-level bound is 1e-4 relative and the end-to-end check is ASR round-trip equivalence (same words, punctuation jitter) — no listening test;
- the fast path has no opt-in switch. Unlike the breeze bf16 change, this one only ever replaces a slower Metal path with a faster one, so I did not add a flag; say the word if you want one anyway;
- the new test is registered in the `ENGINE_BUILD_EXTENDED_TESTS` block, next to the other conv fast-path tests, so it does not run in the linux/mac build workflows (they only set `ENGINE_BUILD_TESTS`). Happy to move it into the CI-gated block if you would rather have it run on every PR — it is platform-safe (the Metal timing section skips itself when Metal is unavailable).

gqf2008:
Moved the new test out of the `ENGINE_BUILD_EXTENDED_TESTS` block (`14ac9783`), so it now runs in the linux/mac build workflows instead of only locally with that switch on. It sits with the other unit tests; the comment next to it says why it deviates from the conv tests beside it.

Verified with a CI-shaped configure — `-DENGINE_BUILD_TESTS=ON`, `-DENGINE_ENABLE_METAL=OFF`, no extended tests, which is what `mac-build.yml` does:

```console
$ ctest -R conv1d_pertap_fast_path_test --output-on-failure
    Start 3: conv1d_pertap_fast_path_test
1/1 Test #3: conv1d_pertap_fast_path_test .....   Passed    0.01 sec
```

The timing section prints `skip: Metal backend not available` on that runner, so the CPU-only macOS job stays unaffected; on a Metal machine it records the per-conv comparison reported earlier (3.4x / 3.8x).

0xShug0:
@gqf2008 Could you fix the Linux CPU and Vulkan issues? I'll do a few more tests, and then we're good to go!

0xShug0:
A few more models that benefit from the fast path:

Model | Audio len | RTF before | RTF after | RTF change | Hot component change
-- | -- | -- | -- | -- | --
qwen3_tts | 10.16s | 0.737 | 0.501 | -32.0% | speech decoder -75.4%
irodori_tts | 17.28s | 0.712 | 0.566 | -20.5% | codec decode -36.5%
voxcpm2 | 10.24s | 0.315 | 0.243 | -22.9% | AudioVAE decoder -63.8%

One issue may be worth fixing: the logs are a bit spammy (qwen3 tts have 216 markers in the test). Since the performance boost is confirmed, it’s safe to remove these logs.

gqf2008:
Both fixed, thanks for the pointer.

**Linux cpu/vulkan build failure** — the test used `std::sort` without `<algorithm>`. libc++ (macOS, Windows/MSVC, Nix) pulls that header in transitively, libstdc++ does not, which is why only the GCC jobs broke. Added `<algorithm>` (and `<cstddef>`), and audited every `std::` symbol in the file against its header so nothing else is implicit.

**Log spam** — removed the four `trace_log_scalar` calls and the now-unused include. The fast path itself is untouched; the model survey numbers I posted earlier (breeze 108 / index 76 / audio8 8 / qwen3-asr 0) were collected while the trace was still there.

Verified locally with the CI shape (`-DENGINE_BUILD_TESTS=ON`, `-DENGINE_ENABLE_METAL=OFF`, no extended tests):

```console
$ ctest -R "conv1d_pertap_fast_path_test|audio8_tts_falcon_kv_cache_test" --output-on-failure
2/2 Test #4: audio8_tts_falcon_kv_cache_test ...   Passed    0.50 sec
100% tests passed, 0 tests failed out of 2
```

And thanks for the extra measurements — qwen3_tts −32% (speech decoder −75.4%), irodori_tts −20.5% (codec decode −36.5%), voxcpm2 −22.9% (AudioVAE decoder −63.8%) — that is exactly the profile the change predicts: the chained codec/vocoder convolutions are where the strided im2col was costing the most. CI is re-running now.

0xShug0:
@gqf2008 Merged! Thanks :tada:

jasonchen31:
@gqf2008  fter this patch the RTF performance of **Audio8-TTS 0.6B** on CUDA GPU now dropped from 0.3x to 1.x after warming up.  Could you have a look?

Before:
audiocpp_server --version
audio.cpp dev
git: 7e90338 %cs
build: RelWithDebInfo, gcc 12.4.0, Linux x86_64
backends: cpu,cuda

performance log:
{
{
  "seed": 1234,
  "chunks": 1,
  "characters": 53,
  "chunk_budget": 1000,
  "timings": [
    {
      "chunk": 1,
      "characters": 53,
      "wall_ms": "3357.496567",
      "rtf": "0.336268"
    }
  ]
}

After:
audiocpp_server --version
audio.cpp dev
git: 662c7bd5 %cs
build: RelWithDebInfo, gcc 12.4.0, Linux x86_64
backends: cpu,cuda

performance log:
{
  "seed": 1234,
  "chunks": 1,
  "characters": 53,
  "chunk_budget": 1000,
  "timings": [
    {
      "chunk": 1,
      "characters": 53,
      "wall_ms": "11683.629531",
      "rtf": "1.113212"
    }
  ]
}

PR #450, opened 2026-09-04
[Feature] ModelScope download

IIIIIllllIIIIIlllll:
Since the repository on ModelScope has already been set up, is it necessary to add a feature to download from here? I'll go ahead and implement it first and put it here. It doesn't really make sense for two people to maintain separate repositories, and I'm not sure if it's necessary either. If all else fails, we could also add a note in the documentation to guide users who need it to that repository.

https://www.modelscope.cn/models/HereIsMark/audio.cpp-gguf

Sure, here's the English version:

## 1. On-demand switch (most common, no config changes)

Use the Python tool to pull any package from ModelScope:

```bash
# With an explicit ModelScope repo
python tools/model_manager_v2.py install qwen3_tts --source modelscope --source-repo HereIsMark/audio.cpp-gguf

# Without --source-repo, the spec own repo name is looked up on ModelScope
python tools/model_manager_v2.py install qwen3_tts --source modelscope

# sizes also supports it, for checking whether an installed package is current
python tools/model_manager_v2.py sizes q3_tts --models-root models --source modelscope --source-repo HereIsMark/audio.cpp-gguf
```

Simple rules:
- No `--source` = exactly the same behavior as before (Hugging Face)
- An unset or `main` revision in the spec → translated to ModelScope's default branch `master`; any other explicit revision (e.g. `v1.2`) passes through unchanged
- `--source-repo` alone (without `--source modelscope`) is rejected
- Note: check for updates the same `--source` you installed with — etags are source-specific, so cross-source checks may falsely report an update

## 2. Declared in the spec (for maintainers)

Set `download.kind` to `modelscope_snapshot` in `_specs/*.json`; both the native C++ manager and the Python tool understand it:

```json
"download": {"kind": "modelscope_snapshot", "repo": "audio-cpp/toy-model"}
```

`revision` is optional (defaults to `master`).

## 3. Mirror/test endpoints (rarely needed)

- ModelScope: `AUDIOCPP_MS_BASE_URL` (default `https://www.modelscope.cn`), honored by both native and Python
- Hugging Face: native uses `AUDIOCPP_HF_BASE_URL`, Python uses the standard `HF_ENDPOINT`

TL;DR: when Hugging Face is unreachable, just append `--source modelscopesource-repo HereIsMark/audio.cpp-gguf` to your install command.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2389,6 +2389,19 @@ if (ENGINE_BUILD_TESTS)
             NAME server_model_installer_test
             COMMAND server_model_installer_test
         )
+
+        add_executable(package_manager_modelscope_test
+            tests/unittests/test_package_manager_modelscope.cpp
+        )
+        target_link_libraries(package_manager_modelscope_test PRIVATE
+            Threads::Threads audiocpp_package_manager)
+        target_compile_definitions(package_manager_modelscope_test PRIVATE
+            AUDIOCPP_NATIVE_MANAGER_FIXTURE="${CMAKE_CURRENT_SOURCE_DIR}/tests/fixtures/native_model_manager_server.py")
+
+        add_test(
+            NAME package_manager_modelscope_test
+            COMMAND package_manager_modelscope_test
+        )
     endif()
 
     if (ENGINE_BUILD_MODEL_TESTS)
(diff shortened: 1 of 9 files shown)


0xShug0:
@IIIIIllllIIIIIlllll The PR builds one shared request-header path and reuses it for both Hugging Face and ModelScope. Since that helper can attach `HF_TOKEN / HUGGING_FACE_HUB_TOKEN,` ModelScope requests can accidentally carry a Hugging Face credential.

I’d suggest making request headers provider-scoped: Hugging Face requests may attach Hugging Face auth, while ModelScope requests should not receive HF auth and should use a separate ModelScope auth path.

IIIIIllllIIIIIlllll:
@0xShug0 Fixed

0xShug0:
@IIIIIllllIIIIIlllll Merged! Thanks!

Issue #429, opened 2026-09-03
[Bug] Higgs Audio TTS: periodic click every ~5.12 s — chunked codec decode overlap (8 frames) is smaller than the decoder receptive field (~27 frames per side)


HaujetZhao:
### Environment

- Build: `audio-v0.7.1-bin-windows-x64-cuda13.3.zip` (Windows 10 x64, CUDA 13.3 backend)
- Model: `higgs_audio_tts_4b_q8_0`
- Commit analyzed: `a088f90`

### Symptom

When generating longer audio with Higgs Audio TTS, the output has an audible click/glitch roughly every 5 seconds (at ~5 s, ~10 s, ~15 s, ~20 s, ...), as if the waveform breaks at a seam. Short generations (≈10 s or less) sound fine.

### Root cause

The official Python implementation decodes the whole sequence in one shot (`HiggsAudioTokenizer.decode()` in `boson_multimodal/audio_processing/higgs_audio_tokenizer.py`). `src/models/higgs_audio_tts/codec.cpp` adds chunked decoding (not present upstream) to bound memory:

```cpp
constexpr int64_t kCodecDecodeWindowFrames  = 128;  // 128 frames = 128 × 960 / 24000 = 5.12 s
constexpr int64_t kCodecDecodeOverlapFrames = 8;    // only 8 frames of left context, hard-trimmed concat
```

(`codec.cpp:64-65`; the chunk loop is around `codec.cpp:1584`. Frames ≤ 512 take the single-window path, which is why short audio is unaffected — and why the seams land at every 128 frames from sample 0, i.e. every 5.12 s.)

The problem: **the overlap is far smaller than the decoder's receptive field.** This DAC-style decoder is purely convolutional (5 × ConvTranspose blocks with residual units at dilations 1/3/9, all symmetric padding, i.e. non-causal). Walking the structure layer by layer:

| Item | Value |
|---|---|
| Decoder total receptive field | ≈ 53 frames (≈ 27 frames per side ≈ 1.1 s of audio) |
| Left context provided by chunking | 8 frames |
| Missing right context at each window tail | ≈ 27 frames |

Since the decoder is non-causal, samples up to ~1 s past each trim point deviate from the full-sequence decode, and each window tail also lacks right context. The hard concat at seams then produces a waveform discontinuity — the periodic click.

#### Option A: widen the context window (recommended, minimal change)

Borrow 32 frames on **both** sides and trim both after decoding:

```cpp
[...]

PR #454, opened 2026-09-04
higgs_audio_tts: give the codec decoder right-hand context at the end of the stream

niksedk:
## Problem

Higgs Audio v3 TTS output ends with a rising hiss over the final ~300 ms that is cut off abruptly. It is most audible on short utterances (e.g. `Hello, how are you doing?` with a cloned voice) but a faint version is there on long ones too. It is independent of sampling settings, seed, `reference_text`, punctuation, and it reproduces on v0.7.1 and current `main`.

The codec decoder is a non-causal conv stack. At the end of the stream the last frames see zero padding on their right instead of real context and decode to garbage that ramps up towards the edge. This is the same mechanism as the periodic click at chunk seams (#429 / #436), but at the end of the stream, where the chunked path's context frames do not help.

## Fix

`HiggsCodecRuntime::decode_codes()` now repeats the last frame `kCodecTailContextFrames` (8) times before decoding and trims the extra samples again, so the public contract (sample count) is unchanged. The previous body becomes `decode_codes_impl()`. 16 and 32 context frames give sample-identical output to 8, so 8 is enough. Cost is 8 extra frames per decode.

## Measurements

Metal (Apple M4), Q8_0 GGUF, `higgs_audio_tts` clone with a fixed seed, before → after. RMS of the last three 40 ms frames (dBFS) and peak sample of the last 100 ms:

| request | before | after |
|---|---|---|
| short, seed 1 | -39 -44 **-34**, peak 3990 | -41 -48 **-46**, peak 864 |
| short, seed 2 | -39 -46 **-32**, peak 3951 | -40 -47 **-41**, peak 1121 |
| long, 11.4 s | -48 -53 **-46** | -48 -53 **-53**, peak 334 |

The delayed/raw code tail was dumped to rule out stray BOC/EOC ids in the kept frames (there are none), and forcing EOC on finished codebooks during the delay flush like the reference implementation does gave bit-identical output, so the issue is purely in the codec decode edge.

Tested with `scripts/build_metal.sh --model-set custom --models higgs_audio_tts --deployment-build`, server mode, multiple requests in one session.

### DIFF
--- a/include/engine/models/higgs_audio_tts/codec.h
+++ b/include/engine/models/higgs_audio_tts/codec.h
@@ -117,6 +117,11 @@ public:
     void release_runtime_graphs();
 
 private:
+    HiggsCodecDecodeOutput decode_codes_impl(
+        const std::vector<int32_t> & codes,
+        int64_t frames,
+        int64_t codebooks) const;
+
     std::shared_ptr<const HiggsAssets> assets_;
     ggml_backend_t backend_ = nullptr;
     core::BackendType backend_type_ = core::BackendType::Cpu;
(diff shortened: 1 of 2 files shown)


0xShug0:
@niksedk Thanks for the PR! I will test it after wrapping up release 0.7.2

niksedk:
> @niksedk Thanks for the PR! I will test it after wrapping up release 0.7.2

thx :)

0xShug0:
@niksedk PR merged. Thanks!

PR #456, opened 2026-09-04
tools: check the WebUI catalog in the loader sync check

CryptVenture:
`tools/check_loader_catalog_sync.py` already checks that runtime loaders, `model_specs/` and `model_manager_v2` agree. It does not look at `webui/configs/models_catalog.json`, which is how a catalog entry can name a `download_id` no package defines, or a `path` no package installs into, and nothing notices.

This teaches the existing checker to read the catalog, and reports what it finds as advisories by default so the check keeps passing on `main` today.

## Behaviour

```
python3 tools/check_loader_catalog_sync.py                   exit 0, 86 catalog advisories
python3 tools/check_loader_catalog_sync.py --strict-catalog  exit 1, the same findings as errors
python3 tools/check_loader_catalog_sync.py --self-test       10 tests OK
```

Default mode prints the advisories and a summary line naming the count, so the state is visible without failing anyone's build. `--strict-catalog` promotes them to errors, for a CI job or a local gate once the catalog is clean.

What it checks per entry, for families that have a `model_specs/<family>.json`:

- `download_id` is a `packages[].id` in that family's spec.
- `path` is the package's `target_directory`, or an installed file inside it.

Example finding:

```
webui/configs/models_catalog.json: catalog entry 'pocket-tts' download_id 'pocket_tts'
  is not a packages[].id in model_specs/pocket_tts.json
webui/configs/models_catalog.json: catalog entry 'pocket-tts' path 'models/pocket-tts'
  is neither the target_directory nor an installed file of model_specs/pocket_tts.json
  recommended package 'pocket_tts_english_q8_0' (expected models/PocketTTS-GGUF/english[/<installed file>])
```

## With the catalog corrected

Run against `main` it reports 86 advisories. Run with #428's catalog in place:

```
ok: runtime loaders, model_specs, model_manager_v2, and the WebUI catalog are in sync
--strict-catalog  exit 0
```

So the two together take the catalog to clean and keep it there.

## Known gap

An entry whose `family` has no `model_specs/<family>.json` at all is skipped rather than reported — currently the `breeze_tts`, `cosyvoice3` and `chatterbox_turbo` entries. Those cannot install either, but flagging a missing spec is a different check with a different judgement call behind it (a family may legitimately ship ahead of its spec), so it is not in this change. Happy to add it as a further advisory if you want it.

## Self-test

`--self-test` runs 10 unit tests over synthetic spec/catalog pairs covering both checks, the advisory/error split, and the exit codes, so the checker's own logic is verified without depending on the repo's current contents.

### DIFF
--- a/tools/check_loader_catalog_sync.py
+++ b/tools/check_loader_catalog_sync.py
@@ -1,11 +1,18 @@
 #!/usr/bin/env python3
-"""Check sync between runtime loaders, model_specs packages, and model_manager_v2.
+"""Check sync between runtime loaders, model_specs packages, model_manager_v2,
+and the WebUI catalog.
 
 Schema correctness is owned by the typed model-spec validator. This script only
 checks cross-system drift:
 
 - registered loader families vs model_specs families
 - model_specs packages vs model_manager_v2 package output
+- model_specs packages vs the packages the native package manager publishes
+- webui/configs/models_catalog.json vs model_specs (family, download id,
+  install path, task/mode vocabulary) and vs the packages a catalog entry can
+  actually reach
+- webui/configs/model_params.json parameter groups vs the request options the
+  owning spec documents
 See docs/maintainers/loader_and_catalog.md.
 """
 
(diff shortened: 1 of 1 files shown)


0xShug0:
The checker treats every installable GGUF package in the resolved target_directory as reachable. But the WebUI normally resolves related packages by exact download_id stem first, not by all packages in the same directory; only special families use “all GGUF in same target dir” behavior.

The PR says catalog issues are advisory by default, but the script now hard-requires models_catalog.json, model_params.json, and session.cpp to exist/read successfully

0xShug0:
will reopen the PR once the author responds and addresses the comments.

PR #490, opened 2026-09-08
refactor: build Fish HIP fast sampler as a proper HIP CMake target

IIIIIllllIIIIIlllll:
Follow-up to #486. Now that the issue has surfaced, let's fix the root cause instead of the symptom.

**Background**

 Fish Audio fast sampler is a small HIP kernel that does exact top-k on the device, so only the compact candidates cross back to the host instead of the full 4096-entry logits on every codebook step.

The kernel itself is fine; the problem is how it is wired in. For readers not familiar with the HIP backend code, the current integration couples the model to HIP in two ways:

1. **Build system:** thecu` is compiled by a hand-rolled `hipcc` custom command and injected into the static `engine_runtime` as an external object. CMake cannot see it, so its HIP runtime dependency (`hip::host`) is invisible — that is exactly what broke with dynamic ggml backends and required #486.
2. **Source code:** `ar.cpp` calls the sampler's HIP entry points directly and passes `logits_->data` as a raw device pointer, relying on ggml-hip's internal memory layout rather than any public ggml contract.

**What this PR does**

This PR addresses (1) only: the sampler becomes a first-class HIP-language CMake (scoped `cmake_minimum_required(3.21)` for `enable_language(HIP)`, mirroring ggml-hip) that PUBLIC-links `hip::host`. The dependency now propagates through `engine_runtime` to executables in both static and `ENGINE_CPU_ALL_VARIANTS=ON` builds, so the #486 workaround is removed. As a bonus, the previously dead `CMAKE_HIP_ARCHITECTURES` path now actually takes effect.

No functional change: the kernel, `ar.cpp`, and non-HIP builds are untouched. Windows HIP still skips the sampler (CMake does not support the HIP language on Windows).

Addressing (2) — decoupling the model code from raw HIP calls, e.g. by moving top-k into the ggml graph so all backends benefit — is deliberately left for a separate discussion.

**Testing** (Linux + ROCm 7.14, gfx115 / Strix Halo)

- Static-backend and `ENGINE_ENABLE_CPU_ALL_VARIANTS=ON` builds both link `libamdhip64.so.7`; sampler symbols present in executables
- Sampler compiles as HIP language with `--offload-arch=gfx1` via both `GPU_TARGETS` and `CMAKE_HIP_ARCHITECTURES`
- `-DAUDIOCPP_MODEL_SET=core` (fish_audio excluded) configures and builds
- fish_audio TTS smoke tests pass on HIP (static + dynamic backends) and CPU

AI usage: Kimi K3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -999,33 +999,6 @@ audiocpp_add_model(fish_audio
         engine::models::fish_audio::make_fish_audio_loader
 )
 
-# Fish Fast-AR emits a 4096-entry logits tensor for each acoustic codebook.
-# On Linux HIP, use a small device-side exact top-k helper so only the compact
-# sampling candidates cross back to the host. Other backends stay unchanged.
-if (ENGINE_ENABLE_HIP AND UNIX AND NOT APPLE)
-    find_program(AUDIOCPP_HIPCC hipcc REQUIRED)
-    set(AUDIOCPP_FISH_HIP_FAST_SAMPLER_OBJ "${CMAKE_CURRENT_BINARY_DIR}/fish_hip_fast_sampler.o")
-    set(AUDIOCPP_FISH_HIP_TARGETS ${CMAKE_HIP_ARCHITECTURES})
-    if (NOT AUDIOCPP_FISH_HIP_TARGETS)
-        set(AUDIOCPP_FISH_HIP_TARGETS ${GPU_TARGETS})
-    endif()
-    set(AUDIOCPP_FISH_HIP_ARCH_FLAGS)
-    foreach(AUDIOCPP_FISH_HIP_ARCH IN LISTS AUDIOCPP_FISH_HIP_TARGETS)
-        list(APPEND AUDIOCPP_FISH_HIP_ARCH_FLAGS "--offload-arch=${AUDIOCPP_FISH_HIP_ARCH}")
-    endforeach()
-    add_custom_command(
-        OUTPUT ${AUDIOCPP_FISH_HIP_FAST_SAMPLER_OBJ}
-        COMMAND ${AUDIOCPP_HIPCC} -O3 -fPIC ${AUDIOCPP_FISH_HIP_ARCH_FLAGS} -std=c++17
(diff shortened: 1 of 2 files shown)


0xShug0:
@IIIIIllllIIIIIlllll Quick question: the original condition is `if (ENGINE_ENABLE_HIP AND UNIX AND NOT APPLE AND "fish_audio" IN_LIST AUDIOCPP_LINKED_MODELS)`  The new condition in the PR removes the fish_audio check, so the sampler is built unconditionally for all Linux HIP builds. Could this break composite builds?

IIIIIllllIIIIIlllll:
@0xShug0 
Good catch — but no, composite builds are safe. This was verified: `-DAUDIOCPP_MODEL_SET=core` (fish_audio excluded) configures and builds cleanly on ROCm 7. / gfx1151.

Why removing the check doesn't break anything:

1. **The old check only gated *linking*, not the compile definition.** The old code set `ENGINE_HAS_HIP_FISH_FAST_SAMPLER=1` on `engine_model_fish_audio` unconditionally (no `fish_audio IN_LIST` guard), and only the `.o` append to `AUDIOCPP_RUNTIME_OBJECTS` was gated. The new code keeps the define unconditional and makes the link unconditional — consistent each other, just like before.

2. **An unreferenced static library is inert.** When fish_audio is not in the composite, none of its objects are in `engine_runtime`, so nothing references the sampler's symbols and the linker pulls zero objects from. Undefined references are impossible.

3. **`hip::host` reaching all Linux HIP executables is already the status quo.** #486 applied `find_package(hip)` + `hip::host` to *every* HIP build, not just fish_audio composites. `libamdhip64` is always present in HIP builds since ggml-hip itself requires it.

One deliberate trade-off: the old custom command only compiled the `.cu` when the `.o` was referenced, while the new target always built in HIP configure (one small file, a few seconds). I accepted that for simplicity — the alternative is gating `add_subdirectory` on the same `IN_LIST` check, which would reintroduce a second condition that must stay in sync with define. Happy to change it if you prefer the lazy build.

0xShug0:
@IIIIIllllIIIIIlllll Thanks and merged!

PR #491, opened 2026-09-08
perf(mira-tts): accelerate CPU and Vulkan inference

mirek190:
## Summary

This continues the MiraTTS performance work from #395, focusing on CPU and Vulkan while preserving generated output exactly.

- vectorizes the CPU FlashSR residual and output convolutions across contiguous output positions while preserving each sample's accumulation order
- uses MiraTTS's model-local 8,193-row speech/EOS output head on Vulkan as well as CPU, avoiding projection of the unused full text vocabulary
- keeps Vulkan prefill on the existing materialized grouped-K/V path, but uses grouped cached K/V views for single-token decode to avoid a copy on every generated token
- adds block-boundary FlashSR coverage and a reusable before/after validator that rejects changed WAV bytes, hashes, metadata, request sets, or streaming event counts
- documents rejected experiments so noisy or output-changing results are not presented as optimizations

The shared Qwen runtime is unchanged. CUDA behavior is unchanged. No precision, sampling, token-budget, or output-quality setting is reduced.

## Same-backend results

Windows Release/MSVC, Ryzen 9 7950X3D, RTX 3090 Vulkan device 1, eight CPU threads. Times exclude initial model loading. Each candidate is compared only with the matching backend and precision baseline.

| Backend | Request | Before | After | Improvement |
|---|---|---:|---:|---:|
| CPU Q8 | Cold, 5.12 s audio | 8.245 s | 7.018 s | 14.9% |
| CPU Q8 | Repeated, 5.12 s audio | 8.210 s | 7.035 s | 14.3% |
| CPU Q8 | Second prompt, 5.12 s audio | 8.211 s | 7.082 s | 13.8% |
| CPU Q8 | Long-form, 15.36 s audio | 26.923 s | 22.381 s | 16.9% |
| CPU Q8 | Repeat after long-form | 8.180 s | 7.016 s | 14.2% |
| Vulkan Q8 | Cold, 5.12 s audio | 0.994 s | 0.910 s | 8.5% |
| Vulkan Q8 | Repeated, 5.12 s audio | 0.881 s | 0.789 s | 10.4% |
| Vulkan Q8 | Second prompt, 5.12 s audio | 0.930 s | 0.797 s | 14.3% |
| Vulkan Q8 | Long-form, 15.36 s audio | 2.942 s | 2.652 s | 9.9% |
| Vulkan Q8 | Repeat after long-form | 0.937 s | 0.837 s | 10.7% |

BF16 is covered as well. CPU BF16 offline reductions range from 13.5% to 16.7% across the same five cases. Warm Vulkan BF16 reductions range from 8.4% to 15.4%; Vulkan BF16 cold time is effectively unchanged and is not claimed as a speedup. Full offline and streaming tables are in `tests/mira_tts/CPU_VULKAN_OPTIMIZATION.md`.

## Context against the requested #395 table

The linked #395 comment measured CUDA Q8. The new Vulkan numbers below are therefore context, not a same-backend before/after claim.

| Test | #395 previous CUDA Q8 | #395 optimized CUDA Q8 | This PR Vulkan Q8 | Vulkan vs optimized CUDA |
|---|---:|---:|---:|---:|
| Cold request, 5.12 s audio | 1.707 s | 0.874 s | 0.910 s | 4.1% slower |
| Repeated request, 5.12 s audio | 1.561 s | 0.798 s | 0.789 s | 1.1% faster |
| Second short prompt, 5.12 s audio | 1.566 s | 0.798 s | 0.797 s | effectively equal |
| Long-form, 15.36 s audio | 4.771 s | 2.530 s | 2.652 s | 4.8% slower |
| Repeat after long-form, 5.12 s audio | 1.584 s | 0.835 s | 0.837 s | effectively equal |

This puts Vulkan Q8 within about 5% of the optimized CUDA Q8 results from #395 on this machine, while the purpose of this PR remains the independently measured same-backend gains above.

## Validation

- all 28 first-pass CPU/Vulkan Q8/BF16 offline and streaming comparisons produced identical complete WAV bytes, hashes, frame counts, formats, durations, and event counts
- all 14 second CPU-pass outputs are byte-identical to their matching first-pass outputs
- all 14 Vulkan decode-K/V comparisons are byte-identical to their matching pre-change outputs
- offline/streaming lifecycle validation passes for CPU and Vulkan, Q8 and BF16
- `flashsr_utility_test` passes two Python-reference fixtures and eight boundary sizes: 1, 7, 85, 86, 255, 256, 257, and 1025 samples
- six comparison-tool unit tests pass
- current CPU and Vulkan CLI/server/benchmark targets build successfully after merging current `origin/main`
- loader/catalog synchronization passes (existing safetensors warnings only)

The benchmark document includes commands, artifacts, stage timings, limitations, and experiments that were tested but removed.

### DIFF
--- a/src/community_models/mira_tts/generator.cpp
+++ b/src/community_models/mira_tts/generator.cpp
@@ -124,11 +124,16 @@ modules::QwenCausalDecoderConfig decoder_config(
         modules::QwenDecoderStaticCacheUpdateMode::DirectSetRows;
     if (backend_type == core::BackendType::Vulkan) {
         // Mira's projections are sensitive to Vulkan's default reduced
-        // precision. Materialize grouped K/V heads for attention as well:
-        // the strided-view path diverges during prompt evaluation.
+        // precision. Prompt evaluation still needs materialized grouped K/V:
+        // the strided-view prefill path diverges. Single-token decoding can
+        // use the cached views without copying the grouped heads each step.
         out.stack.projection_precision = GGML_PREC_F32;
         out.stack.runtime.attention.prefill_mode = modules::QwenDecoderAttentionMode::FlashGrouped;
         out.stack.runtime.attention.static_mode = modules::QwenDecoderAttentionMode::FlashGrouped;
+        const char * view_decode = std::getenv("AUDIOCPP_MIRA_TTS_VULKAN_VIEW_DECODE");
+        if (view_decode == nullptr || view_decode[0] != '0') {
+            out.stack.runtime.attention.static_mode = modules::QwenDecoderAttentionMode::FlashGroupedViewKV;
+        }
     }
     out.logits_size = config.vocab_size;
     out.logits_mode = modules::QwenCausalDecoderLogitsMode::LastStep;
@@ -181,7 +186,9 @@ std::shared_ptr<MiraQwenWeights> load_weights(
     out->store->upload();
     out->lm_head = out->token_embedding;
     const char * sparse_head = std::getenv("AUDIOCPP_MIRA_TTS_SPARSE_HEAD");
-    if (backend_type == core::BackendType::Cpu &&
+    // Vulkan can project the same speech/EOS window as CPU without lowering
+    // precision. Keep the full-head diagnostic available on both backends.
+    if ((backend_type == core::BackendType::Cpu || backend_type == core::BackendType::Vulkan) &&
         !(sparse_head != nullptr && sparse_head[0] == '0')) {
         // Only MiraTTS knows its speech/EOS alphabet. Keep its weight window
         // here, presenting an ordinary, correctly sized head to shared Qwen.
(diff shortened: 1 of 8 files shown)


0xShug0:
@mirek190 Thanks for pushing the perf! Merged.

mirek190:
Optimizations were made by Qwen 3.8 27b locally via PI agent ... can you believe it ;)

0xShug0:
Looking at the optimization history, it's actually pretty impressive. The changes are well scoped and don’t mess with GGML (which  Codex commonly did even with prompts said not do that). I'm wondering this is because of the use of PI, or Qwen3.8 vs Codex. Maybe it could help close the Kokoro CPU performance gap compared with the Python implementation.

mirek190:
> Looking at the optimization history, it's actually pretty impressive. The changes are well scoped and don’t mess with GGML (which Codex commonly did even with prompts said not do that). I'm wondering this is because of the use of PI, or Qwen3.8 vs Codex. Maybe it could help close the Kokoro CPU performance gap compared with the Python implementation.

sure it is worth to check ....

PR #449, opened 2026-09-04
sanotts: sanoTTS voice family community model (seven voices, 294k-2.27M params, en/vi/id, GGUF FP32)

voidash:
Adds the **sanoTTS voice family** as a community model: seven very small offline TTS voices (294k–2.27M parameters; English, Vietnamese, Indonesian), the smallest of which also runs on microcontrollers. Upstream project: https://github.com/Ampixa/sanoTTS · Packages: [ampixa/sanoTTS `gguf/`](https://huggingface.co/ampixa/sanoTTS/tree/main/gguf) — FP32 tensors in PyTorch shapes with `audiocpp.tensor_ranks/shapes` exact-shape metadata and embedded model specs, fully standalone.

Two graphs share the `sanotts` family, dispatched by the package config:

- **nano** (heart 2.27M, heart-nano 294k · 24 kHz): duration student → contextual acoustic student → mel-100 → noise-fed ConvNeXt-1D decoder → [log-magnitude | phase] head → `HostLogMagnitudePhaseISTFT`. A seed picks one of many valid renderings.
- **piperlite** (amy 1.45M, hfc 1.83M, kristin 1.40M, vi 1.57M, id 1.56M · 22.05 kHz): duration + acoustic students → 192-channel latent → 3-stage ConvTranspose1d decoder with dilated residual banks (kristin adds a learned post filter). Deterministic.

## Implementation

- Follows the `inflect_v2` pattern throughout: `SpecBackedVoiceModelConfig`, `BackendWeightStore`, `CacheSlots`-cached graphs keyed by token/frame count, framework modules (`LayerNormModule` eps 1e-6, `GeluModule` ExactErf, `LinearModule`, `EmbeddingModule`, `ggml_conv_1d_dw`, `ConvTranspose1dModule` with the CPU padding construction). Shared front-end structure lives in an internal `graph_common.h`.
- eSpeak-ng is GPL-3.0 and is opened with `dlopen` at runtime, never linked — same treatment as `inflect_v2`. The nano front end reproduces the phonemizer punctuation-preservation pipeline plus the misaki E2M rewrite (exact tie-mode bits); the piperlite front end reproduces Piper's convention (untied phonemes, NFD decomposition to codepoints, per-voice `phoneme_id_map`, `[BOS, PAD, (id, PAD)…, EOS]` framing, schwa fallback for out-of-vocab ids, and regional-variant-first voice selection matching phonemizer's behavior on espeak-ng ≥ 1.49).
[...]

PR #495, opened 2026-09-09
sanotts: eleven more language voices (cs, de, es, fr, it, pt, ro, ru, tr, ne, hi) and the acoustic calibration adapter

voidash:
Follow-up to #449. Adds **eleven more sanoTTS voices** — Czech, German, Spanish, French, Italian, Brazilian Portuguese, Romanian, Russian, Turkish, **Nepali and Hindi** — one package each. Packages: [ampixa/sanoTTS `gguf/`](https://huggingface.co/ampixa/sanoTTS/tree/main/gguf).

Nine of the eleven are the same piperlite lineage as `amy`, `vi` and `id` and needed no graph code: duration student → contextual acoustic student → 192-channel latent → 3-stage ConvTranspose1d decoder with dilated residual banks, deterministic, 22.05 kHz. Each voice carries its own eSpeak-ng voice and `--language` code in the config the exporter embeds, so `pt` drives `pt-br` and validates `--language pt` with no code aware of it.

**Hindi needed one new op**, described below. Nepali did not.

| Package | Params | eSpeak voice | Teacher |
|---|---:|---|---|
| `sanotts_cs_orig` | 1,568,044 | `cs` | `cs_CZ-jirka-medium` |
| `sanotts_de_orig` | 1,565,324 | `de` | `de_DE-thorsten-medium` |
| `sanotts_es_orig` | 1,562,924 | `es` | `es_ES-davefx-medium` |
| `sanotts_fr_orig` | 1,565,324 | `fr` | `fr_FR-siwis-medium` |
| `sanotts_it_orig` | 1,565,484 | `it` | `it_IT-serena-medium` |
| `sanotts_pt_orig` | 1,565,324 | `pt-br` | `pt_BR-cadu-medium` |
| `sanotts_ro_orig` | 1,565,164 | `ro` | `ro_RO-mihai-medium` |
| `sanotts_ru_orig` | 1,566,764 | `ru` | `ru_RU-irina-medium` |
| `sanotts_tr_orig` | 1,562,284 | `tr` | `tr_TR-dfki-medium` |
| `sanotts_ne_orig` | 1,474,771 | `ne` | `ne_NP-chitwan-medium` |
| `sanotts_hi_orig` | 1,499,407 | `hi` | `hi_IN-pratham-medium` |

## The one graph change: the acoustic calibration adapter

Hindi's acoustic student is `calibrated` rather than plain `token_context` — a finished token_context body with a small layer trained onto its output to re-calibrate the latent it hands the decoder. The runtime had no such op, so this adds one:

```
latent -> [depthwise conv, bias-free] -> [x + up(tanh(down(x)))] -> x * scale + bias -> decoder.pre
```

Seven extra tensors at `acoustic.adapter.*`, declared per voice as `acoustic.adapter` in the embedded config (`mode`, `kernel`, `rank`) and validated at load like the rest of the inventory. A voice that omits the key keeps `AdapterMode::None` and its latent goes straight to `decoder.pre` as before — of the eighteen packages only `hi` declares one.

Two things make this safe to add rather than something to take on faith:

- **The ten voices that predate it re-measure to correlations identical digit for digit** (`cs`, `de`, `es`, `fr`, `it`, `pt`, `ro`, `ru`, `tr`, and the `amy` control), which is what says the new path is inert when a voice declares no adapter.
- **Hindi's own parity number is the evidence the adapter is right.** An adapter implemented wrongly does not land within half an int16 LSB of a reference that computes it; it lands somewhere else entirely.

## Runtime parity

Ubuntu x86-64, gcc, 12 cores, default 4 threads. Each voice installed from Hugging Face with `tools/model_manager_v2.py install sanotts_<voice>_orig`, then rendered from 3–4 out-of-domain Tatoeba sentences and compared with the project's numpy reference — same host, same eSpeak-ng 1.52.0 build for both stacks. As in #449, the reference's float audio is compared against audio.cpp's int16 WAV.

| Voice | Sample counts | Min correlation | Samples that quantise identically | Max int16 gap |
|---|---|---:|---:|---:|
| cs | identical | 0.9999999961 | 99.847% | 1 LSB |
| de | identical | 0.9999999865 | 99.929% | 1 LSB |
| es | identical | 0.9999999970 | 99.645% | 1 LSB |
| fr | identical | 0.9999999851 | 99.897% | 1 LSB |
| it | identical | 0.9999999962 | 99.832% | 1 LSB |
| pt | identical | 0.9999999762 | 99.897% | 1 LSB |
| ro | identical | 0.9999999859 | 99.893% | 1 LSB |
| ru | identical | 0.9999999678 | 99.944% | 1 LSB |
| tr | identical | 0.9999999435 | 99.958% | 1 LSB |
| **ne** | identical | **0.9999999865** | **99.916%** | 1 LSB |
| **hi** (adapter) | identical | **0.9999999943** | **99.859%** | 1 LSB |
| amy (already shipped, re-measured as a control) | identical | 0.9999999920 | 99.818% | 1 LSB |

Nepali and Hindi both clear the ≥ 0.99999996 bar #449 set; their implied float-domain divergence between the runtimes is 5.3e-07 and 1.9e-06.

**Turkish does not clear that bar — it lands at 0.99999994 — and I am not lowering the bar quietly, so here is what that number is.** Correlation against a 16-bit WAV has a per-voice ceiling: the quantisation noise is fixed, so the correlation it permits scales with the voice's amplitude. Predicting each voice's ceiling from its RMS alone reproduces the measured correlation to within 1e-8 for every voice in the table, including the amy control. Turkish is simply the quietest voice in the set (RMS 0.027, peak 0.271) and sits on its own ceiling like everyone else.

The scale-free version of the same measurement says all eleven agree at least as tightly as the shipped `amy`: audio.cpp's int16 sample is the correctly rounded quantisation of the reference float sample for 99.6–99.96% of samples, never off by more than one LSB, and the implied float-domain divergence between the runtimes is ≤ 5.1e-06 — Turkish's is 4.5e-07, the *tightest* in the set. If you would rather hold the literal 0.99999996 line, say so and I will drop `tr` from the PR; I think the evidence says the runtime is fine and the metric has a floor.

**Other checks**: language validation works from config alone (`--language de`/`ne`/`hi` accepted, `--language en` on those voices rejected); a 806-character German paragraph renders 37.7 s of audio in 2.7 s (~14× real time) through the long-form chunker; `tools/check_loader_catalog_sync.py` passes; `sanotts_frontend_test` passes under `-DENGINE_BUILD_MODEL_TESTS=ON`.

**eSpeak-ng version sensitivity** (caveat found while verifying): against Debian's eSpeak-ng 1.51 instead of 1.52.0, eight of the nine European voices render byte-identical audio, but two of the four Russian probe sentences phonemize differently and come out 3.5–8% shorter. That is the front end, not the runtime, and it applies to the family generally — a voice is only as reproducible as the eSpeak-ng under it.

## Quality evidence — intelligibility only

The only quality evidence is automatic transcription: 16 out-of-domain Tatoeba sentences per language, rendered by the numpy reference, transcribed by Whisper small, scored against the source.

| Voice | CER | WER |
|---|---:|---:|
| pt | 0.014 | 0.038 |
| it | 0.026 | 0.080 |
| ru | 0.031 | 0.120 |
| de | 0.032 | 0.099 |
| es | 0.050 | 0.147 |
| tr | 0.059 | 0.303 |
| fr | 0.088 | 0.221 |
| ro | 0.104 | 0.392 |
| cs | 0.111 | 0.351 |
| **hi** | **0.289** | **0.662** |
| **ne** | **0.576** | **1.116** |

Read that narrowly. It says the words survive; it says nothing about how these voices sound. 16 sentences per language is directional, not precise. **No listening study has been run on any of these voices, and there is no SCOREQ or MOS estimate for any of them.**

**Nepali and Hindi are the two weakest voices in the set and I want that stated plainly rather than buried.** CER is not comparable across languages, though, because Whisper small is far better at some than at others — and these are two it is weak at. The comparable number is each student against its own Piper teacher through the same ASR, since the teacher is the ceiling on what any distillation of it can reach:

| Voice | Student CER | Teacher CER | Retained |
|---|---:|---:|---:|
| ne | 0.576 | 0.515 | 0.89× |
| hi | 0.289 | 0.154 | 0.53× |

Whisper small puts the *Nepali teacher* — a published Piper voice, not ours — at 0.515 CER, so the Nepali absolute is mostly a statement about the transcriber and only the small gap to its teacher is evidence about the voice. There is no better ASR judge of Nepali available to me. Hindi keeps about half its teacher's margin, which is the same retention German shipped at in this PR (0.475×). If you would rather these two waited for a listening test, that is a reasonable call and I will pull them.

## Not in this PR

- **Chinese**, and *not* because of its front end — that question is settled, not open. Piper ships Chinese under two different front ends: `zh_CN-chaowen-medium` and `zh_CN-xiao_ya-medium` declare `phoneme_type: "pinyin"` over an 85-symbol pinyin inventory that eSpeak-ng cannot drive, while our teacher `zh_CN-huayan-medium` declares `phoneme_type: "espeak"` with `espeak.voice: "cmn"` over the same 152-symbol IPA inventory as every other eSpeak Piper voice. It is the eSpeak one, eSpeak-ng's `cmn` reads it correctly, and the teacher transcribes at **0.090 CER through exactly this path**. What fails is the distilled student: **0.468 CER** on the same sentences, keeping **0.19×** of its teacher against 0.475× for the weakest voice shipping here. Nearly half the characters are wrong. It is held back for being unintelligible, not for being mis-phonemized, and ships when a student worth shipping exists. (Both numbers fold traditional and simplified to simplified before scoring — Tatoeba mixes the two and Whisper answers in either, which would charge the voice for orthography it never pronounced. Unfolded, the same rows read 0.144 teacher and 0.502 student; the gap is the same.)
- **Arabic.** The `ar_JO` weights are trained, exported and run correctly — measured on the same harness at 0.99999998 correlation with identical sample counts — but they are held back, because *both* stacks are wrong in the same way. Piper runs a separate 1.16M-parameter ONNX diacritizer over Arabic text before eSpeak-ng sees it: the script omits the short vowels eSpeak-ng needs. Without it, 44.55% of the phoneme ids are correct (measured over 1,977 FLORES sentences) — that is a different word, not a degraded one — and this front end has no diacritizer. Documenting "pass in pre-diacritized text" would not reach anyone who simply types Arabic, so Arabic ships when a diacritizer does. We have a working port of that model (browser/JS) to draw on.
- **The ~511k "tiny" variants** of eight of the European languages. Separate size tier, separate evidence; they should come with their own numbers rather than ride along here.

## Limitations

- Offline, FP32, CPU-verified, as with #449.
- One language per voice; a session rejects text tagged with another language.
- eSpeak-ng must be installed with the data for that language, or pointed at via the `sanotts.espeak_library_path` / `sanotts.espeak_data_path` session options.
- The eleven new GGUFs embed the model spec as it stood when each was exported; the seven packages from #449 still embed the spec as it stood then. Harmless (the embedded spec is only a fallback for a standalone GGUF), but it does mean an old file lists fewer siblings than a new one.

https://claude.ai/code/session_01P1iL37FdfJkGxdGrpjH1we


0xShug0:
@voidash Merged, thanks!

PR #497, opened 2026-09-10
Add VibeVoice ASR Streaming 7B

0xShug0:
Draft PR for early review and conflict tracking.

This adds the VibeVoice ASR Streaming 7B runtime path and GGUF packaging/spec/docs updates. The branch is opened as draft first so review can happen before spending more time on conflict resolution and final validation.

Current validation includes basic CUDA ASR runs for the streaming 7B GGUFs. Speaker-attributed output is still being checked against the official Python baseline before this is marked ready.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -968,6 +968,7 @@ audiocpp_add_model(vibevoice_asr
         engine/models/vibevoice_asr/loader.h
     LOADERS
         engine::models::vibevoice_asr::make_vibevoice_asr_loader
+        engine::models::vibevoice_asr::make_vibevoice_asr_streaming_loader
 )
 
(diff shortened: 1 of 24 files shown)


0xShug0:
Updated with a decoder-cache ownership fix for streaming VRAM.

Validation on a 300s Q8 streaming server request:

| Case | Peak VRAM | Streaming RTF | Events | Words | Output |
|---|---:|---:|---:|---:|---|
| Before | 25053 MiB | 0.0638 | 103 | 814 | baseline |
| After | 17951 MiB | 0.0456 | 103 | 814 | byte-identical transcript |

Offline VibeVoice ASR four-speaker output also matches the existing baseline exactly.

PR #500, opened 2026-09-10
perf(fish_audio): keep Fast-AR inputs on HIP stream

reezex0-ux:
## Summary

Follow-up to #386. The HIP Fast-AR top-k path still rebuilds each next-codebook input on the CPU, uploads that embedding back to the GPU, and updates position/mask state through small host-to-device transfers.

This change keeps those recurrent Fast-AR inputs/state updates on the HIP stream:

- uploads the Fast-AR embedding table once into the HIP sampler workspace;
- gathers the selected embedding row directly on the device for subsequent codebooks;
- updates Fast-AR position and attention-mask state with a small HIP kernel on the GGML HIP stream;
- reuses the existing exact HIP top-k sampler after each graph run.

The resident path is only used for the same Linux HIP / supported `top_k` range already handled by the HIP sampler. Other backends and unsupported sampling settings keep the existing path.

## Validation

Tested against current `main` (`16976028a60ea6a70dc8e25ab70fa7a80077ba96`) on:

- Radeon RX 7900 XTX (`gfx1100`)
- ROCm 7.2.1 / HIP
- Fish Audio S2 Pro Q8_0 GGUF

Build:

```bash
bash scripts/build_linux.sh \
  --backend hip \
  --gpu-targets gfx1100 \
  --model-set custom \
  --models fish_audio \
  --target audiocpp_cli
```

For an isolated comparison, three identical requests were run in one loaded process using the same public benchmark sentence and settings as #386. Request 1 was treated as warmup; requests 2-3 are the warm measurements.

| Path | Warm #2 | Warm #3 | Warm mean |
| --- | ---: | ---: | ---: |
| Existing HIP top-k path | 4068.91 ms | 4138.90 ms | 4103.91 ms |
| Resident embedding/state path | 3959.96 ms | 4016.92 ms | 3988.44 ms |

That is about **2.81% lower warm wall time** in this isolated test.

All six WAVs from the OFF/ON comparison were bit-identical:

```text
SHA-256: dfbc6ec75603a05c85d97e5dbf24b75a77dcc2386e5087159cd9d7d673989533
```

## Scope

This PR intentionally does not include chained Fast-AR, Slow-AR changes, BF16/cache fusions, or other experiments from the larger AMD optimization branch.

### DIFF
--- a/src/models/fish_audio/ar.cpp
+++ b/src/models/fish_audio/ar.cpp
@@ -13,6 +13,7 @@
 
 #ifdef ENGINE_HAS_HIP_FISH_FAST_SAMPLER
 #include "hip_fast_sampler.h"
+#include <ggml-cuda.h>
 #endif
 
 #include "engine/framework/core/constant_tensor_cache.h"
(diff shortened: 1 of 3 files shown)


0xShug0:
@reezex0-ux Thanks! PR merged.

PR #512, opened 2026-09-11
fix(vocoders): fix narrowing conversion in HiFT vocoder on 32-bit tar…

weidu-tpvision:
In 32-bit architectures, size_t is an unsigned 32-bit integer (uint32_t). Evaluating runtime expressions with 64-bit integer literals (ull) produces an unsigned long long that triggers a strict C++ -Wnarrowing error during aggregate initialization of ggml_init_params.

Compute context_size with an explicit static_cast<size_t> before initializing ggml_init_params.

### DIFF
--- a/src/framework/modules/vocoders/hift_vocoder.cpp
+++ b/src/framework/modules/vocoders/hift_vocoder.cpp
@@ -12,6 +12,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstring>
+#include <limits>
 #include <numeric>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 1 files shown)


0xShug0:
@weidu-tpvision Thanks for the PR. The issue is real, but the current fix only hides the compile error. On a 32-bit target, the explicit static_cast<size_t> can still wrap the requested HiFT vocoder graph context size.
For example, when frames == 896, the vocoder requests 4096 MiB, but a 32-bit size_t receives 0 MiB after the cast.

```
frames=1    requested_context_mib=516   actual_32bit_context_mib=516
frames=895  requested_context_mib=4092  actual_32bit_context_mib=4092
frames=896  requested_context_mib=4096  actual_32bit_context_mib=0
frames=897  requested_context_mib=4100  actual_32bit_context_mib=4
frames=1024 requested_context_mib=4608  actual_32bit_context_mib=512
```

Instead of silencing the narrowing warning with a cast, I’d suggest making the error explicit. The code should check that the computed context size fits in `size_t` before initializing `ggml_init_params`, and throw a clear error if it does not.

mwzkhalil:
Reviewed the change: hoisting `context_size` into an explicit `static_cast<size_t>(...)` before the `ggml_init_params` aggregate init is the right fix — the narrowing was specifically about brace-initializing a `size_t` field from a `ull`-typed expression the compiler can't prove fits, which `-Wnarrowing`/`-pedantic-errors` correctly rejects; a plain assignment to a same-typed local doesn't have that restriction. Confirmed this doesn't change behavior on 64-bit targets (same value, just computed outside the brace-init), and CI is green across all platforms. Looks correct to me.

weidu-tpvision:
Hi guys, thanks for the review. @0xShug0 I think the problem is 4096M is the max size of 32 bit so it is warped to 0, static-cast is fine but indeed there is no size check.

I had the same size_t problem with litert and something similar with onnxruntime, it shows that 32 bit is less and less supported. We have a armv8 cpu but we currently build in user space with armv7. I had whisper.cpp and I will propose to swtich to this one, very cool and thanks!

0xShug0:
@weidu-tpvision Thanks! PR Merged.

PR #513, opened 2026-09-11
fix(models): prevent size_t overflow on 32-bit targets in model confi…

weidu-tpvision:
On 32-bit architectures, size_t is 32-bit (SIZE_MAX = 4 GB).

Default weight context sizes exceeding 4 GB (such as 4096 MB, 6144 MB, and 8192 MB) overflow size_t, wrapping around to 0 or 2 GB and triggering compiler warnings (-Woverflow) or zero-byte allocations.

Add pointer-width preprocessor guards to cap default weight context sizes to safe thresholds on 32-bit targets while preserving original defaults on 64-bit systems.


mwzkhalil:
Reviewed this one too: capping the 32-bit defaults (1024MB/2048MB) well under the 4 GiB `size_t` ceiling — with real margin for the rest of the process's address space rather than cutting it as close as possible — looks like the right call rather than just silencing the overflow. Same underlying issue as #512 (aggregate-init narrowing under `-Wnarrowing`/`-pedantic-errors` on 32-bit `size_t`), and the 64-bit defaults are untouched. CI is green across the board. Looks correct to me.

0xShug0:
@weidu-tpvision I think the changes may be no-ops because the values are also set in .cpp? Let me know if there will be follow-up PRs.

weidu-tpvision:
Ok, didn't pay attention to those consts in cpps, I did another pr but I don't know whether it is useful. The compiling warnings in headers bothered me so I made a pr. In reality, you never run a model larger than 1B on a 32-bit system, so maybe it is easier to just forbid to build for some models.

PR #521, opened 2026-09-12
feat(server): decode MP3/FLAC uploads and reject unsupported audio 

### LINKED ISSUE #313: [Feature Request / Bug]: Support non-WAV / browser audio uploads (WebM/Opus, MP3, OGG) on `/v1/audio/transcriptions` for Open WebUI compatibility

mwzkhalil:
…eanly

/v1/audio/transcriptions and /v1/audio/alignments only accepted RIFF/WAVE bytes, gated by a filename-extension check that trusted client-supplied names like "blob" or "audio.webm". A non-WAV upload either got silently waved through the gate and then threw deep inside the WAV parser, or hit the WAV-only 400 if its extension matched but its bytes didn't - and either way, any exception the decode/build step threw was uncaught at this layer, so it fell through to the generic top-level handler and came back as a 500 server_error instead of a 400 invalid_request_error.

This adds engine::audio::decode_audio_upload_f32(), which sniffs the upload's actual bytes instead of trusting its filename: WAV goes straight to the existing read_wav_f32() unchanged, MP3/FLAC decode through a new decode-only vendored miniaudio build, and Ogg/WebM (what browsers record microphone input as via MediaRecorder) get a specific error naming the container and pointing at an ffmpeg transcode instead of an opaque failure - full Opus decoding needs a codec backend this change doesn't add. Both multipart handlers now also catch decode errors locally and map them to a 400, matching the pattern already used elsewhere in this file.

Fixes #313


0xShug0:
@mwzkhalil Thanks for the PR! The issue has been addressed in https://github.com/0xShug0/audio.cpp/pull/330.

Issue #471, opened 2026-09-06
Higgs_Audio_TTS voice cloning not working and Higgs TTS generation reached max_tokens before EOC (Arabic, CPU)


user6654126:
**Environment:**

OS:Linux Mint 22.3 - Cinnamon 64-bit
CPU:AMD Ryzen 5 8400F
GPU: RTX 5060 8G VRAM
RAM: 6000MHz 16GB RAM 
Model: higgs-audio-v3-tts-4b-q8_0.gguf
version: latest v0.7.2
 * branch            main       -> FETCH_HEAD
   bf3315f..49b0322  main       -> origin/main
Updating bf3315f..49b0322

****Symptom**:**

when i use this model with backend cpu sometime it work , if the TTS script is small meaning less than 2 minutes.
but the problem it doesn't even use reference voice in voice cloning instead it use different and random voices and i noticed it switch into different voice after whatever limit i set in   --text-chunk-size 300 \
so how i make it use the reference voice for voice cloning

**Commands arabic test:**

./build/linux-cpu-release/bin/audiocpp_cli \
  --task tts \
  --family higgs_audio_tts \
  --session-option "higgs_audio.mem_saver=true" \
  --model /home/xxxx/audio.cpp/build/linux-cpu-release/bin/models/Higgs-Audio-v3-TTS-4B-GGUF/higgs-audio-v3-tts-4b-q8_0.gguf \
  --backend cpu \
  --threads 10 \
  --text "$(cat /home/xxxx/audio.cpp/inputs/gemini-arabic-mini.txt)" \
  --voice-ref /home/xxxx/audio.cpp/inputs/stoic_10seconds.wav \
  --reference-text "$(cat /home/xxxx/audio.cpp/inputs/stoic_10seconds.txt)" \
  --out /home/xxxx/audio.cpp/outputs/higgs_ara_1.wav \
  --text-chunk-size 300 \
  --metrics \
  --log

**Logs:**
[higgs_ara_1 log.txt](https://github.com/user-attachments/files/31872602/higgs_ara_1.log.txt)

also if i removed   --session-option "higgs_audio.mem_saver=true" \ and   --text-chunk-size 300 \ i get  Higgs TTS generation reached max_tokens before EOC
and if script is long +5 minutes i get EOC so how to fix that

**Commands arabic test:**

./build/linux-cpu-release/bin/audiocpp_cli \
  --task tts \
  --family higgs_audio_tts \
[...]

PR #522, opened 2026-09-12
Improve Higgs TTS token-limit error guidance

mwzkhalil:
## Change

Improve the error reported when a Higgs TTS text chunk reaches max_tokens before EOC:

- Include the configured token limit.
- Clarify that the limit applies to the current text chunk.
- Point users to --max-tokens / max_tokens and --text-chunk-size / text_chunk_size.

The failure condition and generation behavior are unchanged. The proposed seed change has been removed; this PR does not claim to fix voice drift in #471.

## Validation

Reviewed the final diff: only the error message in generator.cpp changes. git diff --check passes. No new inference tests were run for this message-only change.

The original author commit is retained; the scope reduction is a maintainer follow-up commit.


0xShug0:
@mwzkhalil Unless I’m missing something, setting the seed won’t fix this issue. Please provide wavs generated by higgs tts on CPU with your PR applied using the test case provided in the issue.

0xShug0:
@mwzkhalil I’ve kept the error-message improvement, thanks! A follow-up PR for the seed change would be welcome, with logs showing that it helps resolve the voice drift.

PR #526, opened 2026-09-12
perf(vibevoice_asr_streaming): flash suffix attention and bounded history window

LysanderdeJong:
Streaming `vibevoice_asr_streaming` (Q8, one 16 GB GPU) stops on long audio for two reasons. The suffix graph uses approximately 4.8 GiB of temporary data at 1024 cache steps in addition to approximately 9.9 GiB of weights (memory becomes full after approximately 35 s). Then the decoder history increases without limit toward the 131072-position limit. This PR removes both causes. It makes no engine changes for other families. All new functions are off by default.

## Change 1: suffix path uses prefix-flash attention

The suffix graph is the only path that adds data onto the prefix KV cache. `make_qwen_decoder_config` selects flash modes for prefill and cached-step graphs, but it leaves the default `prefix_mode=Exact`. As a result, the suffix graph uses the eager branch. `FlashWithPrefix` (same as higgs/personaplex) sends it through `flash_attention_from_grouped_heads_view_kv`. The suffix graph decreases from 4.8 GiB to 1.1 GiB at 1024 steps. The increase rate decreases from approximately 4.7 to approximately 0.4 MiB/step.

## Change 2: `max_history_steps` rolling window (default `0` = off)

New session option (`vibevoice_asr_streaming.max_history_steps`, `vibevoice_asr.` prefix also accepted). When the option has a value, the program allocates the cache and the graphs one time at the window size. It never increases them. New steps replace the oldest entries in their positions:

- General ring support in `TransformerKVCache` (`ring_mode` +
`ring_pinned_steps`, off by default): wrap on advance, chronological export, wrap placement on import, `slot_for_position` mapping, shared `ring_stored_position` rule. No existing method changes function unless `ring_mode` is true.
- The streaming prompt has a pin and the program never removes it.
Removal of the prompt first causes the output to collapse into repetition loops (measured). Positions continue to increase. The mask uses stored counts.
- `tests/unittests/test_transformer_kv_ring.cpp`: mapping, wrap
round-trip, advance, option validation, non-ring regression.
- Documentation, subsection "Long streams and the history window", with
use and size data.

This PR also corrects the suffix causal-triangle mask. The mask was at key columns `[cache_steps_, …]` (padding). The real suffix slots had no mask to each other. The new code puts the mask on the actual appended slots. This agrees with the reference (`modeling_vibevoice_asr.py`). Transcriptions show small changes versus #497 even with the window off.

## Outputs / performance / memory

- Memory: the suffix graph is smaller in all runs. With a window, memory
use is flat. A 2 h 40 m file uses approximately 9.9 GB for 24 min wall time on a 16 GB card. Without a window, memory use exhausts the card. With the window off, the program uses memory as before.
- Outputs: flash and eager calculations can show very small differences.
The mask correction causes small changes in transcriptions. The window changes outputs when removal of old steps starts (expected). Runs give the same results for a fixed binary and window (identical rerun verified). For reproducible runs, do not change the binary or the window size.
- Performance: no change (43 s with window versus 44 s without window on
a 300 s slice, approximately 7 times real time; chunk delay approximately 0.5 s against a 4.4 s chunk).

## Families / routes checked

- `vibevoice_asr_streaming`: full list below.
- `parakeet_tdt` (other family, same framework and CLI paths): golden
clip is identical to the golden transcription on the rebuilt tree.
- `transformer_kv_ring_test` has a pass in ctest (includes non-ring
regression).
- Full-tree Release build: `audiocpp_cli` and `audiocpp_server` build.

## Validation (RTX 4070 Ti SUPER, CUDA, Q8, `--device 1`)

| Run | Result |
|---|---|
| Golden clip (7.4 s EN) | Correct transcript with `Speaker 0:` labels |
| 6 s sine / 6 s silence | `[Music] [Music]` / `[Silence] [Silence]`, exit 0 |
| 60 s file, window 256 (wrap torture) | Exit 0, normal output |
| 300 s file, window 4096 vs uncapped | Same structure and labels, only local variants |
| Identical rerun | Identical result (deterministic) |
| **2 h 40 m file, window 4096** | **Exit 0, 4185 lines, 5 speaker labels, memory flat at ~9.9 GB** |
| `max_history_steps=-5` / `> 131072` | Clear startup errors |

Sizing: approximately ten steps per audio second. Thus `4096` stores approximately ten minutes of context. Stable state is approximately 10.0 GB on Q8/CUDA.

## Known limits (nothing pending release; model stays as merged in #497)

- No ground-truth WER exists for the long file. Structural checks and
the comparisons above are the quality evidence.
- Positions can increase past the trained 131 k cap on very long streams
(same as the reference). The window limits memory, not positions.
- `--turns-out` gives no output in CLI streaming mode (existing
function).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2699,6 +2699,9 @@ if (ENGINE_BUILD_TESTS OR ENGINE_BUILD_EXTENDED_TESTS OR ENGINE_BUILD_MODEL_TEST
             tests/personaplex/personaplex_ring_kv_cache_parity.cpp
         )
         target_link_libraries(personaplex_ring_kv_cache_parity PRIVATE engine_runtime ggml)
+
+        add_engine_unittest(transformer_kv_ring_test tests/unittests/test_transformer_kv_ring.cpp)
+        add_test(NAME transformer_kv_ring_test COMMAND transformer_kv_ring_test)
         if (ENGINE_ENABLE_OPENMP)
             target_link_libraries(personaplex_ring_kv_cache_parity PRIVATE OpenMP::OpenMP_CXX)
         endif()
--- a/docs/models/vibevoice_asr.md
+++ b/docs/models/vibevoice_asr.md
@@ -219,3 +219,36 @@ Common request options:
 | `--repetition-penalty` | float | `1` | Generation repetition penalty. |
 | `--audio-chunk-mode` | `auto`, `fixed`, `vad`, `none` | `auto` | Offline audio chunking mode. |
 | `--audio-chunk-seconds` | float seconds | `1200` | Offline chunk duration for fixed and VAD chunking. |
+| `--session-option vibevoice_asr_streaming.max_history_steps=<n>` | integer | `0` (uncapped) | Rolling window for decoder history, in steps. Refer to [Long streams](#long-streams-and-the-history-window). |
+
+### Long streams and the history window
+
+By default, the decoder stores the full conversation history. Memory use
+increases with stream length (approximately 0.4 MiB per step on Q8/CUDA).
+On very long streams, the GPU memory becomes full and transcription stops.
+
+Set `max_history_steps` to prevent this. The model then stores only the
+most recent steps. Memory use remains constant for streams of any length.
+
+```bash
+audiocpp_cli --task asr --family vibevoice_asr_streaming \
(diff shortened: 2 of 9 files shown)


0xShug0:
@LysanderdeJong Merged, thanks! 

One issue I found is with the test integration: transformer_kv_ring_test initializes the CPU backend directly through GGML, which fails in the normal dynamic-backend debug build. It should use audio.cpp’s backend initialization path instead. I’ll do the follow-up myself since I’m planning to have a release today :)

PR #528, opened 2026-09-12
fix(pocket-tts): make reference audio fields visible again

SmilingWolf:
Pretty much what the commit message says.

I have only tested the change with pocket-tts, I don't have Kokoro or other possibly affected models around to try.


0xShug0:
@SmilingWolf Thanks for the fix! PR merged.

Issue #525, opened 2026-09-12
[Proposal] A stable C ABI (`audiocpp.h` + optional shared library) for in-process embedding


christopherthompson81:
## Summary

audio.cpp currently exposes three surfaces: `audiocpp_cli`, `audiocpp_server`, and the WebUI. All three are
process-level. There is no way to embed the engine **in-process** in a non-C++ application, because
`include/engine/**` is C++ headers in `engine::runtime` (templates, `std::optional`, `std::filesystem`,
virtual interfaces, exceptions), and `CMakeLists.txt:297` force-sets `BUILD_SHARED_LIBS OFF` outside the
CPU-all-variants path.

I'd like to propose adding a **thin, additive C ABI** over the existing runtime interfaces, plus an
off-by-default CMake option that builds it as a shared library. This is the "app-facing workflows" item in
CONTRIBUTING.md, approached from the embedding side rather than the serving side.

The runtime is already shaped almost perfectly for a C facade, which is what makes me think it's worth
proposing rather than just forking:

1. **Options are already `std::unordered_map<std::string, std::string>`** — in `ModelLoadRequest::options`,
   `SessionOptions::options`, and `TaskRequest::options`. The C layer needs a string→string bag and nothing
   else. It carries **zero per-family knowledge**, so adding a model family never touches the ABI.

2. **The task surface is already generic.** `TaskRequest` → `TaskResult` covers audio in/out, text,
   `speech_segments`, `speaker_turns`, `word_timestamps`, and artifacts for *every* family. There are
   exactly two run interfaces to wrap: `IOfflineVoiceTaskSession::run` and the
   `IStreamingVoiceTaskSession` chunk/finalize pair.

3. **It's already self-describing.** `ModelInspection::cli` carries `request_options`, `session_options`,
   and `load_options` as `CliOptionInfo` (name, value name, description, required, default, min, max). If
   the C API exposes that, a binding can enumerate a model's options **at runtime** and generate typed
   wrappers from it — instead of hardcoding per-family knowledge that rots.

4. **Streaming is already pull-based.** `next_stream_event()` avoids callbacks across the FFI boundary,
   which is the part that's genuinely painful from GC'd runtimes.

So the shim is a facade over `ModelRegistry` → `ILoadedVoiceModel` → `IVoiceTaskSession`, not a new
abstraction. On the order of 50 functions, one translation unit, no changes to `src/framework` or
`src/models`.

## Sketch

```c
[...]

PR #530, opened 2026-09-12
Add an opt-in C ABI for in-process embedding

christopherthompson81:
Implements the C ABI proposed in #525 — an in-process entry point for applications
that embed audio.cpp rather than shelling out to the CLI or running the server.

One commit, 10 files, +3565/-0. `AUDIOCPP_BUILD_C_API` defaults to OFF; with it off,
configure produces no new targets, no tests, and no changed flags.

| | |
|---|---|
| `include/audiocpp.h` | the C header — C99, 69 entry points |
| `src/capi/audiocpp.cpp` | the facade — one translation unit |
| `src/capi/audiocpp.map` / `.symbols` | ELF / Mach-O export lists |
| `tests/capi/` | four tests, the C ones compiled as C |
| `docs/c_api.md` | ownership and lifetime rules |

## Design

A translation layer over `ModelRegistry` → `ILoadedVoiceModel` → `IVoiceTaskSession`,
the same surfaces `audiocpp_cli` uses. It depends only on framework headers, never on
`app/`, so the CLI and the C API stay independent.

Three properties of the runtime keep it thin. Options are already
`unordered_map<string, string>` in `ModelLoadRequest`, `SessionOptions` and
`TaskRequest`, so the ABI carries no per-family knowledge and **adding a model family
never changes the header**. `TaskRequest` → `TaskResult` is already generic, so there
are two run paths to wrap. And `ModelInspection::cli` is already self-describing, so
it is exposed as-is: a binding enumerates a family's options at runtime — names,
types, defaults, bounds — instead of hardcoding them.

Opaque handles only; no exception escapes; accessors rather than struct copies so
`TaskResult` can gain fields without breaking the ABI; returned strings are never
NULL. Handles hold their parents alive, so freeing out of order is safe — callers
from garbage-collected runtimes cannot control finalizer order.

## Two notes against the proposal

**The surface is 69 functions, not the ~50 sketched in #525**, and includes streaming,
artifacts and style conditioning rather than deferring them. It now covers everything
`TaskRequest`, `TaskResult`, `ModelLoadRequest` and `CliOptionInfo` carry, so a caller
is never pushed back to the CLI mid-workflow. Happy to split the offline core out if
you would rather review that first.

**One line falls outside the new files**: `cjson_vendor` is now compiled with
`CJSON_HIDE_SYMBOLS`. cJSON annotates its public functions with
`__declspec(dllexport)` by default, and since `dllexport` is additive, any DLL
absorbing that static library re-exports all 78 — the first Windows build produced 147
exports against 69 intended. No DLL is built from this tree today, so this was latent;
this is simply the first target to expose it.

## Validation

| platform | |
|---|---|
| Linux x86_64 | CPU and CUDA (RTX 3090, CUDA 12.8); single- and multi-config generators |
| macOS arm64 | CMake 3.31.12 and 4.4.3, Metal on and off, Xcode/clang 21 |
| macOS x86_64 | links, exports verified |
| Windows x64 | MSVC 14.44.35207, SDK 10.0.26100, VS 2022 generator |

Every platform produces the same results for the same inputs.

- **`audiocpp_c_api_path`** — the ABI contract, offline and streaming, against the
  Silero VAD model and speech clip already in the repo. No download, no network.
- **`audiocpp_c_api_model`** — five families across five task types, covering every
  result accessor: Kokoro (audio, 12 declared options), Citrinet and Parakeet
  (transcript, 28 word timestamps), Sortformer (5 speaker turns), BS-RoFormer (2 named
  streams).
- **`audiocpp_c_api_parity`** — the same inputs through `audiocpp_cli`, compared. The
  C API drives the same runtime, so it must produce not merely plausible output but
  the *same* output. Passes across all five families. Kokoro draws noise per run, so
  both sides are seeded.
- **`audiocpp_c_api_exports`** — asserts the library exports exactly this header, on
  all three platforms. 69 declared, 69 exported, nothing else.

The last three report SKIP (exit 77) when models or tooling are absent, so a checkout
with no downloads stays green.

Session reuse is the reason to embed rather than shell out, so: one session, repeated
requests, RSS sampled after warm-up — 1072 kB delta at 200, 500 and 1000 iterations.
Flat, not linear. 40.5 ms per request on CPU (347× realtime, 14 s of audio per
request).

A C# binding over this header exists in my fork
(christopherthompson81/audio.cpp#2) and is **not** proposed here — #525 scopes
language bindings out. It is worth mentioning only as validation: an independent
consumer, in another language, exercising the out-of-order-free guarantee that
motivated the design, since .NET finalizer order is not deterministic. Its tests and
the C tests report identical values across all five families — 11 lines,
byte-identical — on Linux, macOS and Windows. It also confirms the cJSON change
removed nothing a real consumer needs: every `audiocpp_*` symbol resolves through
P/Invoke at runtime.

## Notes for review

- `CMAKE_POSITION_INDEPENDENT_CODE` is set inside `if (AUDIOCPP_BUILD_C_API)` only —
  `BUILD_SHARED_LIBS` is forced OFF at `CMakeLists.txt:297`, so the objects are
  otherwise non-PIC. Enabling the option forces a full ggml rebuild.
- `enable_testing()` is called in the C API block. The existing call is further down
  and gated behind `ENGINE_BUILD_*`, so an `add_test()` here would silently never
  register.
- The backend-name mapping is duplicated from `app/cli/args.cpp`, to keep the C API
  independent of `app/`. Glad to lift it into the framework instead.
- The library does not call `omp_set_num_threads()`, unlike the CLI — that is
  process-global state a library shouldn't touch. Threads come from
  `audiocpp_backend_config`.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,9 @@ option(ENGINE_BUILD_TESTS "Build framework unit tests" OFF)
 option(ENGINE_BUILD_EXTENDED_TESTS "Build extended non-model tests and probes" OFF)
 option(ENGINE_BUILD_MODEL_TESTS "Build model-specific tests and probes" OFF)
 option(ENGINE_BUILD_WARMBENCH "Build warmbench helper binaries" OFF)
+option(AUDIOCPP_BUILD_C_API "Build the C ABI (include/audiocpp.h) as a shared library" OFF)
+set(AUDIOCPP_C_API_MODEL_ROOT "" CACHE PATH
+    "Models root for the C API model test. Unset leaves the test registered but skipping.")
 set(AUDIOCPP_GGML_SOURCE_DIR
     "${CMAKE_CURRENT_SOURCE_DIR}/external/ggml"
     CACHE PATH "ggml source tree to build with AudioCPP")
@@ -282,6 +285,12 @@ if (ENGINE_ENABLE_CUDA AND NOT ENGINE_ENABLE_HIP)
 endif()
 if (ENGINE_ENABLE_OPENMP)
     find_package(OpenMP REQUIRED COMPONENTS CXX)
+endif()
+
+if (AUDIOCPP_BUILD_C_API)
+    # engine_runtime and ggml are static/object libraries here; linking them into
+    # a shared libaudiocpp requires position-independent code throughout.
+    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 endif()
 
 if (ENGINE_ENABLE_CPU_ALL_VARIANTS)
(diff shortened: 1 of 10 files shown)


0xShug0:
@christopherthompson81 Thanks! No blockers found. I will merge it after the new release.

christopherthompson81:
Would you be willing to take the C# bindings, or do you want those to be external?

0xShug0:
@christopherthompson81 I’d prefer to keep language bindings as a separate repo/package for now rather than inside the main audio.cpp repo.  The main repo should stay focused on the C++ runtime. Also, I don’t think I currently have the bandwidth or expertise to maintain that properly within the core repo. The language bindings could follow a structure similar to https://github.com/0xShug0/audio.cpp-server-frontends, staying well decoupled from the main repo. If you create a separate repo for the bindings, I’m happy to link to it from audio.cpp docs, and we can consider adding it as a submodule if tighter integration becomes useful.

0xShug0:
@christopherthompson81 Merged :tada:  Thanks!

PR #536, opened 2026-09-13
parakeet_tdt: size the encoder graph for the path run() will take

christopherthompson81:
## Problem

`prepare()` chooses between full-context and bounded-window capacity from `offline_mode` alone, but `run()` also takes a bounded path whenever `audio_chunk_mode` is `vad` or `fixed`. The two disagree, so a recording long enough to need chunking gets a graph sized for the whole thing and throws `Parakeet TDT encoder relative position frames exceed maximum`.

The throw originates inside `prepare()` — `prepare_capacity()` builds the relative positional encoding eagerly — so over-sizing fails immediately, while under-sizing is harmless because the graph grows at run time. Only the one direction is a defect.

The effect is that `audio_chunk_mode=vad` and `=fixed` do not work on the inputs they exist for, unless the caller also sets `offline_mode=long_form`. That combination happens to make `prepare()` agree, but it is not discoverable and not what the option means.

## Fix

`prepare()` mirrors the dispatch in `run()`.

Two details the mirror has to get right:

- A `vad` window is capped at `audio_chunk_duration_sec`, which can exceed the long-form window — the contexts total 14s by default and the duration is settable past that — so `vad` capacity takes the larger of the two, resolving the duration with the same request-then-session precedence `run_vad_chunks()` uses.
- `quiet_energy` is sized bounded even though `run()` rejects it, so `prepare()` does not pre-empt that rejection with a ceiling error about input `run()` will never encode.

## Measurements

One tree, one binary swapped between stock and patched. 10-minute conversational recording, q8_0 Parakeet GGUF, CUDA, `offline_mode` left at its default:

| `audio_chunk_mode` | before | after |
|---|---|---|
| `vad` | throws | 4885 ms, 1483 words |
| `fixed` | throws | 6881 ms, 1372 words |
| `none` | throws | throws — correct |
| `quiet_energy` | rejected with the mode error | rejected with the mode error |
| `vad` + `offline_mode=long_form` | 4852 ms, 1483 words | 4959 ms, 1483 words |

`none` still throwing is the intended outcome: chunking is off and the clip genuinely exceeds the encoder's ceiling.

Unchanged where it should be. A 90s clip that already fit gives the same 226 words (`vad`) and 236 (`none`) either way, and `vad` with `offline_mode=long_form` gives the same 1483 words.

## Reproducing

Load a Parakeet GGUF, create an offline `asr` session leaving `parakeet_tdt.offline_mode` at its default, and run a clip longer than the encoder ceiling — 5000 frames of `max_position_embeddings` at 80ms per frame after the 8x subsampling, so about 400s — with `audio_chunk_mode=vad` and `audio_chunk_duration_sec=10` on the request. Before this change that throws; after it, it transcribes.

No test is included: the suite here is weight-dependent warm benches rather than unit tests of `prepare()`, so there is no cheap place for a regression test that does not pull a checkpoint.

### DIFF
--- a/src/community_models/parakeet_tdt/session.cpp
+++ b/src/community_models/parakeet_tdt/session.cpp
@@ -345,14 +345,47 @@ void ParakeetTDTOfflineSession::prepare(const runtime::SessionPreparationRequest
     int64_t capacity_samples = request.audio->max_input_samples;
     int capacity_channels = request.audio->channels;
     int capacity_sample_rate = request.audio->sample_rate;
-    if (offline_mode_ == "long_form" ||
-        (offline_mode_ == "auto" &&
-         request.audio->max_input_samples / std::max(request.audio->channels, 1) >
-             auto_full_context_max_samples_)) {
+
+    // Mirror the dispatch in run(). vad and fixed take a bounded-window path
+    // whatever offline_mode says, so sizing the graph from offline_mode alone
+    // leaves prepare() and run() disagreeing: a long clip with
+    // audio_chunk_mode=vad gets a graph built for the whole recording and
+    // fails the encoder's relative-position ceiling before a chunk is cut.
+    const auto chunk_mode = engine::audio::parse_audio_chunk_mode(request.options);
+    const bool bounded_windows =
+        chunk_mode == engine::audio::AudioChunkMode::Vad ||
+        chunk_mode == engine::audio::AudioChunkMode::Fixed ||
+        // run() rejects quiet_energy outright. Size it bounded so prepare() does
+        // not pre-empt that with a ceiling error about input run() never encodes.
+        chunk_mode == engine::audio::AudioChunkMode::QuietEnergy ||
+        (chunk_mode == engine::audio::AudioChunkMode::Auto &&
+         (offline_mode_ == "long_form" ||
+          (offline_mode_ == "auto" &&
+           request.audio->max_input_samples / std::max(request.audio->channels, 1) >
+               auto_full_context_max_samples_)));
+
+    if (bounded_windows) {
         capacity_samples =
             left_context_samples_ + center_samples_ + right_context_samples_;
         capacity_channels = 1;
         capacity_sample_rate = assets_->config.frontend.sample_rate;
+
+        // A vad window is capped at audio_chunk_duration_sec, which is free to
+        // exceed the long-form window: the contexts default to 14s in total and
+        // the chunk duration is settable well past that.
+        if (chunk_mode == engine::audio::AudioChunkMode::Vad) {
+            auto chunk_seconds =
+                engine::audio::parse_audio_chunk_seconds_override(request.options);
+            if (!chunk_seconds.has_value()) {
+                chunk_seconds =
+                    engine::audio::parse_audio_chunk_seconds_override(this->options().options);
+            }
+            if (chunk_seconds.has_value() && *chunk_seconds > 0.0F) {
+                capacity_samples = std::max(
+                    capacity_samples,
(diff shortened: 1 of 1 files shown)


0xShug0:
@christopherthompson81 Found one regression:

420-second recording, 10-second VAD limit: previously failed; now works. :tada: 
10-second recording, 600-second VAD limit: previously worked; **now fails because preparation sizes for 600 seconds.**

I think the issue is the PR tries to prepare for 600 seconds of audio even when the recording is only 10 seconds long. It uses configured chunk-duration limit without capping it by the actual input length.

christopherthompson81:
Thank you — reproduced exactly as you described, and your diagnosis was right. Fixed in `ad0a525`.

`audio_chunk_duration_sec` is a ceiling on a span, not a promise of one. Spans are cut from the recording, so none can be longer than the recording is, and sizing `prepare()` from the configured duration alone built a 600-second graph for a 10-second clip. The window is now bounded by the input length.

## That alone was not enough

A 420-second recording with a 600-second limit is still over the encoder's ceiling after the input bound — but it is cut into spans that fit, and `run()` transcribes it. Failing in `prepare()` would reject input `run()` handles.

So on a bounded path the prepared capacity is also clamped to `max_position_embeddings`. `prepare_capacity()` builds the relative positional encoding eagerly, so over-sizing throws immediately while under-sizing only grows the graph at run time — the clamp can never turn a working case into a failing one.

Full-context sizing is deliberately untouched: there the whole recording really is encoded in one pass, so the early, clear error is the right one.

## Validation

CPU, q8_0 Parakeet GGUF, `audio_chunk_mode=vad`. A 10s clip cut from `assets/resources/sample_16k.wav`; a 420s clip built by repeating it with 1.5s silences between.

```
                                     base    this PR   with ad0a525
  10s clip, limit 600s                ok      FAIL         ok
  420s clip with gaps, limit 600s    FAIL     FAIL         ok          756 words
  420s clip, limit 10s               FAIL      ok          ok          830 words
  10s clip, no chunking               ok       ok          ok           19 words
  420s clip, no chunking             FAIL     FAIL        FAIL
```

The last row is deliberate — chunking is off and the clip exceeds what the encoder can represent in one pass.

Note the second row: the gapped 420s case at a 600s limit **did not work before this PR either**. The clamp fixes it rather than only restoring previous behaviour.

Build and run:

```
cmake -S . -B build-test -DCMAKE_BUILD_TYPE=Release -DGGML_CUDA=OFF -DAUDIOCPP_BUILD_TESTS=OFF
cmake --build build-test --target audiocpp_cli -j 6

./build-test/bin/audiocpp_cli --task asr --family parakeet_tdt \
  --model <parakeet-tdt-0.6b-v3-q8_0.gguf> --backend cpu \
  --audio <clip.wav> \
  --request-option audio_chunk_mode=vad \
  --request-option audio_chunk_duration_sec=600
```

I built each of `base`, this PR, and the fix in turn and ran the same matrix against each, rather than only checking the fix passes.

## One case that is not a regression

A 420-second clip of **continuous** speech (no silence to split on) with a 600s limit fails on every variant, including base. The VAD returns one span covering the whole recording — `parakeet.frontend.frames 41984` — and no prepare-time sizing can change that. I mention it because my first 420s clip was built without gaps and produced exactly this, which briefly looked like the clamp not working.

Removing the clamp and keeping only the input bound fails the gapped 420s case, so both bounds are load-bearing.

0xShug0:
@christopherthompson81 Could you remove the unrelated C# changes (71 files) from the PR? Then we are good to go.

0xShug0:
@christopherthompson81 Merged. Thanks!

PR #540, opened 2026-09-13
vulkan-shaders-gen: retry an empty shader compile instead of failing at link (potential fix for CI failures)

christopherthompson81:
## Problem

A shader whose compile produces no SPIR-V becomes an undefined reference at link, ~20 minutes into the build, naming a symbol in vendored code with nothing earlier in the log to explain it. The Nix (vulkan) job on #539 failed this way:

```
ld.bfd: ggml-vulkan.cpp:3964: undefined reference to
  `matmul_id_subgroup_nvfp4_f32_aligned_f16acc_cm1_len'
```

`write_output_files()` writes the declaration before it reads the artefact, then skips the definition if the file is empty, with no diagnostic:

```cpp
hdr << "extern const uint64_t " << name << "_len;\n";      // always declared
...
std::string data = read_binary_file(path);
if (data.empty()) {
    continue;                                               // definition skipped
}
```

Compile success was judged by stderr alone, with no exit code, so a shader that merely warns is discarded while one that reports nothing and writes nothing is accepted.

## Fix

Success is judged by the artefact — the SPIR-V must exist and be non-empty. An empty result is retried up to three times with short backoff; after the last attempt the shader is named and generation fails. Diagnostics alongside a usable artefact are treated as warnings and the shader is kept.

## Why retry rather than just fail

The trigger does not reproduce locally. Investigating the #539 failure I built shaderc 2026.1 — the version this repo's Nix build uses, per the job log — and compiled the exact failing shader with the exact 17 defines: exit 0, no stderr, 27,700 bytes of valid SPIR-V. 120 repeat compiles under parallel load produced no anomalies. A full Vulkan build with that compiler and all four shader extensions enabled links cleanly: 2320 symbols declared, 2320 defined.

So the shader is fine and the compiler is fine; something environmental produced an empty file while reporting success. Retrying absorbs that. Failing loudly covers the case where it is not transient.

## Verification

Against an injected failure, both directions:

- transient (first attempt empty, later attempts normal) — recovers, build links, exit 0:

```
shader matmul_id_subgroup_nvfp4_f32_aligned_f16acc_cm1 produced no SPIR-V; retrying (2/3)
[333/333] Linking CXX executable bin/audiocpp_cli
```

- permanent (always empty) — stops at generation naming the shader, exit 1, no link error:

```
shader matmul_id_subgroup_nvfp4_f32_aligned_f16acc_cm1 produced no SPIR-V; retrying (2/3)
shader matmul_id_subgroup_nvfp4_f32_aligned_f16acc_cm1 produced no SPIR-V; retrying (3/3)
cannot compile matmul_id_subgroup_nvfp4_f32_aligned_f16acc_cm1 after 3 attempts
shader generation failed; see errors above
```

## Note on upstream

The same declaration-before-definition path exists in ggml-org/llama.cpp master. Upstream does already check the exit code from `execute_command`, which this tree's copy does not — worth a look when these files are next synced.


0xShug0:
@christopherthompson81 Thanks! I’ll keep it open for now until I get some feedback from the nix sub.

christopherthompson81:
Moving this to draft so it isn't sitting in your review queue while you wait on feedback from the nix sub — no pressure either way, just ping me and I'll mark it ready again.

0xShug0:
Got nothing from nix sub. I'm going to merge this PR.

0xShug0:
@christopherthompson81 I don’t know much about CI, so I left the review entirely to Codex. Does that make sense?

> Two issues
> - It can accept a failed compile. The PR treats any non-empty .spv file as success without checking the compiler exit code. A failed compile could leave a non-empty stale file from an earlier build. 
> - Its failure flag has a data race. Shader jobs run concurrently, but generation_failed is a plain bool written by those jobs. llama.cpp uses an atomic flag. 
> 
> Current llama.cpp checks the subprocess exit status and fails shader generation on an error; it does not implement the PR empty-file retry. The retry may be useful, but it should be added on top of reliable subprocess failure handling, with the old output removed before each attempt

christopherthompson81:
It makes sense — both points were right, and I've pushed a fix for each in ae72096.

**Exit code.** Worse than it looked: `execute_command` in this tree discards the status at the syscall (`waitpid(pid, nullptr, 0)`, and no `GetExitCodeProcess` on the Win32 side), so no caller could have checked it. It returns the status now. A non-zero exit or any stderr fails the shader immediately, with no retry — a compiler that reports failure gets believed the first time. Only a silent empty result is retried, and `out_path` is removed before each attempt so a stale artefact can't stand in for a missing one.

The stale-file case wasn't hypothetical. Against a fake `glslc` that exits 0, prints nothing and writes nothing, with a 26-byte leftover `.spv` already in the output directory, the old binary exited 0 and embedded the stale bytes as the shader. New one stops at generation and names it.

**Data race.** Correct — `generation_failed` is `std::atomic<bool>` now. I also set it in the `catch` handler, which was a third way to the same undefined reference: an exception there leaves the shader out of `shader_fnames` entirely.

On the ordering point — "the retry should be added on top of reliable subprocess failure handling" — that's exactly the shape now: exit code first, artefact second, retry only on the silent-empty case.

Verified standalone over `add.comp`'s 16 variants: healthy exits 0 clean; non-zero exit fails at once naming the exit code; transient first-attempt-empty recovers with all 16 definitions emitted; permanent-empty and stale-file both stop at generation with exit 1.

And yes — llama.cpp master has the exit-code check but not the empty-file retry. I have the retry on a local branch there; I'll send it upstream once this one settles, so the two trees don't drift further.

christopherthompson81:
Two follow-ups — the branch has been force-pushed, so the `ae72096` in my last comment is now `f031d28`. Same content, rebuilt; see the second point for why.

**One more backport from llama.cpp master.** The failure check belongs *before* `write_output_files()`, not after it. Writing a partial header and source gives them fresh timestamps, so a second build finds them newer than their inputs, skips regeneration and links the gap — the original failure mode, one build later. Upstream returns early there; this branch now does too.

I kept the check after `write_output_files()` as well, because the embed step has its own way to fail. Worth flagging upstream: llama.cpp master sets `compile_failed` inside `write_output_files()` and then returns `EXIT_SUCCESS` without re-reading it, so that one is detected and dropped. Backporting the early return alone would have imported that.

`execute_command` is now byte-identical to llama.cpp master, which should make the next ggml sync of this file a clean one.

**Line endings.** My first push normalized the whole file CRLF→LF and showed up as a 2496-line diff. `external/ggml/**` blobs in this repo really are CRLF, and `text=auto` in `.gitattributes` skips renormalizing a file whose *indexed* blob is already CRLF — which is how the original commit added 50 CRLF lines cleanly, and why every commit after the first LF one normalized. Both commits are rebuilt on the original with endings preserved. Cumulative diff is 48 insertions / 24 deletions now, and the content is byte-identical to what I tested.

**On testing CI locally** — you mentioned not knowing much about CI, so: I added `ci/run-local.sh`, which replays the workflow steps in the same base image via docker. `ci/run-local.sh linux-vulkan` runs the Vulkan job end to end in about 7 minutes against a real `glslc`, which is the check this fix never had. I'll send that as its own PR rather than bury it here — say the word if you'd rather not carry it.

Running it against this branch: 1932 shader symbols declared, every one defined, binary links with no undefined generated symbols.

One unrelated thing it turned up: the "Build and run unit tests" step ends in `No tests were found!!!` and `ctest` exits 0, so the job is green while running nothing. The suites are gated behind `ENGINE_BUILD_TESTS` / `ENGINE_BUILD_EXTENDED_TESTS` / `ENGINE_BUILD_MODEL_TESTS`, and `linux-build.yml` sets none of them. Separate issue; I can open one if it's useful.

0xShug0:
Thanks! Merged and let's see how it goes.

PR #542, opened 2026-09-13
vulkan: fix COL2IM_1D dispatch for large tensors

AdityaAWaghmare:
## Problem

The Vulkan implementation of `GGML_OP_COL2IM_1D` flattened the
`[T_out, OC]` output into a single X dispatch dimension.

For sufficiently large tensors this can exceed
`maxComputeWorkGroupCount[0]` on Vulkan devices with a lower X
workgroup-count limit, triggering the dispatch assertion.

This was first reproduced on Intel ADL GT2 with PocketTTS and
Qwen3-TTS for longer inputs, but the issue is not Intel-specific.
Any Vulkan device whose `maxComputeWorkGroupCount[0]` is smaller than
the required flattened dispatch can hit the same problem.

For example:

- `T_out = 52000`
- `OC = 384`
- total elements = `19,968,000`
- local size = `256`
- required X workgroups = `ceil(19,968,000 / 256) = 78,000`

This exceeds a device limit of `65,535`.

## Fix

Dispatch the two logical output dimensions separately:

- X = `T_out`
- Y = `OC`

The shader now maps `gl_GlobalInvocationID.x` to `t_out` and
`gl_GlobalInvocationID.y` to `oc`.

This changes the example dispatch from:

    (78000, 1, 1)

to:

    (204, 384, 1)

while preserving the existing output layout and COL2IM computation.

The existing push-constant layout is preserved.

This avoids relying on a large flattened X dispatch and makes the
operation portable across Vulkan devices with different
`maxComputeWorkGroupCount` limits.

## Testing

Tested with Vulkan on Intel ADL GT2:

- PocketTTS, 779 characters: 11.15 s, RTF 0.261
- Qwen3-TTS 0.6B, 779 characters: 97.60 s, RTF 1.946

Both previously reproduced the `maxComputeWorkGroupCount` assertion.

Also tested PocketTTS on NVIDIA RTX 3050 Vulkan:

- 779 characters: 3.23 s, RTF 0.076

Qwen3-TTS on the RTX 3050 could not be tested because the 4 GB VRAM
configuration ran out of device memory.


AdityaAWaghmare:
A couple of points for further discussion:

The 2D dispatch approach here follows the upstream llama.cpp implementation of GGML_OP_COL2IM_1D, introduced in ggml-org/llama.cpp#24425 ([commit ad39cca](https://github.com/ggml-org/llama.cpp/commit/ad39cca?utm_source=chatgpt.com)). The upstream implementation dispatches T_out and OC across X/Y respectively, with the same local_size_x = 256 approach used here.

Regarding the possibility of still exceeding maxComputeWorkGroupCount: that's a general property of dispatch dimensions and isn't specific to this change. The existing assertion already checks all three dispatch dimensions. The problem reproduced in this issue was specifically the flattened X dispatch: for the failing workload it required 78,000 X workgroups, exceeding the 65,535 limit on the Intel device. Splitting the logical dimensions gives (204, 384, 1) for that workload.

The concern about extra invocations for small T_out is valid and could potentially be optimized with a different/adaptive dispatch strategy. However, I kept this PR aligned with the upstream implementation and focused it on the correctness issue that caused the crash. Happy to discuss whether an adaptive approach would be preferable here.

0xShug0:
@AdityaAWaghmare Thank you for digging deep and finding the root cause. I need more time to test it because it changes dispatch cost for every shape and have performance implications. 
E.g., 
```
T_out=1, OC=384
old = 512 invocations
new = 98,304 invocations
```
It being upstream is a good sign, but we've tried bumping the GGML version before and couldn’t move forward because of performance regressions.

Could you try another apporach: keep shader untouched and cap the X dispatch size on the host side:

E.g.

```
case GGML_OP_COL2IM_1D:
         {
-            const uint32_t total = (uint32_t)(dst->ne[0] * dst->ne[1]);
-            elements = { total, 1, 1 };
+            const uint64_t total = uint64_t(dst->ne[0]) * uint64_t(dst->ne[1]);
+            const uint64_t max_x_elements =
+                uint64_t(ctx->device->properties.limits.maxComputeWorkGroupCount[0]) *
+                uint64_t(pipeline->wg_denoms[0]);
+
+            elements = {
+                uint32_t(std::min(total, max_x_elements)),
+                1,
+                1
+            };
```

The change is much smaller. The existing shader already has a grid-stride loop based on the actual dispatched workgroup count, so capping X should still cover the full output while preserving the existing dispatch for shapes that fit. 

Could you test this against the failing workload and compare correctness and performance around the dispatch boundary?

AdityaAWaghmare:
@0xShug0 I made the requested change: kept the shader untouched and capped the COL2IM_1D X dispatch size on the host side based on maxComputeWorkGroupCount[0].

I tested the patched build with a native CPU Release build.

Test results:

Input A — 96 chars
"Steven Paul Jobs (February 24, 1955 – October 5, 2011) was an American businessman and investor."

Official v0.7.4 portable Vulkan:
- Intel iGPU: 1767.95 ms, RTF 0.216661
- RTX 3050: 498.93 ms, RTF 0.061749

Patched build:
- Intel iGPU: 1730.22 ms, RTF 0.212036
- RTX 3050: 493.63 ms, RTF 0.061093

There is no apparent performance regression for this workload.

Input B — 274 chars
"Steven Paul Jobs (February 24, 1955 – October 5, 2011) was an American businessman and investor. A pioneer of the personal computer revolution of the 1970s and 1980s, Jobs co-founded Apple Inc. with his early business partner Steve Wozniak as Apple Computer Company in 1976."

Official v0.7.4 portable Vulkan:
- Intel iGPU: crashes with the maxComputeWorkGroupCount assertion
- RTX 3050: 2178.22 ms, RTF 0.093889

Patched build:
- Intel iGPU: 3957.17 ms, RTF 0.206103
- RTX 3050: 1983.26 ms, RTF 0.103727

The previously crashing Intel workload now completes successfully.

Input C — original issue reproducer (779 chars)

Patched build:
- Intel iGPU: 10877.84 ms, RTF 0.254631
- RTX 3050: 3115.25 ms, RTF 0.073473

Official v0.7.4 portable Vulkan:
- RTX 3050: 3132.40 ms, RTF 0.073877

The original long Intel reproducer now completes successfully. NVIDIA performance is essentially unchanged compared with the official build (~0.5% difference in RTF).

Overall:
- Short Intel workload remains at essentially the same performance.
- The previously crashing Intel workloads now complete successfully.
- NVIDIA performance is essentially unchanged.
- The shader remains untouched.
- The existing grid-stride loop covers the full output after the host-side dispatch is capped.

0xShug0:
@AdityaAWaghmare Thanks for testing! PR merged.

PR #546, opened 2026-09-14
webui: make the committed bundle reproducible

### LINKED ISSUE #545: webui/native/dist/index.html is committed but not reproducible, so any two web UI PRs conflict
## Summary

`webui/native/dist/index.html` is committed to the repository and is **not reproducible**: building the same source twice produces two different files. Any two pull requests that rebuild the web UI therefore conflict in that file, whether or not their source changes overlap.

## Reproducing

Two consecutive builds of an unmodified tree:

```
$ cd webui/native && npm run build
  build 1: id=__sveltekit_1ia7gsf  sha256=bb3f9c3c3be60ded…
  build 2: id=__sveltekit_1xqgjp8  sha256=6cd91a941e0e9d91…
```

The difference is a per-build identifier that SvelteKit embeds in three places:

```diff
-  __sveltekit_191iq4z = { … }
+  __sveltekit_1w7a9yp = { … }
-  this.__sveltekit_191iq4z = this.__sveltekit_191iq4z || {}; …
+  this.__sveltekit_1w7a9yp = this.__sveltekit_1w7a9yp || {}; …
-  __sveltekit_191iq4z.app.start(element);
+  __sveltekit_1w7a9yp.app.start(element);
```

It is derived from `kit.version.name`, which SvelteKit defaults to `Date.now().toString()`. `webui/native/svelte.config.js` does not set it, so every build gets a new one.

This is what made #539 conflict: the only conflicting file was `dist/index.html`, while `src/lib/text.ts` merged cleanly and no upstream commit had touched either file.

## Suggested fix

Set `kit.version.name` to something stable for a given source state. Verified — with

```js
kit: {
  …,
  version: { name: 'audiocpp' }
}
```

two consecutive builds produce an identical identifier and an identical SHA256.

**One trade-off worth deciding rather than inheriting:** SvelteKit uses `version.name` for its client-side "app has been updated" detection. A hard-coded constant disables that. A release version or a `git describe` output would keep the bundle reproducible for any given source tree *and* keep update detection meaningful across releases — probably the better value, though for a UI embedded in the server binary the whole binary is replaced anyway, so the detection may not be load-bearing here.

## Why not simply stop committing `dist/`

Worth stating explicitly, since it is the obvious first suggestion: `CMakeLists.txt:2226` embeds this file into the server binary at configure time, with a fallback page when it is absent:

```cmake
set(AUDIOCPP_UI_DIST ".../webui/native/dist/index.html")
if (EXISTS "${AUDIOCPP_UI_DIST}")
    file(READ "${AUDIOCPP_UI_DIST}" AUDIOCPP_UI_HEX HEX)
else()
    string(HEX "<!doctype html>…<p>Embedded WebUI assets are unavailable.</p>" AUDIOCPP_UI_HEX)
endif()
```

Committing the bundle is what lets the project build a working server **without a JavaScript toolchain installed**. Removing it would mean either requiring node for every build, or shipping a server whose UI is a placeholder for anyone who builds without running `npm run build` first — and the failure would be quiet, because the fallback succeeds.

Making the artifact deterministic fixes the conflicts without touching that property, which is why it seems the better lever.

## Lesser mitigations, if the above is unwanted

- `.gitattributes` with `webui/native/dist/index.html -merge` (or `merge=binary`) stops git attempting a line-merge of minified output. Conflicts still occur but resolve as "take either, then rebuild" instead of producing a corrupted hybrid.
- Regenerating `dist/` only on release rather than in every PR, with contributors changing `src/` alone.

Neither addresses the cause; both reduce the damage.

christopherthompson81:
Fixes #545.

`dist/index.html` is committed **and** embedded into the server binary at configure time (`CMakeLists.txt:2431`) — that is what lets the project build a working server without a JavaScript toolchain. But it is not reproducible: SvelteKit defaults `kit.version.name` to `Date.now()` and derives the embedded `__sveltekit_<id>` from it.

| | build 1 | build 2 |
|---|---|---|
| before | `__sveltekit_8hf8y8` / `416fcbab…` | `__sveltekit_12z8hlh` / `c1990991…` |
| after | `__sveltekit_1wn864` / `25cc1247…` | `__sveltekit_1wn864` / `25cc1247…` |

So any two branches that rebuild the web UI conflict in that file regardless of whether their source overlaps. That is what happened to #539, where `dist/index.html` was the only conflict while `src/lib/text.ts` merged cleanly and no commit on main had touched either file. It happened to this PR too — rebased now, and the only conflict against main was `dist/index.html`.

## The version number

Naming `kit.version.name` after the UI package makes the bundle a pure function of the source tree.

`webui/native/package.json` was still at the placeholder `0.1.0` it has carried since the web UI landed in #185, so this bumps it to **`0.8.1`**, the current release, and the identifier means something. Nothing derives it automatically: audio.cpp's version is tag-driven — `AUDIOCPP_VERSION` defaults to `"dev"` and is passed at configure time, and no version string is committed in the tree. So it needs bumping alongside a release, and the comment in `svelte.config.js` says so.

The comment also says plainly that this is an identifier, **not** a live update channel. An earlier draft of this PR justified the package version over a constant by claiming it kept SvelteKit's "app has been updated" check working. That was wrong three ways: the app never imports the `updated` store, `version.pollInterval` is left at its default of `0`, and the single-file bundle ships no `_app/version.json` for a client to poll. A stale version here costs a stale label and nothing else.

## Verification

**Reproducible beyond repeat runs.** A clean `npm ci` build on node **20**, **22** and **24** each produces a `dist/index.html` byte-identical to the one committed here, so the artifact is stable across contributors rather than only across builds on one machine.

**A rebuild leaves the tree clean.** `npm run build` on the committed tree produces no diff — previously it produced one on every run.

## A note on what the rebuild changed

main's committed bundle does not currently match a clean build of its own source: about **4.5 KB** apart, and 1668 differing lines once the build id is normalised. The differences are **minifier identifier renaming** (`M8` → `z8` → `E8`, cascading), not missing UI — the feature strings I sampled from the recent web UI work (ZipVoice, Confucius4-R2T2) are present in both, at identical counts.

So the bundle committed here is what this tree actually produces, which is not quite what was committed before. I could not attribute that drift to a specific commit; every web UI commit in the range did rebuild `dist/`.

That is the argument for a guard: a CI step asserting `npm run build` leaves `git status` clean would catch this at PR time instead of as a mystery conflict later. Deliberately left out of this PR as separate work — happy to open it if you want it.

## Alternatives considered

**Stop committing `dist/`.** This was the first idea and it has a real cost: the CMake step above falls back to a placeholder page when the file is absent, and the fallback *succeeds*. Removing the bundle would mean either requiring node for every build, or quietly shipping a server whose UI is a placeholder for anyone who did not run `npm run build` first. Making the artifact deterministic fixes the conflicts without giving that up.

**`.gitattributes` with `-merge`** on the file would stop git attempting a line-merge of minified output. Conflicts would still occur, but resolve as "take either, then rebuild" rather than producing a corrupted hybrid. Worth doing as well, perhaps, but it treats the symptom.

https://claude.ai/code/session_01TMmMgd5xNGnjsQgybuUiK3


christopherthompson81:
Moving this to draft so it isn't sitting in your review queue while you think over the trade-offs in #545 — no pressure either way, just ping me and I'll mark it ready again.

0xShug0:
@mirek190 Any thoughts on this PR?

mirek190:
> @mirek190 Any thoughts on this PR?

I look on that

christopherthompson81:
Rebased onto main and rebuilt — this was stale enough to be dangerous, so it's worth saying what changed.

**The bundle was 41 commits behind.** The version committed here previously was built at #539 and was 680 KB against main's 1.24 MB, with **zero** references to the YuE2 UI that landed since. Merging it by taking either side of the conflict would have silently shipped a server with no YuE2 support — silently, because the CMake embed falls back to a placeholder only when the file is *absent*, never when it is stale. Rebuilt from current main's source, so the YuE2 UI is present and the bundle matches a fresh build of main exactly.

**The version number.** `webui/native/package.json` was still at the `0.1.0` placeholder it has carried since #185 — never bumped, so as a "version" it was doing nothing. Bumped to `0.8.1` to match the current release. There is no in-tree source to derive it from (`AUDIOCPP_VERSION` defaults to `"dev"` and is supplied at configure time), so it needs bumping alongside a release; the comment in `svelte.config.js` says so.

**I removed a claim from the original description that was wrong.** It said using the package version kept SvelteKit's "app has been updated" check working, and that a constant would have disabled it silently. That check isn't wired up at all: nothing imports the `updated` store, `version.pollInterval` is at its default of `0`, and the single-file bundle ships no `_app/version.json` to poll. The version is an identifier, nothing more, and the comment now says that instead of implying otherwise.

**One content change rides along.** Commit `048c9a0c` changed a utility description and added a load option to the catalog without regenerating `dist/index.html`, so main's committed bundle is currently 17 lines behind its own source. Rebuilding closes that. Normalising the version constant, this bundle differs from a fresh build of main in **zero** lines.

**Verification:** byte-identical output under node 20.20.2, 22.12.0 and 24.21.0; `npm run build` on the committed tree now leaves `git status` clean; merges into main with no conflicts.

That staleness is worth a guard — a CI step asserting a rebuild leaves `git status` clean would have caught both this PR's stale bundle and main's current one at PR time. Say the word and I'll open it separately.

christopherthompson81:
Correction to my previous comment — I rebased onto my fork's `main`, which was 33 commits behind yours, so the numbers in it were wrong. Redone against `0xShug0/audio.cpp@6c70f32d`; the PR now reports mergeable. Sorry for the noise.

What stands from that comment: the branch was badly stale (the old bundle predated the YuE2 web UI entirely), and the "app has been updated" justification in the original description was wrong on three counts, so it's gone. The version is bumped from the `0.1.0` placeholder — untouched since #185 — to `0.8.1`.

What changed with the correct base: the missing-YuE2 framing no longer applies, since that work is well inside the new base. The real finding is different and worth your attention.

**main's committed bundle doesn't match a clean build of its own source.** About 4.5 KB apart, 1668 differing lines once the build id is normalised. I checked whether it was missing UI and it isn't — the feature strings I sampled (ZipVoice, Confucius4-R2T2) appear in both at identical counts, and every web UI commit in the range did rebuild `dist/`. The differences are minifier identifier renaming, cascading through the file.

I could not attribute it to a specific commit. The most likely explanation is that the committed bundle was produced in a slightly different environment than a clean `npm ci` gives today. Either way it means the artifact in the tree is not currently reproducible from the tree, which is the same class of problem this PR is fixing, one layer down.

So the bundle here is what the tree actually produces now, verified with a clean `npm ci` on node 20, 22 and 24 — all three byte-identical to the committed file.

That drift is a good argument for the guard I mentioned: a CI step asserting `npm run build` leaves `git status` clean would have surfaced it at PR time. Say the word and I'll open it separately.

0xShug0:
Thanks for the update. Merged as a temporary fix.

PR #547, opened 2026-09-14
fix(models): align default context sizes in model implementations on 32-bit targets

weidu-tpvision:
Follow up on #513 by applying pointer-width preprocessor guards to cap default weight and graph context sizes to safe 1 GB thresholds on 32-bit targets while preserving original defaults on 64-bit architectures.


0xShug0:
@weidu-tpvision Merged. Thanks!

PR #555, opened 2026-09-15
refactor(framework): promote greedy Qwen decoder runtime and Qwen BPE bundle helpers

gqf2008:
## Summary

Moves the greedy Qwen decoder runtime and the Qwen BPE bundle helpers out of the `audio8_asr` community model into the framework, so other Qwen-based families can reuse them instead of carrying their own copies.

## Changes

New framework pieces (moved from `src/community_models/audio8_asr/thinker.cpp`):

- `include/engine/framework/runtime/greedy_qwen_decoder.h` / `src/framework/runtime/greedy_qwen_decoder.cpp`
- `include/engine/framework/tokenizers/qwen_bpe_bundle.h` / `src/framework/tokenizers/qwen_bpe_bundle.cpp`

`audio8_asr` keeps using the same logic through the promoted runtime; the moved types are renamed to match their new home (`ThinkerWeights` -> `DecoderWeights`, `TextLayerWeights` -> `DecoderLayerWeights`, ...). `CMakeLists.txt` registers the new sources.

## Verification

- Rebased onto current `main` (cherry-pick, no conflicts).
- Both targets build: `audiocpp_cli`, `audiocpp_server`.
- ASR output is identical to `main` on the same input — `assets/resources/b.wav` transcribes to
  `Some call me nature. Others call me Mother Nature. I've been here for over four point five billion years, twenty-two thousand five hundred times longer than you.`
  byte for byte on both trees, so the move does not change behaviour.


0xShug0:
@gqf2008 PR merged. Thanks!

PR #557, opened 2026-09-15
yue2: eager NAR attention on Intel Vulkan

engival:
**Why:** ggml-vulkan's flash-attention kernels are slower than the plain matmul+softmax path on Intel GPUs, so the Yue2 NAR was spending most of the song on a losing kernel.

**How:** Adds a `yue2.attention=auto|flash|eager` session option and an additive `vulkan_device_is_intel()` probe; `auto` picks eager for the NAR only when the Vulkan device is Intel. The shared `resolve_flash_attention()` is untouched, so no other model family or backend changes behaviour, and the AR stays on flash (eager measured 4x slower there).

**Result:** On an Arc Pro B70 (cot=full, 32 ODE steps) the NAR drops from 275 s to 126 s and song wall time from 354 s to 205 s. Explicit `flash` is byte-identical to the previous output; eager vs flash on the same seed is 36 dB.


0xShug0:
@engival Appreciate the optimization for Intel GPUs. PR merged!

PR #560, opened 2026-09-15
qwen decode runtime: zero the decode graph buffer after allocation

engival:
**Crash:** Yue2 with `cot=full` on AMD Vulkan (RX 7900 XTX, RADV) dies on the first decode step, every run:
`Yue2 AR sampler has no finite logits`. `cot=off` works.

**Cause:** the decode KV cache buffer is allocated and never zeroed. The ABC-stage prefill writes only the prompt slots, but decode attention reads every slot (masked to -inf). NaN in the unused slots still leaks through, and RADV doesn't zero fresh device memory.

**Fix:** clear the buffer once after allocation, same as llama.cpp does for its KV cache.


0xShug0:
@engival I think this change is too broad (though it doesn’t affect the models I care about). If the issue is specific to YuE2, a narrower fix should go in `prefill_tokens_into_decode_cache`() to ensure the KV cache is properly zero-initialized. YuE2 is currently the only model that calls this function.

engival:
Okay, I moved it. Fable tested it:

Regression runs on the 7900 XTX with the rebuilt binary, all exit 0 and no "no finite logits":

| Run (RX 7900 XTX, RADV, Vulkan) | Result |
|---|---|
| cot=full, default 32 NAR steps | completes, 118 s, no "no finite logits" |
| cot=full, repeated | byte-identical wav to the first run |
| cot=full, 4 NAR steps | byte-identical to the earlier whole-buffer-clear output |
| cot=off, full song | completes, 110 s |

0xShug0:
@engival Thanks! PR merged.

PR #561, opened 2026-09-15
Export generated Yue2 ABC plan as a score artifact

chbgdn:
When cot=melody/cot=full runs without an external abc/abc_file, Yue2 generates its own ABC transcription (chain-of-thought plan), but the plan was discarded after conditioning the semantic stage.

Changes
- Yue2TextTokenizer::decode() — wraps the existing decode_bpe
- Yue2PipelineRuntime::run() returns Yue2RunResult { audio, plan_abc_text, plan_abc_truncated }
- Yue2Session::run() attaches a score artifact (text/vnd.abc) for model-generated plans only; user-provided ABC is not re-emitted
- CLI file sink honors the artifact meta extension and writes non-JSON payloads raw, so --out-dir produces score.abc instead of a JSON envelope. Note: this renames the pre-existing SheetSage2 CLI/batch score output from score.json to score.abc (the payload was always raw ABC text)

No WebUI changes needed — the native WebUI already renders result artifacts as downloads (Save ABC appears in the result panel), and the Yue2 panel ABC textarea accepts the exported text for re-import/iteration.

Verification: full release build of the commit (server + CLI); plan token ids are provably pure-text (AR sampling window [0, kEodToken), stop token stripped), so the decoded text is a faithful byte-exact inverse.


0xShug0:
@chbgdn Thanks! PR merged. 

One small issue: const auto run_result makes std::move(run_result.audio) copy the audio instead of moving it. I will fix  that in a small follow-up.

PR #564, opened 2026-09-15
kokoro_tts: skip phonemes the vocab has no id for, as KModel does

### LINKED ISSUE #556: kokoro_tts: throws on phonemes its own G2P produces — cannot synthesize "button", "written", "kitten"
`kokoro_tts` fails on ordinary English words. eSpeak-ng glottalises /t/ before a syllabic nasal, so "button" phonemizes to `bˈæʔn̩`, and `encode_input_ids_and_count` then rejects the U+0329 (COMBINING VERTICAL LINE BELOW) that the engine's own G2P just produced:

```
Kokoro vocab is missing phoneme symbol: ̩ (runtime error)
```

## Reproduction

Model `kokoro_82m_q8_0` (`Kokoro-82M-GGUF/kokoro-82m-q8_0.gguf`), voice `af_heart`, language `en-us`, task `tts`/`offline`:

| text | result |
|---|---|
| `button` | **throws** — missing phoneme symbol U+0329 |
| `kitten` | **throws** |
| `written` | **throws** |
| `Rustenburg` | **throws** |
| `hidden` | ok |
| `Wittenberg` | ok |

I hit this through the C ABI (`audiocpp_session_run`), but the throw is in `frontend.cpp` before any entry-point divergence, so it should reproduce identically from the CLI and the server.

Confirm the phonemization with eSpeak-ng directly:

```console
$ espeak-ng -q --ipa -v en-us "button kitten Rustenburg"
bˈʌʔn̩ kˈɪʔn̩ ɹˈʌsʔn̩bˌɜːɡ
```

## Cause

`src/models/kokoro_tts/frontend.cpp:149`:

```cpp
const auto it = assets.vocab.find(phonemes.substr(i, width));
if (it == assets.vocab.end()) {
    throw std::runtime_error("Kokoro vocab is missing phoneme symbol: " + phonemes.substr(i, width));
}
```

This diverges from the reference implementation. hexgrad/Kokoro's `KModel` tokenizes with

```python
input_ids = list(filter(None, map(vocab.get, phonemes)))
```

— it **drops** phonemes with no token id rather than failing. Kokoro's vocab has 114 entries and simply has no id for the syllabic diacritic; dropping it gives `bˈʌʔn` for "button", which is a correct reading. Throwing gives no audio at all.

## Impact

Across the first 40,000 entries of a general English pronunciation dictionary, 100 words (0.25%) produce an out-of-vocab symbol — 99 of them U+0329, one U+026C. That percentage understates the effect, because the affected words are the common `-tten`/`-tton` family (beaten, bitten, written, forgotten, gotten, kitten, button, cotton), and the throw aborts the whole request. Any document of reasonable length is likely to contain one.

## Suggested fix

Skip unknown symbols instead of throwing, matching `KModel`. A debug trace on the skip would keep the information without failing the request. The existing UTF-8 validation throws above it should stay — malformed UTF-8 is a real error; an unknown-but-valid phoneme is not.

Happy to send a PR if the approach looks right.

christopherthompson81:
Fixes #556.

`encode_input_ids_and_count` throws on any symbol missing from Kokoro's 114-entry vocabulary — **including symbols the model's own G2P has just produced** — which aborts the whole request.

The reference implementation does not. hexgrad/Kokoro's `KModel` tokenizes with `input_ids = list(filter(None, map(vocab.get, phonemes)))`: it *drops* phonemes with no token id. Dropping gives a correct reading; throwing gives no audio at all.

Malformed UTF-8 in the checks above still throws, because that is a real error. An unknown but well-formed phoneme is not. Each skip is recorded as a debug trace (`kokoro.skipped_phoneme`), so the information is kept without failing the request.

## What it costs today

40 real corpus sentences per language through the release `kokoro_82m_q8_0` on the current code:

| language | result | symbol |
|---|---|---|
| en-us / en-gb | 40 ok / 0 threw | — |
| **es** | 35 ok / **5 threw** | `«` U+00AB |
| **fr-fr** | 34 ok / **6 threw** | `«` U+00AB |
| **it** | 39 ok / **1 threw** | `̪` U+032A |
| hi / pt-br | 40 ok / 0 threw | — |

So roughly 12% of Spanish and 15% of French prose fails outright — the vocabulary carries `(` and `)` but not the guillemets. English is affected more rarely but on ordinary words: eSpeak glottalises /t/ before a syllabic nasal, so `button`, `kitten`, `written` and `forgotten` produce a U+0329 the vocabulary lacks. 100 of the first 40,000 entries of a general English pronunciation dictionary, and one anywhere aborts the request.

## Impact on outputs, performance and memory

**Outputs:** unchanged for any text that did not previously throw — verified byte-for-byte below. Text that previously threw now produces audio.

**Performance and memory:** unaffected. The change replaces a throw with a `continue` in a loop that was already running.

## Runtime validation

Built and run against `kokoro_82m_q8_0`, not inspected.

Every previously-failing case renders, with durations in line with words that always worked (`button` 1.00 s beside `hidden` 1.07 s), and 114/114 distinct English golden sentences render.

Regression checks for the affected route — `ctest`, with a models root configured:

| gate | result |
|---|---|
| `audiocpp_c_api_path` / `_exports` | Passed |
| `audiocpp_c_api_model` | Passed (146 s) |
| `audiocpp_c_api_parity` | **PARITY OK across 5 families** |

```
kokoro_tts audio MATCH cli=79800@24000Hz peak=0.3383 api=79800@24000Hz peak=0.3383
```

79800 frames at 24 kHz is 3.325 s — **byte-identical to what the unpatched engine produces for that sentence**, and identical between the CLI and the C API. That is the evidence for "inert on text that never had an out-of-vocabulary symbol".

## A note on measuring this

The per-language numbers above came from restoring the throw in a patched build and running real corpus text. They did **not** come from diffing `espeak-ng -q --ipa` output against the vocabulary, which is the obvious shortcut and is wrong in both directions: it predicts 507 hyphen failures in French that never occur — `espeak_text()` ends with `replace(out, "-", "")` — and it misses the guillemets that actually fail. Anyone reviewing this by reaching for the CLI will get a different answer than the engine gives.


0xShug0:
@christopherthompson81 Can't wait for the Nix Vulkan CI to finish. PR merged!

christopherthompson81:
Filed #565, which fixes the same `button` symptom at its cause, so the two should be read together.

#565 restores misaki's English eSpeak arm — this port has `EspeakG2P` and uses it for English too, where misaki keeps a separate `EspeakFallback` with a much richer table. Among other things that table rewrites the syllabic consonant as schwa + consonant, so `button` becomes `bˈʌtn` instead of the `bˈʌʔn̩` that throws here.

That makes the framing of this PR narrower than I wrote it, and I would rather say so than let the example stand:

- **`button` is not really a vocabulary-tolerance problem.** It is a missing normalisation, and #565 fixes it properly. Dropping the mark yields `bˈʌʔn` — in vocabulary, but a glottal stop with no schwa, which is not what Kokoro was trained on.
- **The guillemets in the Spanish and French numbers are the same story** — misaki converts `« »` to curly quotes, which are in the vocabulary. #565 does that too.

What this PR is still for, on its own merits: an unknown but well-formed phoneme should not abort a whole request, whatever produced it. That remains true for anything neither table anticipates, and it is what the reference implementation does. I would keep it as the backstop — but **#565 is the one that improves output quality**, and if only one lands it should be that one.

Happy to close this in favour of #565 if you would rather have a single change.

PR #573, opened 2026-09-16
HIP: speed up RoFormer FP32 attention with contiguous operands

lkarlslund:
RoFormer explicit attention on HIP can fall back to a separate hipBLASLt GEMM launch for each batch element when the left operand is noncontiguous. Materialize that operand as contiguous on HIP so the existing FP32 path can use strided batched GEMM.

This is a **HIP-only optimization for RoFormer separation**. CUDA, Vulkan and other backends are unchanged. It does not accelerate YuE2 music generation, enable HIP flash attention, or downcast attention operands to FP16.

### Measured speedup

Tested on Radeon 8060S (`gfx1151`), ROCm 10.1 nightly, BS-RoFormer ep368 Q8_0, against upstream `048c9a0`. Same model and inputs, sequential fresh CLI processes; elapsed times include startup.

| Input | Upstream | HIP FP32 layout fix |
|---|---:|---:|
| 3-second speech/music mixture | 140.536 s | 5.925 s (**23.7× faster**) |
| 1-second speech | 4.421 s | 3.971 s |
| 12-second chirp probe | timed out at 180.030 s | 14.142 s |

These are bounded workstation observations, not a general speedup guarantee. The timeout is a lower bound, not a completed baseline measurement.

### Output validation

Compared both emitted PCM16 stems sample by sample after checking sample rate, channel count and frame count, using identical model/input files. The benchmark helper records executable/input hashes, command, relevant environment variables, elapsed time, maximum absolute sample difference and normalized RMS difference.

**Outputs are not bitwise identical.** For the 3-second mixture, the candidate differed from the first baseline by at most 9 PCM16 steps (vocal RMS difference 3.785e-5). Repeating the unchanged baseline binary with the same input itself differed by up to 12 steps (vocal RMS difference 3.962e-5). A repeated candidate run took 5.824 s and differed from the first candidate by up to 9 steps. The one-second comparison differed by at most 3 steps. No parity claim is made for the timed-out twelve-second baseline.

The observed differences are comparable to baseline repeat variability; this does not prove identical output, greater accuracy, or inaudibility. The exact source of baseline variability has not been isolated. FP32 types, attention precision, scaling and softmax remain unchanged; GEMM scheduling can affect rounding.

### Validation and scope

- Built HIP gfx1151 CLI and server; verified current executable hashes match the measured binaries.
- Added a reproducible, bounded benchmark/comparison helper and report with commands and limitations.
- Other AMD architectures, Mel-Band RoFormer, longer real recordings and listening quality need additional coverage.

Implemented using **OpenAI Codex**.


0xShug0:
@lkarlslund Thanks for the optimization! PR merged. Looks like this optimization can potentially help other models on HiP.

PR #576, opened 2026-09-16
cmake: say what the CUDA build is about to cost, and add an opt-in --ccache

christopherthompson81:
A local CUDA build compiles every `.cu` file once per entry in the default arch list — nine of them — and caches nothing. **No default changes in this PR:** it makes the cost visible and adds an opt-in flag, nothing more. (Revised after review — see [the note at the end](#revised-after-review).)

One machine, 16 cores, RTX 3090 (sm_86), CUDA 12.8, NVMe. One variable at a time, cold configure into an empty directory, `-j 16`.

| configuration | cold build | warm rebuild |
|---|---:|---:|
| as it is today | **1361 s** | 1361 s |
| + ccache, **arch list untouched** | 1412 s (+3.7%) | **30 s** |
| single arch (`native`) | 401 s | 401 s |
| single arch + ccache | 409 s | 29 s |

Two independent effects:

**CUDA is ~70% of the wall time.** A process census at 16% shows `cicc=16, cc1plus=0` — the 779 C++ objects cannot start while CUDA saturates every core, so the C++ tail is serialised behind it rather than overlapped. One architecture is 3.4x and takes `libaudiocpp.so` from 201 MB to 69 MB.

**Nothing is cached.** `CMakeLists.txt` forces ggml's ccache probe off (`"Do not probe for ccache in this local setup"`, with `FORCE`, so `-DGGML_CCACHE=ON` cannot override it). Wiring a launcher up instead gives a 97.9% hit rate on a 0.2 GB cache, and a rebuild of **1361 s → 30 s**.

The warm column is worth a second look: **30 s, 29 s across both arch configurations.** On a cache hit nothing compiles, so the arch list is decisive for a cold build and nearly irrelevant to a rebuild.

## 1. `CMakeLists.txt` — interpret what configure already prints

```
-- Using CMAKE_CUDA_ARCHITECTURES=50-virtual;...;120a-real CMAKE_CUDA_ARCHITECTURES_NATIVE=86-real
```

The count about to be compiled and the one this host can run, adjacent on the same line, with no indication that the second is an option. Now:

```
--   portable default: 9 architectures, so every .cu file is compiled 9 times
--   this host is 86-real; for a local build roughly 3x faster:
--       -DCMAKE_CUDA_ARCHITECTURES=native    (the result then runs on this GPU only)
--       scripts/build_linux.sh: --cuda-arch native
-- ccache is installed (/usr/bin/ccache) and NOT in use. Rebuilds are much faster with:
--     -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_CUDA_COMPILER_LAUNCHER=ccache
--     scripts/build_linux.sh: --ccache
```

`--cuda-arch native` is already documented in `README.md` and `docs/build/linux.md`. It was still missed across every build on this machine for days — configure is where the choice is actually being made. Each hint names the script flag next to the cmake one, because `build_linux.sh` rejects unknown arguments and so takes no pass-through `-D`.

Each property verified, not assumed:

- **Silent once correct** — `native` plus the launchers prints neither hint. A hint that keeps firing after you comply gets filtered out along with the real ones.
- **Never fires on a GPU-less host** — gated on the detected arch being numeric, so a build server is never advised toward `native`, which is exactly the case #280 protects.
- **Distinguishes asking for the portable list from getting it by not asking**, via a flag set only in the defaulting branch.
- The ccache hint sits **outside** the CUDA block, so CPU-only, Vulkan and HIP builds get it too.

## 2. `scripts/build_linux.sh` — an opt-in `--ccache`

One flag. Unset, the script behaves exactly as it does on `main`; the configure hint above is what tells you the flag exists.

```bash
scripts/build_linux.sh --backend cuda --cuda-arch native --ccache --target audiocpp_cli
```

A `CMAKE_<LANG>_COMPILER_LAUNCHER` already set in the environment is kept, not replaced, and the script says so:

```
CMAKE_CXX_COMPILER_LAUNCHER=sccache is set; leaving the CXX launcher alone.
```

`--ccache` with no ccache installed exits 1 rather than silently doing nothing.

## Deliberately not proposed

Changing the default arch list. #280 (`62735ea`) fixed a real defect and its reasoning still holds: a GPU being present does not mean it is the target. Build hosts have GPUs, and a native default there silently produces a binary for the wrong hardware — the same class of failure #280 removed.

Reading that diff also showed the pre-#280 behaviour was worse than its commit message implies. The only `CUDA_ARCHITECTURES` setting in the tree named one target:

```cmake
set_target_properties(engine_runtime PROPERTIES CUDA_ARCHITECTURES native)
```

so `engine_runtime` built for the host GPU while ggml-cuda built for nvcc's fallback — one binary, two notions of the target. Moving the decision before `enable_language(CUDA)` and applying it globally was the right fix, and this PR does not touch it.

## Measured and dropped

- **Linker** — relinking the 201 MB `.so` takes **2 seconds**, 0.15% of the build. `mold`/`lld` would save nothing worth having.
- **Ninja** — Unix Makefiles' no-op rebuild on 953 objects is **1 second**. No incremental win to chase.
- **`-j` level** — 12/16/24/32 measured: 16 (`nproc`) is best at 399 s, 24 and 32 are slower (411 s, 422 s). The script's existing `nproc` default is already right.
- **`AUDIOCPP_MODEL_SET=custom`** — works and composes (285 s for two families), but the 141 CUDA objects are a floor it cannot touch, and the resulting binary cannot load the families it skipped.

## Also noticed, not changed here

`.github/workflows/release.yml:501` says the default is `"native" (the build host GPU)`. That was true before #280 and is not now. The conclusion — pin the architecture for releases — is still correct.

<a name="revised-after-review"></a>
## Revised after review

The first version also auto-detected the CUDA arch in `build_linux.sh` and turned ccache on whenever it was installed. Both are gone: they changed the default behaviour, which was never the point of the PR.

The auto-detection had two defects beyond the objection itself, both confirmed:

- Its detection pipeline was a bare assignment under `set -euo pipefail`. `grep` matching nothing returns 1, `pipefail` propagates it, and `set -e` exits — so a host with `nvidia-smi` present but no numeric compute capability (`N/A`, a stub driver, a failed query) exited 1 *before* reaching the portable fallback it was supposed to fall back to.
- It gated only on `--cuda-arch`, so `CUDAARCHS=75;86` in the environment was silently replaced by `-UCMAKE_CUDA_ARCHITECTURES -DCMAKE_CUDA_ARCHITECTURES=86`.

The ccache wiring claimed in a comment that appending the `-D` let a caller's launcher win. That was wrong twice over: `-D` beats both the environment and an existing cache entry (checked directly), and the script rejects unknown arguments, so no caller-supplied `-D` reaches that cmake invocation at all.

## Note on the numbers

Measured at `ff1bcc4`, before rebasing onto current `main`. Commits since then add a model family, so today's absolute cold-build figure is a little higher; the ratios are structural and unaffected.

https://claude.ai/code/session_014X4jSwJu1ScZVgUspcA9HC


0xShug0:
@christopherthompson81 

Please make ccache usage opt-in and preserve the existing default behavior. When ccache is installed, the script unconditionally sets the C, C++, and CUDA compiler launchers to ccache, overriding existing cache or environment settings. Explicit launcher settings should always take precedence.

I’d suggest removing auto-detection. The GPU on the build machine isn’t necessarily the GPU that will run the binary, and --cuda-arch native already lets users make that choice explicitly.

Besides, some issues about auto-detection:

It is unguarded under `set -euo pipefail`. If `nvidia-smi `fails, or grep finds no valid numeric compute capability, the assignment returns nonzero and the script exits before reaching the portable fallback. 

It  checks only whether `--cuda-arch` was supplied. It ignores `CUDAARCHS` and any existing `CMAKE_CUDA_ARCHITECTURES` cache value. For example, with `CUDAARCHS='75;86'` and a detected capability of 8.6, it passes -`UCMAKE_CUDA_ARCHITECTURES` -`DCMAKE_CUDA_ARCHITECTURES`=86, replacing the explicitly requested architecture list.

christopherthompson81:
Both changed as you asked, in f4a6097. The defaults are back to what `main` does; the PR is now just the configure-time hints plus one opt-in flag.

**Auto-detection is gone.** `--cuda-arch` unset means the portable list again. You were right about both defects, and I confirmed each:

- Under `set -euo pipefail` it was worse than not reaching the fallback — the script *exits*. `grep -xE` matching nothing returns 1, `pipefail` propagates it, and it was a bare assignment rather than an `if` condition, so `set -e` takes it. Isolated repro of that exact pipeline with a stand-in printing `N/A`: exit 1, and the fallback's `echo` never runs.
- `CUDAARCHS` likewise: the gate was `-z "$CUDA_ARCH"` only, so `CUDAARCHS='75;86'` was replaced by `-UCMAKE_CUDA_ARCHITECTURES -DCMAKE_CUDA_ARCHITECTURES=86`.

**ccache is opt-in.** `--ccache`, off by default. My comment claiming the appended `-D` let an explicit launcher win was wrong twice over — `-D` beats both the environment and an existing cache entry:

```
env CMAKE_C_COMPILER_LAUNCHER=sccache, no -D    -> sccache /usr/bin/cc ...
env sccache + -DCMAKE_C_COMPILER_LAUNCHER=ccache -> ccache
cached sccache, reconfigure with -D ccache       -> ccache
```

and the script rejects unknown arguments, so no caller-supplied `-D` ever reached that cmake line anyway. With the flag, a launcher already in the environment is kept and reported (`CMAKE_CXX_COMPILER_LAUNCHER=sccache is set; leaving the CXX launcher alone.`), and `--ccache` without ccache installed exits 1 instead of quietly doing nothing.

What is left is the part that was actually measured: configure says how many architectures it is about to compile, and that ccache is installed and idle. Both lines now name the script flag next to the cmake one, since the script takes no pass-through `-D`, and both still go silent once you have taken them up — verified in both directions on this host.

PR description updated to match.

0xShug0:
Thanks, PR merged!

PR #579, opened 2026-09-16
gguf: carry a GGUF input's embedded sidecars, and check the finished package against the one it was made from

### LINKED ISSUE #578: audiocpp_gguf: a gguf-to-gguf re-emit drops embedded sidecars while reporting that it reused them
Feeding an existing package back through `audiocpp_gguf` to change one piece of metadata produces a package with **no voices**, exits 0, and prints that the sidecars were reused.

## Repro

```bash
audiocpp_gguf --input  kokoro-82m-q8_0.gguf \
              --output kokoro-respec.gguf \
              --type orig --family kokoro_tts \
              --model-spec model_specs/kokoro_tts.json --overwrite
```

Output:

```
note: reusing the sidecars embedded in .../kokoro-82m-q8_0.gguf; pass --root to override
gguf=.../kokoro-respec.gguf
weight_type=orig
tensor_sources=1
embedded_sidecars=true
embedded_model_spec=true
model_spec_family=kokoro_tts
```

Exit code 0. Both sidecar claims are false:

| | voice sidecar entries | size |
|---|---:|---:|
| `kokoro-82m-q8_0.gguf` | **54** | 189,611,360 |
| `kokoro-respec.gguf` | **0** | 161,345,952 |

`--inspect` reports `tensors=548` and `embedded_sidecars=true` for both, so it does not show the difference either.

The result is unusable:

```
load FAILED: failed to open binary file: /tmp/audiocpp-gguf/911a3828bf36b9f1/voices/zm_yunxia.bin
```

## The spec half works

This is specifically the sidecars. The re-emitted package's embedded spec is correctly replaced — the new option I was testing is present in the output and absent from the input — so `--model-spec` does its job and only the sidecar payloads are dropped.

## Why it is worth more than the exit code suggests

For this package there is no way back. The installed directory holds only the `.gguf` and a manifest:

```
kokoro-82m-q8_0.gguf
.audiocpp-package-kokoro_82m_q8_0.json
```

The 54 voices exist **only** inside the GGUF, so `--root` has nothing to point at and the re-emit is lossy with no local recovery — you would have to re-download or re-convert from the original weights. Someone updating a package in place could reasonably delete the input afterwards.

The misleading note is the sharp edge rather than the omission itself. A hard error ("cannot carry embedded sidecars from a GGUF input; pass --root") would be fine, and would have taken about a second to act on. Printing `embedded_sidecars=true` invites the opposite conclusion.

## Suggested shape

Either carry the embedded sidecars through a GGUF-input re-emit, or refuse the combination with a message naming what is missing. Both are fine; silently producing a voiceless package is the part that is not.

To be clear about scope: this is not a request for an in-place spec-patching mode. The embedded spec being authoritative is the right design — the package is self-contained and the package is the unit you rebuild. The ask here is only that a re-emit either preserves what it claims to preserve or says it cannot.

Found while checking whether an existing Kokoro package could pick up a newly declared request option without a full rebuild.

christopherthompson81:
Fixes #578.

Re-encoding a finished package drops every sidecar the exclusion heuristic happens to match, exits 0, and says it reused them:

```
$ audiocpp_gguf --input kokoro-82m-q8_0.gguf --output out.gguf \
                --type orig --family kokoro_tts --model-spec model_specs/kokoro_tts.json

note: reusing the sidecars embedded in .../kokoro-82m-q8_0.gguf; pass --root to override
embedded_sidecars=true
```

| | voice entries | size |
|---|---:|---:|
| in | **54** | 189,611,360 |
| out | **0** | 161,345,952 |

```
load FAILED: failed to open binary file: /tmp/audiocpp-gguf/.../voices/zm_yunxia.bin
```

There is no way back for that package: the installed directory holds only the `.gguf` and a manifest, so the voices exist *only* inside the file being re-encoded.

## Cause

`resolve_sidecar_root` materialises the input's embedded sidecars and points the conversion at them — that part is right. The writer then **rediscovers** them by walking that directory, and the walk applies the filter meant for a source tree full of weights:

```cpp
extension == ".safetensors" || ".gguf" || ".bin" || ".pt" || ".pth" || size > 64 MiB
```

Kokoro's voices are `voices/*.bin`. They were read out of the package and thrown away by a rule whose job is to stop `pytorch_model.bin` being mistaken for a config.

That heuristic is right for a directory someone points `--root` at, and wrong for one this tool just wrote out of a finished package: those files are not candidates to be filtered, they **are** the sidecar set.

It never bit during packaging because `tools/prepare_kokoro_gguf.py` writes the GGUF itself with Python's `GGUFWriter` and embeds the voices directly — it never goes through this tool. The path only runs on a re-encode.

## Fix

Carry them across by name, read from the GGUF rather than from a walk. Explicit sidecars already bypass both copies of the filter (`app/gguf/main.cpp` for planning, `tensor_source.cpp` for the write) and are already de-duplicated against the walk, so nothing is embedded twice and the safetensors path is untouched.

`gguf_embedded_sidecar_names` is the small accessor that was missing — `gguf_has_embedded_sidecars` already reads exactly that array and discards it. It reads only `audiocpp.embedded_files.names`, not the contents, and keeps the unsafe-name rejection the full reader applies.

## The tool now checks its own output

This shipped because nothing compared the finished file against what was asked for. `embedded_sidecars` was read from the output but was only a boolean, so 5 of 59 still printed `true`; `embedded_model_spec` was read from the in-memory intent and never from the file at all.

A re-encode now reads the package back and compares it to the one it came from:

```
error: re-encoding kokoro-82m-q8_0.gguf kept 5 of its 59 sidecars; missing:
  voices/af_alloy.bin
  voices/af_aoede.bin
  ... and 46 more
```

and exits 1. `--inspect` and the conversion summary both report `embedded_sidecar_count` alongside the existing boolean, which is left alone so anything parsing it still works:

```
kokoro-82m-q8_0.gguf   embedded_sidecars=true embedded_sidecar_count=59
kokoro-respec.gguf     embedded_sidecars=true embedded_sidecar_count=5   <- the broken one
```

⚠ **The first version of that check did not work**, and only a deliberate regression revealed it. It compared the output against `planned_sidecar_destinations` — built by the same filtered walk that caused the bug — so it agreed with a run that kept 5 of 59 and exited 0. A check derived from the thing it is checking cannot fail. The expected set is now read from the **input package**, independently of the plan that produced the output.

## Verified

With the carry deliberately truncated, the run fails with the list above and exits 1. Intact, against the shipped Kokoro package with no `--root`:

```
note: reusing the 59 sidecars embedded in kokoro-82m-q8_0.gguf
embedded_sidecars=true
embedded_sidecar_count=59
```

- extracted sidecar tree **byte-for-byte identical** to the input's (`diff -r` clean, 59 files, 38,543,416 bytes)
- output 189,548,384 bytes against 189,611,360 in; the delta is the replaced spec
- the re-emitted package loads and synthesizes

Unchanged paths, each re-checked: `--no-sidecars` still refuses when the spec requires sidecars; `--root` still wins and still suppresses the reuse; a `--root` pointing at a directory without the sidecars fails exactly as it did before this change; the safetensors→GGUF walk is untouched.

The round trip it was blocking now works: re-emitting a package with a spec that declares a new request option produces a package which offers the option *and* still has its voices — verified by driving that option through the C ABI against the re-emitted file with no spec override.


0xShug0:
@christopherthompson81 There is one regression. The new helper does not check whether the payload exists or whether its offsets are valid, so` --inspect` now succeeds on malformed packages that previously failed. You can check with the attached GGUF (need to change .json to .gguf).

[bad-range.json](https://github.com/user-attachments/files/32299940/bad-range.json)

christopherthompson81:
Good catch — fixed in e867b85.

You are right about the cause: the name-only accessor was a second implementation of the reader, not a shortcut through it. It read `audiocpp.embedded_files.names` directly and checked only the names, so it never looked at `offsets` or `data`. The boolean it replaced went through the full reader, which is why `--inspect` used to fail on your file and then started printing a count for a package no loader could read.

The shortcut is now a `with_contents` flag on `read_gguf_embedded_sidecars` rather than a parallel path. With it false, the whole metadata and every byte range are still validated; the only thing skipped is the copy out of the blob — which is all the accessor was for (38 MB per call on the Kokoro package). `gguf_has_embedded_sidecars` takes the same path and stops copying too.

Against the GGUF you attached:

```
$ audiocpp_gguf --inspect bad-range.gguf
gguf=.../bad-range.gguf
tensors=1
rank0_scalars=0
error: GGUF embedded sidecar byte range is invalid
$ echo $?
1
```

There is a regression test now (`test_malformed_embedded_sidecars_are_rejected_by_the_name_reader`): it builds a package with one sidecar, overwrites its offsets with a range past the end of the blob, and requires both accessors to throw. Reverting only `tensor_source.cpp` and rebuilding makes it fail with `name reader accepted an out-of-range sidecar byte range`, so it is checking something the fix actually changed.

0xShug0:
Thanks! PR merged.

PR #582, opened 2026-09-16
ASR: real native streaming for Nemotron 3.5 + Granite Speech; granite encoder graph cache

drzsdrtfg:
# ASR: real native streaming for Nemotron 3.5 + Granite Speech (and a granite graph cache)

## Summary

Two ASR families shipped a `streaming` mode that never streamed. Both `process_audio_chunk`
implementations only buffered incoming audio and returned an empty event — all decoding
happened in `finalize()` over the whole buffer. Clients of `/v1/audio/transcriptions/live`
therefore saw **zero partials while the user was speaking** (nemotron emitted its deltas as
one burst after the stream closed; granite emitted nothing at all until the final result),
which defeats the point of the live endpoint.

This PR makes both sessions stream the way the model publishers intend, reusing machinery
that was already in the tree:

### nemotron_asr — incremental chunk pipeline

The decoder was already fully incremental *inside* `decode_streaming()` (stateful greedy
TDT loop with `on_text_delta` per token), and the encoder already had cache-aware
`encode_stream_chunk()` + `NemotronEncoderStreamState`. The session simply never used them
per chunk. It now:

- feeds each native chunk (`subsampling × (lookahead + 1)` mel frames, the publisher's
  geometry) through `encode_stream_chunk()` with a persistent `NemotronEncoderStreamState`,
- decodes it incrementally and emits `transcript.text.delta` partials per chunk,
- zero-pads the final partial window at `finalize()` so the tail is encoded instead of
  dropped (previously up to `win_length + hop·(lookahead+1)` samples of trailing audio
  never reached the encoder),
- exposes the decoder state machine as a public incremental API
  (`begin_stream_decode()` / `decode_stream_chunk()` / `finish_stream_decode()`), with
  `decode_streaming()` reimplemented as a thin wrapper over it — pull-style callers keep
  their exact behavior.

### granite5asr — chunked CTC streaming (the publisher's recipe)

`Granite5ASRStreamingSession` was a stub: buffer + `transcribe_audio()` at close. It now
implements chunked TurboCTC inference: every center chunk
(`granite5asr.center_chunk_sec`, default 1 s) is re-encoded together with its left context
(`granite5asr.left_context_sec`, default 2 s) and CTC-greedy decoded with a continuous
collapse across window boundaries (running previous-token state), streaming a partial per
chunk. The window→CTC-frame mapping accounts for the full token stride
(`hop_length × stack_factor × 2^|subsample_layers|` — the encoder subsamples twice), which
was the subtle part: mapping frames at the mel rate silently decodes nothing from window 2
onwards. Both options are declared in `model_specs/granite5asr.json`.

### granite5asr CPU optimization: encoder graph cache + cache-aware streaming

`transcribe_features()` rebuilt the entire 16-block conformer graph (context + gallocr +
graph) on every call. Chunked streaming re-decodes the *same* window shape every chunk and
repeated offline turns cluster around a few lengths, so graphs are now cached per input
shape (small LRU, 6 entries).

More importantly, the streaming session no longer re-encodes a left-context window per
chunk. Granite's attention is block-local (`context_size` 128-frame blocks attend only
within themselves — the rel-pos slice is block-local too), so a chunk needs no attention
cache and no left context: each center chunk is encoded exactly once, prepended only with
a 0.3 s waveform carry covering the two stride-2 subsample blocks' convolution reach, the
avg-pool pair alignment, and the frontend's reflect padding; tokens fully inside the carry
are skipped (one CTC token = hop x stack x 2^|subsample_layers| = 1280 samples). Every
frame is decoded exactly once, which also eliminates the word-duplication artifacts
windowed re-encoding showed at small left contexts. `left_context_sec` is removed;
`center_chunk_sec` stays as the partial cadence (center 2 s trades ~20-25% more
end-of-turn latency for ~30% less total compute).

## Measured behavior (v0.8.0 → this branch, same machine)

6.4 s English sample, `/v1/audio/transcriptions/live`, paced realtime upload, SSE timing
client-side; "end-of-turn" = last audio byte → `transcript.text.done`:

| model | threads | partials while speaking (v0.8.0 → PR) | end-of-turn → final text (v0.8.0 → PR) |
|---|---|---|---|
| nemotron 0.6B | 1 CPU | none (burst after end) → streamed per chunk | ~3934 ms → ~617 ms |
| nemotron 0.6B | 2 CPU | none (burst after end) → streamed per chunk | ~1845 ms → ~312 ms |
| nemotron 0.6B | 3 CPU | none (burst after end) → streamed per chunk | ~1423 ms → ~253 ms |
| nemotron 0.6B | 6 CPU | none (burst after end) → streamed per chunk | ~1170 ms → ~201 ms |
| nemotron 0.6B | Vulkan | none (burst after end) → streamed per chunk | ~567 ms → ~91 ms |
| granite 470M | 1 CPU | none at all → streamed per chunk | ~1841 ms → ~1007 ms |
| granite 470M | 2 CPU | none at all → streamed per chunk | ~923 ms → ~347 ms |
| granite 470M | 3 CPU | none at all → streamed per chunk | ~671 ms → ~261 ms |
| granite 470M | 6 CPU | none at all → streamed per chunk | ~465 ms → ~146 ms |
| granite 470M | Vulkan | none at all → streamed per chunk | ~148 ms → ~72 ms |

Offline transcripts are unchanged (checked on the reference samples), and offline RTFs are
unchanged (the granite graph cache only removes fixed build overhead; on repeated
same-shape requests the encoder no longer rebuilds the graph at all).

## CPU profiling at low thread counts (1-3 threads)

Both models are compute-bound with no significant waste left in the graph implementations:

- nemotron offline @ 1 thread: encoder compute 2218 ms of 2391 ms total (93%), decoder 167 ms, frontend 5 ms. Thread scaling is near-ideal (1T→2T = 2.0x).
- granite offline @ 1 thread: encoder compute 1691 ms of 1710 ms (99%).
- granite streaming: windowed re-encode amplifies compute ~2.9x vs offline (3 s window per 1 s of audio). The `granite5asr.center_chunk_sec` / `left_context_sec` knobs trade this off: center 2 s cuts total streaming encoder compute ~30% (3390 vs 4807 ms @ 1T, 4 windows vs 7) at equal accuracy, but grows the final flush window and adds ~20-25% end-of-turn latency, so the 1 s default stays for latency; left contexts below 2 s duplicate words across window boundaries. These options are currently only reachable through the CLI (--session-option); the server does not expose config/request session options — worth addressing upstream.

Remaining, larger follow-ups (not in this PR):

1. **Lookahead hold for granite streaming**: each frame is now decoded exactly once, but frames near a chunk boundary lack right context and can drop a boundary word (observed: one word on the reference sample). Holding the last tokens of a chunk until the next chunk confirms them (small right lookahead) would recover those.
2. ~~Sub-q8_0 encoder kernels~~ **measured: no gain on AVX2 without VNNI.** Implemented and benchmarked (weight re-quantization to q4_0/q4_k/q6_k via the existing storage pipeline): nemotron offline @ 1 thread q8_0 2623 ms vs q4_k 3376 ms (slower) vs q4_0 2825 ms; granite streaming unchanged within noise; transcripts unchanged even at q4_0. Comet Lake has no int8 dot-product instructions, so q8_0 GEMM is already the efficient path. The weight-type plumbing is kept (opt-in) for VNNI-capable CPUs.
3. **Server-side session-option exposure** (config or /live query) so the tunables are reachable outside the CLI.
4. **VAD-gated offline for nemotron** (granite already supports `audio_chunk_mode=vad`; measured neutral on dense speech, helps pause-heavy real turns).
5. **Lookahead-hold for granite boundary words** — attempted and analyzed: decoding each window with 1-2 future tokens and holding their tokens recovers the dropped boundary word but re-decodes the previous word at shifted CTC alignments (the same audio region aligns differently under a different attention window), producing duplicated words at every boundary. Reverted; the windowed left-context design (whose boundary drops are rarer) is the better trade-off until a CTC re-scoring pass exists.

## Known issue found while testing: short-turn empty text (pre-existing, upstream)

While validating the plugin integration end-to-end we found that the
streaming decode under-contexts each window's trailing `lookahead`
frames: a 0.85 s "Hello" decodes only blank tokens over /live (the word
emerges only when a window tail finally provides future context), while
the offline full-context decode of the same audio is correct. This
reproduces on the SHIPPED v0.8.0 build as well (both 0.85 s and 1.3 s
clips finalize empty in streaming mode), so it predates this PR - the
PR's per-chunk streaming makes the mid-stream blanks visible as missing
partials instead of a single empty final.

The proper fix is center-only emission with overlapping windows (each
window = [center | lookahead] encoded frames, advance by center only,
the lookahead region re-encoded by the next window). We prototyped it;
it requires the encoder graph builder to support overlapping window
advances (the current attention/prefix cache bookkeeping assumes
contiguous non-overlapping windows - a time_mask dim mismatch arises
for the new shapes), so it is out of scope here and filed as follow-up
work. Downstream, the integrating application should treat the offline
POST as the authoritative final transcript for each turn and use /live
only for live partials.

## Notes

- `decode_streaming()` keeps its old signature and semantics; the new incremental API is
  additive.
- The nemotron finalize path keeps the old "shorter than the first required chunk" error
  for streams that end before the first native chunk.
- The granite continuous CTC collapse can (rarely) revise an earlier word when a later
  window disagrees; in that case the current full transcript is re-emitted as the delta
  (documented in the code). The final transcript is always the authoritative text.

Requesting review — happy to adjust the granite defaults or fold the incremental decoder
API differently if you prefer a different split.


drzsdrtfg:
work in progress do not review yet.

drzsdrtfg:
## Follow-up fixes pushed to this branch (2026-09-18)

All measured on the same machine (6C/12T), real-time-paced streaming upload, median of 5 turns.

- **Whole-buffer streaming path now flushes the final partial window** (`84b300a`): the chunk producer dropped the tail after the last full window, so turns whose speech ended inside that tail lost the last word(s) — short utterances ("Hello") came back empty. The tail is zero-padded like the reference processor's final chunk. Also fixes the below-first-chunk request failing outright, and adds `nemotron_asr.decoder.token_ids` TRACE output for empty-transcript diagnosis.
- **Configurable streaming latency knobs on `/v1/audio/transcriptions/live`** (`02b9280`): `lookahead_tokens` (model right-context; chunk duration = (lookahead+1) x 80 ms on nemotron) and `stream_chunk_ms` (ingest batching bound). The session policy batched a full second per read, which delayed every partial by that much regardless of the model window.
- **Lookahead-0 crash fixes** (part of `02b9280`): the second streaming window at lookahead 0 starts at `hop - n_fft/2 = -96` samples (before the signal) — zero-pad the left context instead of indexing before `begin()` (heap corruption); streaming encoder graphs get a smaller default arena and a bounded warm-up ladder (~56 prefix variants at lookahead 0 exhausted commit); cross-graph cache hand-off is layout-checked by name and moved to host staging.
- **`lookahead_tokens=1` (160 ms chunks) accepted, `0` rejected** (`88bd897`): 160 ms decodes correctly (validated on three phrases at two ingest cadences) and cuts end-of-turn latency on a 1.9 s phrase from 423 -> 277 ms (1 thread), 130 -> 100 ms (6 threads). Lookahead 0 measured wrong text *and* no speed benefit, so it is rejected with a clear 400.
- **TCP_NODELAY on the live endpoint's accepted sockets** (`88bd897`): small SSE events were Nagle/delayed-ACK-stalled while the client was still uploading — first partial on the wire for a 1.9 s phrase went from 1861 ms to 526 ms (1 thread) / 356 ms (Vulkan).

Ingest cadence below the model window size changes nothing (window completion drives the decode schedule), so the plugin-recommended config stays `stream_chunk_ms=160` at the default 320 ms window.

drzsdrtfg:
## Follow-up: short utterances came back empty on the live path — root cause and fix (`24b7b7e`)

**Diagnosis** (measured 2026-09-18, 6C/12T CPU): the incremental schedule emits every encoded frame the moment its 4-frame chunk is done, so chunk-end frames are encoded without their lookahead right-context. The offline endpoint decodes the *identical* audio correctly while the live path returns blank:

| turn (preroll 0.2 s + speech + tail) | live (la3) | offline |
|---|---|---|
| 0.55 s "Okay" | `''` | `Okay.` |
| 0.53 s "Hello" | `''` | `Hello` |
| 0.47 s "Yes" | `''` | `Yes.` |

Long turns are unaffected (most frames sit mid-chunk). Reducing the chunk to 160 ms (`lookahead_tokens=1`) narrows the loss but does not close it (`'Hel'`).

**Fix**: `finalize()` re-decodes turns of up to 1 s — and any turn whose incremental final is blank — through the **offline encoder** (one full-context pass over the buffered turn; note `run_streaming_audio` does *not* help, it shares the chunked graph). Cost is bounded by the turn length: 310-370 ms EOT at 2 threads for sub-second turns; long turns keep the streaming fast path (1.9 s phrase: 99-131 ms, 5 live partials). The full 18-case short-turn matrix passes with the fix (the only two failures are zero-preroll turns, which fail offline too).

drzsdrtfg:
## Lookahead-0 verified against the official reference implementation

Validated this branch's nemotron streaming against the NVIDIA-contributed reference in transformers (`nemotron3_5_asr`, the checkpoint's own model class, which carries both the language-prompt projector and the streaming/cache machinery). Harness: chunked encoder driven with the reference chunk contract (first chunk `1+8*la` mel, subsequent `8*(la+1)` mel, final chunk zero-padded), f32 weights, greedy RNNT decode.

Findings:

- Mel windows produced by this implementation are identical to the reference (within float tolerance) for lookahead 0 at both 80 ms and 320 ms native chunks.
- First-chunk encoder output matches the f32 reference at cosine 0.9997 (after aligning the language prompt id, `en-US` = 0).
- Remaining per-frame divergences are isolated (a minority of frames in prefixed chunks, cos 0.6-0.9), plausibly q8_0 quantization + accumulation-order amplified by near-tied attention; transcripts still match on hello/okay/hi/capital/count.
- With the gallocr aliasing fix + the 4-frame chunk floor, `/v1/audio/transcriptions/live` at `lookahead_tokens=0` is deterministic and correct: 25/25 across 5 phrases x 5 repeats (previously flip-flopping).

Lookahead 0 native 320 ms vs lookahead 3 (same binary, end-of-turn -> final transcript, realtime-paced upload):

| config | hello | okay | hi | capital | count | story |
| --- | --- | --- | --- | --- | --- | --- |
| la3 cpu x1 | 208 | 201 | 612 | 432 | 220 | 428 |
| la0 cpu x1 | 421 | 403 | 614 | 206 | 428 | 216 |
| la3 cpu x2 | 108 | 96 | 306 | 218 | 102 | 234 |
| la0 cpu x2 | 210 | 220 | 323 | 100 | 224 | 106 |
| la3 cpu x3 | 75 | 73 | 223 | 183 | 84 | 171 |
| la0 cpu x3 | 156 | 161 | 236 | 102 | 157 | 91 |
| la3 vulkan | 54 | 66 | 138 | 119 | 58 | 120 |
| la0 vulkan | 103 | 88 | 125 | 63 | 93 | 54 |

(All ms, median of 3.) Known edge: a quiet trailing syllable can be dropped at la0 (story -> "Tell me a sto") that la3 still catches; the f32 reference decodes the same audio fully.

drzsdrtfg:
## Root cause of the remaining per-frame divergence: aliased hand-off buffers (fixed)

Dumped every hand-off tensor (NEMOTRON_DUMP_CACHES) and diffed against the reference post-chunk caches. Result: attention K/V next-cache tensors were views over base-node gallocr regions; the allocator reuses a base region once the base node's last in-graph use passes, and it does NOT extend the region's lifetime for later reads through a view. The zero-weighted dependency chain only pins the view nodes, not the base buffers. Measured: 42 overlapping ranges, exact pointer collisions (layer-12 next-K == layer-17 next-conv buffer, layer-18 next-V == layer-20 next-conv). The hand-off copied that overwritten content into the next chunk's attention prefix.

Fix: wrap every hand-off tensor (3 subsampling + 24 conv + 24 K + 24 V per chunk) in an unconditional ggml_cont so each owns its region. After the fix:

- zero buffer overlaps;
- every encoder frame of every chunk matches the f32 reference at cosine 0.9987-0.9997 (was 0.68 on the worst frames) — pure q8_0 quantization noise;
- quiet trailing syllables decode: 'Tell me a story' at lookahead 0 via /live (previously 'Tell me a sto');
- all six phrases correct at la0 and la3 on cpu x1/x2/x3 and vulkan.

Post-fix end-of-turn -> final (median of 3, same binary both modes):

| config | hello | okay | hi | capital | count | story |
| --- | --- | --- | --- | --- | --- | --- |
| la3 cpu x1 | 208 | 201 | 598 | 455 | 203 | 414 |
| la0 cpu x1 | 422 | 409 | 606 | 207 | 416 | 211 |
| la3 cpu x2 | 103 | 97 | 359 | 272 | 98 | 236 |
| la0 cpu x2 | 224 | 205 | 298 | 108 | 196 | 137 |
| la3 cpu x3 | 78 | 79 | 280 | 179 | 87 | 175 |
| la0 cpu x3 | 149 | 152 | 218 | 73 | 144 | 92 |
| la3 vulkan | 55 | 53 | 131 | 110 | 54 | 112 |
| la0 vulkan | 93 | 88 | 129 | 55 | 106 | 55 |

drzsdrtfg:
## Speculative flush + synthesized end-of-turn silence

Measured follow-ups to the tail-wait analysis:

- The trailing-token requirement is ~500 ms of post-speech silence in the encoded stream; the content is irrelevant (the reference wavs' own tails are digital zeros). The 32-mel flush window capped synthesizable silence below the requirement, which is what forced the client tail wait.
- The greedy RNNT has cut-length dead pockets (hi.wav at 560-570 ms decodes empty in the f32 reference too). The blank-final fallback now re-decodes with ~500 ms of appended silence (retry at 1.5 s) and escapes them.
- Speculative flush: ~150 ms of trailing silence (10 ms windowed RMS gate) triggers the flush encode on the ingest thread; finalize reuses the frames unless speech resumed (state snapshot/restore). Flush graphs prebuilt per ladder prefix.

All phrases correct at client tail 150/0/250 ms, both lookaheads.

End-of-turn -> final (median of 3, MinTail emulated at 150 ms):

| config | hello | okay | hi | capital | count | story |
| --- | --- | --- | --- | --- | --- | --- |
| la0 cpu x1 | 357 | 341 | 903 | 401 | 348 | 351 |
| la0 cpu x2 | 175 | 185 | 446 | 162 | 170 | 166 |
| la0 cpu x3 | 141 | 128 | 328 | 115 | 122 | 131 |
| la0 vulkan | 59 | 66 | 136 | 66 | 63 | 66 |
| la3 cpu x2 | 158 | 185 | 446 | 184 | 182 | 199 |
| la3 cpu x3 | 130 | 111 | 358 | 125 | 144 | 152 |
| la3 vulkan | 62 | 60 | 136 | 87 | 62 | 71 |

User-perceived (tail wait + EOT->final) drops ~40-190 ms vs the previous build at the same thread counts; on Vulkan EOT->final itself is ~60-90 ms. hi at short tails hits the greedy-decode pocket and pays the offline fallback (~2 extra encodes) - a known edge; on Vulkan it still nets faster than before. x1 CPU: the 8-frame flush is expensive enough that keeping the old MinTail=250 is reasonable there.

drzsdrtfg:
## Fallback removed; final no-fallback numbers

The blank-final offline re-decode is removed: the streaming decode is the
result. Pocketed utterances (verified to fail in the f32 reference in
every streaming shape) now return an empty transcript instead of paying
250-330 ms for the offline rescue at end of turn.

EOT->final, median of 3, MinTail emulated at 150 ms, lookahead 0 and 3:

| config | hello | okay | hi | capital | count | story |
| --- | --- | --- | --- | --- | --- | --- |
| la0 cpu x1 | 380 | 334 | 357* | 418 | 347 | 375 |
| la0 cpu x2 | 181 | 178 | 167* | 175 | 183 | 175 |
| la0 cpu x3 | 148 | 130 | 143* | 146 | 140 | 144 |
| la3 cpu x2 | 179 | 170 | 192* | 191 | 186 | 196 |
| la3 cpu x3 | 140 | 150 | 145* | 136 | 141 | 157 |
| la0 vulkan | 109 | 109 | 100* | 118 | 108 | 112 |
| la3 vulkan | 110 | 101 | 97* | 116 | 112 | 115 |

(*) hi at this cut length is a greedy-decode dead pocket that reproduces
in the reference; it returns '' immediately. With the previous MinTail of
250 ms the same cut decodes 'Hi. ' normally. Full analysis:
docs/nemotron_streaming_end_of_turn.md

drzsdrtfg:
I hate this

0xShug0:
@drzsdrtfg Thanks for digging into it and sharing your findings. The issue is real,  and your investigation helped clarify the scope. I really appreciate the time you put into investigating the issue.

I picked it up and opened #630 as a first step. We can iterate on it further in follow-up PRs.

PR #591, opened 2026-09-17
cli: add --out-format to choose the WAV sample format of --out/--out-dir

niksedk:
## Why

Generative models can decode samples above ±1.0. The CLI always wrote `--out` / `--out-dir` audio through `WavPcm16Sink`, so those samples were hard clipped to 16-bit before the caller could lower the gain or run a limiter.

Example: ACE-Step 1.5 turbo Q8_0, `text2music`, 20 s, seed 42. The float output peaks at **1.1443 (+1.17 dBFS)**. **329 of 1,920,000 samples** go over full scale, and every one of them lands on the 16-bit rail in today's output.

#358 added `WavSink` / `WavWriteOptions` with `pcm24` and `float32`, and noted in its known limitations that they were "reachable from the library API but not yet from a CLI/server flag". This PR adds that CLI flag.

## What changed

- New CLI option `--out-format pcm16|pcm24|float32`, default `pcm16`.
  - Help text (under Outputs): `--out-format pcm16|pcm24|float32  WAV sample format for --out/--out-dir audio, default pcm16`
  - It is parsed before the model loads, so a bad value fails right away: `audiocpp_cli failed: unsupported --out-format: wav (expected pcm16, pcm24, or float32)`
- `FileOutputPolicy` now carries `WavWriteOptions`. `write_wav_output` in `app/workflow/file_sink.cpp` uses `WavSink(options)` instead of `WavPcm16Sink()`. With default options `WavSink` produces the same bytes as before (#358 tests this, and it is checked again below).
- The option reaches every place `file_sink.cpp` writes WAV files:
  - offline `--out` and `--out-dir` named outputs
  - streaming event and final outputs
  - batch per-request `--out-dir` files and the merged `--out`
- `emit_task_result` gets a trailing `wav_options = {}` parameter, so the `--pipeline` workflow call sites don't change. `--pipeline` runs don't read the flag, so `audiocpp_cli` warns that it was ignored.
- One row added to the options table in `docs/usage.md`.
- `tests/unittests/test_cli_request_options.cpp` now checks the flag: the default stays pcm16 with hard clipping, `float32` and `pcm24` parse, and an unknown value is rejected.

No change to models, the server, the workflow steps or the writer itself. Peak policy and dither are still not exposed. They could be a separate flag later if wanted.

## Validation

macOS 26 arm64 (Apple M4), Metal backend.

**Build**
```
scripts/build_metal.sh --build-type Release --archs arm64 --native-cpu OFF --deployment-build --model-set custom --models ace_step --target audiocpp_cli
```
I built origin/main (c0b26a5) first and saved its binary as the baseline, then applied the change and built again.

**Run** (the same command for every output; only `--out` and `--out-format` change)
```
audiocpp_cli --task gen --family ace_step --model ace-step-1.5-turbo-q8_0.gguf --backend metal \
  --task-route text2music \
  --text "light cheerful background music for a baking and cooking video, warm acoustic guitar, ukulele, soft piano, gentle hand claps, light glockenspiel, relaxed and friendly, unobtrusive underscore for someone talking, instrumental, no vocals, steady tempo" \
  --lyrics "[Instrumental]" --duration-seconds 20 --seed 42 --request-option bpm=120 \
  --out out.wav [--out-format pcm16|pcm24|float32]
```

**Default output is unchanged**

| Output | SHA-256 | `cmp` vs baseline |
|---|---|---|
| baseline (origin/main), no flag | `d4635bf5…d6d67946` | n/a |
| this PR, no flag | `d4635bf5…d6d67946` | identical |
| this PR, `--out-format pcm16` | `d4635bf5…d6d67946` | identical |

**float32 / pcm24** (read with numpy)

| Output | Format tag | Bits | Channels / rate | Frames | Chunks |
|---|---|---|---|---|---|
| pcm16 (default) | 1 (PCM) | 16 | 2 / 48000 | 960000 | fmt, data |
| `--out-format float32` | 3 (IEEE float) | 32 | 2 / 48000 | 960000 | fmt, fact, data |
| `--out-format pcm24` | 1 (PCM) | 24 | 2 / 48000 | 960000 | fmt, data |

- float32 peak: 1.1443 (**+1.17 dBFS**). 329 samples are over ±1.0.
- pcm16 output: 329 samples sit on the rail (±32767), which matches the float32 overshoot exactly.
- Clamping the float32 samples to ±1 and comparing:
  - against pcm16: correlation 0.999999998, max difference 1.5e-5 (16-bit quantisation)
  - against pcm24: correlation 1.0000000000, max difference 8.9e-8
- Requantising the clamped float32 samples with the writer's pcm16 rule (`lrint(x * 32767)`) gives the pcm16 file exactly, with 0 mismatches.

**Batch path**: `--batch-text-file` with 2 lines, `--out-dir`, `--batch-merge-audio concat --out merged.wav --out-format float32`. `line_1.wav`, `line_2.wav` and `merged.wav` were all format tag 3, 32-bit.

**Unit test**: `cli_request_options_test` passed. It is built with `-DENGINE_BUILD_EXTENDED_TESTS=ON`.

**Timing**: wall time for the whole CLI run is dominated by generation; writing the WAV file takes almost none of it.

| Binary / flag | Wall time |
|---|---|
| baseline, no flag | 83.0 s (first run after the build) |
| this PR, no flag | 69.9 s |
| this PR, pcm16 | 68.3 s |
| this PR, float32 | 63.7 s |
| this PR, pcm24 | 63.0 s |


0xShug0:
@niksedk Merged! Thanks!

PR #593, opened 2026-09-17
fix(clap): materialize contiguous layout before the mel-projection matmul

weidu-tpvision:
ClapAudioConditionerRuntime's STFT frontend (build_logmel_frontend) computes the power spectrogram, transposes it to [batch, time, freq_bins], and feeds that directly into a matmul against the mel filterbank. A TransposeModule result is a view (same buffer, permuted strides), and the CPU backend's ggml_compute_forward_mul_mat requires its second operand to be contiguous -- it asserts GGML_ASSERT(nb10 == ggml_type_size(src1->type)) and aborts otherwise.

This crashes ClapAudioConditionerRuntime::encode_audio() on the CPU backend every time whenever the audio-reference conditioning path is exercised. It does not crash on the CUDA backend, I think CUDA's GEMM path tolerates the non-contiguous operand, while the generic CPU kernel does not.

Fix: materialize a contiguous copy of the transposed tensor before the matmul, via core::ensure_backend_addressable_layout -- the same idiom already used at several other call sites in this file and in open_clip_conditioner_runtime.cpp right after their own transposes.


0xShug0:
@weidu-tpvision Thanks! Merged. Did you try to run ControlFoley on CPU?

weidu-tpvision:
I did. I wanted to add clap, I had done it with the microsoft clap using tflite so wanted to export it here, but I found the laion clap in controlfoley with only the audio encoder. I added the text encoder and the tokenizer. If you think it has a value here, I can also do a PR.

0xShug0:
> I did. I wanted to add clap, I had done it with the microsoft clap using tflite so wanted to export it here, but I found the laion clap in controlfoley with only the audio encoder. I added the text encoder and the tokenizer. If you think it has a value here, I can also do a PR.

Let’s find a good excuse to get that in... Any audio-related models that might use it?

weidu-tpvision:
> > I did. I wanted to add clap, I had done it with the microsoft clap using tflite so wanted to export it here, but I found the laion clap in controlfoley with only the audio encoder. I added the text encoder and the tokenizer. If you think it has a value here, I can also do a PR.
> 
> Let’s find a good excuse to get that in... Any audio-related models that might use it?

That is why I didn't do PR directly. I need clap for zero shot classification and there is half of it implemented, so I added the other half. I actually won't run text encoding on device, I need only the text embeddings, so it is not really necessary even for me. It helped me understand the project, and the only excute to include it is that it completes the clap model.

PR #596, opened 2026-09-17
Add deterministic YuE2 plan inputs and fix injected-ABC prefix assembly

mirek190:
## Summary

Two related YuE2 changes:

1. **Opt-in request options** that let a caller drive the pipeline from externally computed tensors so YuE2 runs can be reproduced and compared stage by stage against the Python reference.
2. **A prefix-assembly fix** that these inputs exposed.

## The prefix bug

`token_prefixes()` in `src/models/yue2/pipeline.cpp` closed the symbolic block only when ABC **text** was present:

```cpp
out.insert(out.end(), abc_ids.begin(), abc_ids.end());
if (!request.abc.empty()) {          // only true for abc text / abc_file
    out.push_back(kAbcEndToken);
    out.push_back(kMusicStartToken);
}
```

The helper receives `abc_ids`, so the guard should key off the IDs it actually consumed. When a plan carries ABC token IDs without the text, the returned prefix is missing `ABC_END` + `MUSIC_START`, and the acoustic stage conditions on a prompt that never closes the symbolic block.

This is latent today because the only upstream producer of `abc_ids` is `tokenizer.encode(request.abc)`, so `abc_ids` is non-empty exactly when `abc` is non-empty. It becomes live as soon as the IDs can come from anywhere else — which is what the `abc_ids_file` option in this PR does, and why the fix is included here rather than in a separate PR (the option is not correct without it).

The generated-ABC path was never affected: `generate_semantic()` appends the two tokens itself after the planner returns.

## New request options

All model-local, opt-in, default off, and documented in `docs/models/yue2.md`:

| Option | Direction | Purpose |
| --- | --- | --- |
| `abc_ids_file` / `abc_ids_out_file` | in / out | Exact ABC token IDs (raw int32), bypassing tokenization |
| `semantic_tokens_file` / `semantic_tokens_out_file` | in / out | Exact codec token IDs `0..32767` (raw int32), bypassing AR sampling |
| `latent_in_file` / `latent_out_file` | in / out | Decode precomputed NAR latents, or dump latents before VAE decode |
| `nar_noise_out_file` | out | Resolved NAR noise, `[frames,64]` float32 |
| `nar_velocity_out_file` | out | First `v_theta(x, t)` of the flow solve |
| `nar_tap_out_file` + `nar_tap_stage` | out | Intermediate NAR activation: `embed`, `prehead`, `layer<N>`, `layer<N>.norm|attn|mlp` |

These turn "does the port match the reference?" into a per-stage question (tokens, noise, latents, velocity, layer activations) instead of an end-to-end listen test. `latent_in_file` also gives a bit-faithful VAE path: decoding reference latents reproduces the reference audio exactly.

## Build / run

Built on Windows 11 with the repo's own script (MSVC 14.43, CUDA 12.4, RTX 3090, sm_86):

```powershell
$env:NVCC_PREPEND_FLAGS = "-allow-unsupported-compiler"
.\scripts\build_windows.ps1 -Preset windows-cuda-release -Target audiocpp_cli `
  -ModelSet custom -Models yue2 -CudaArchitectures 86 -Jobs 16
```

```powershell
.\build\windows-cuda-release\bin\audiocpp_cli.exe --task gen --family yue2 `
  --model <path>\Yue2-3B-GGUF --backend cuda --threads 8 `
  --lyrics "<lyrics>" `
  --request-option "style=<style>" --request-option cot=full `
  --request-option "abc_ids_file=<dir>\abc_ids.i32" `
  --request-option "semantic_tokens_file=<dir>\semantic_tokens.i32" `
  --request-option "nar_noise_file=<dir>\nar_noise.f32" `
  --request-option num_inference_steps=32 --seed 831001 `
  --out out.wav --out-dir out
```

Models: `audio-cpp/Yue2-3B-GGUF` (`yue2-3b-bf16.gguf`, `yue2-3b-q8_0.gguf`, `yue2-vae-f16.gguf` + sidecars).
Reference: `m-a-p/YuE2-3B` + `m-a-p/YuE2-Vae` through the Python pipeline, same request and seed.

## Validation

Injected plan/tokens/noise, 32 steps, compared against the Python reference audio:

| Weights | NAR latent cosine | Decoded audio correlation | Duration | Alignment |
| --- | --- | --- | --- | --- |
| `yue2-3b-bf16.gguf` | 0.99994 | **0.9999** | exact | 0 ms |
| `yue2-3b-q8_0.gguf` | 0.99989 | **0.9998** | exact | 0 ms |
| `latent_in_file` (reference latents) | — | **1.0000** | exact | 0 ms |

Same comparison **before** the prefix fix: latent cosine 0.611, audio correlation 0.34. The gap was entirely the two missing prefix tokens, not kernel numerics.

Standalone generation (no injected tensors) is unchanged by these options and by the fix: the same seed and request reproduce the same audio, and the guard change only affects plans with IDs and no text.

A second cross-check that helped isolate the issue: feeding the CUDA sampler's exact tokens and noise into the Python reference reproduces audio.cpp's latents at 0.9998 and even reproduces its occasional decode clicks at the same timestamps, so those are model/VAE behaviour for particular token draws and not a decode bug.

## Known limitations

- The options are raw little-endian files (int32 tokens, float32 tensors) with no header, matching the existing `nar_noise_file` convention.
- `abc_ids_file` requires `cot=melody` or `cot=full`; `semantic_tokens_file` bypasses AR sampling entirely, so `guidance_scale`/ABC options have no effect on that route.
- No automated test is added here: `pipeline.cpp` helpers live in an anonymous namespace, so a unit test would need a small refactor to expose the prefix builder. Happy to add one if you want the shape first.


0xShug0:
just released 0.8.1. stopped the workflows and let me check

mirek190:
Before / after (same tokens, same noise, same 32 steps)

| Metric | Before fix | After fix |
| --- | --- | --- |
| NAR latent cosine vs PyTorch | 0.611 | 0.99994 |
| Audio wave_r vs PyTorch | 0.341 | 0.9999 |
| Q8 audio wave_r | 0.354 | 0.9998 |
| BF16 audio wave_r | 0.341 | 0.9999 |
| Duration / lag | exact / 0 ms | exact / 0 ms |

Both columns use identical injected inputs (`abc_ids_file` + `semantic_tokens_file` + `nar_noise_file`, 32 steps), so the only variable is the prefix fix. The "before" numbers are what surfaced the two missing `ABC_END` / `MUSIC_START` tokens.

mirek190:
I was testing that whole day :) 

Now is perfect reproduction from PyTorch 
Before were totally different results .

mirek190:
Rebased onto `542bb4e` (v0.8.1) so this is mergeable again.

The only conflict was `docs/models/yue2.md`: 0.8.1 moved the request options into their own `## Request Options` table. The new options now live there with the bare-name style, and the CLI table keeps your `--seed` row. No code changed in the rebase.

Validation is unaffected — v0.8.1 does not touch any YuE2 code (`src/models/yue2`, `include/engine/models/yue2` are identical across the rebase), only that docs file.

0xShug0:
@mirek190 I think these parity debug options should be removed. The original dev implementation had similar options, but I removed them during cleanup because they added substantial complexity to the runtime.

The reported prefix issue does not appear to affect current behavior. Normally, users provide ABC text, not pre-tokenized ABC IDs. The code tokenizes that text itself, so `request.abc `and `abc_ids` are populated together and the closing tokens are appended correctly. Generated ABC also appends `ABC_END` and `MUSIC_START` explicitly, while `cot=off `does not consume ABC content.

The problematic state, where `abc_ids` exists without ABC text, cannot occur through the current public interface. It only becomes possible after this PR adds the debugging-only `abc_ids_file` option. Therefore, the PR introduces the path that exposes the issue and then fixes that new path. Could you double check this part? 

The parity investigation is useful, but I do not think these model-specific diagnostic inputs and intermediate tensor dumps should become part of the public runtime.

Update: For parity, using the standalone probes is a better way.

0xShug0:
If you really want to keep some debugging options,  options for "control" are more useful than output-only tensor dumps, such as noise that applications can easily provide themselves. It does not require the Python reference and still produces meaningful output.

mirek190:
You're right, thanks for the pointer to the probes.

I re-checked reachability on `542bb4e` (pre-PR):

- `Yue2Request` carries `abc` (text) and `nar_noise` only. The recognised options are `style`, `lyrics`, `abc`, `abc_file`, `cot`, `nar_noise_file` — there is no way to supply token IDs.
- `plan()` is the only caller of `token_prefixes()`, and it fills `plan.abc_ids` from exactly one source: `tokenizer.encode(request.abc)`.
- The generated-ABC path assigns `plan.abc_ids` later, in `generate_semantic()`, after the prefix has already been built, and appends `ABC_END` + `MUSIC_START` itself.

So `plan.abc_ids.empty() == request.abc.empty()` for every reachable state, and the old check was equivalent — the mismatch only became observable through `abc_ids_file`. One extra note in the same direction: if `encode(abc)` ever returned no tokens for non-empty text, the new check would be the weaker of the two, so it is not a strict superset either.

Removing the diagnostic options and closing this. Parity is being redone as a standalone probe under `tests/yue2/` (same shape as `yue2_vae_parity_probe.cpp`) so nothing model-specific has to live in the request surface — thanks for the suggestion.

My bad ;) too long in from of computer today

mirek190:
Parity now lives in a standalone probe: #598.

`tests/yue2/yue2_nar_parity_probe.cpp` drives `Yue2ArRuntime`/`Yue2NarRuntime` directly, the way `yue2_vae_parity_probe` drives the VAE, and takes everything the acoustic flow consumes from files produced by `tests/yue2/yue2_nar_reference_dump.py` — prefix, codec tokens, the fp32 noise the reference integrated, and the reference latents. No request options involved.

Measured on an RTX 3090: cosine `0.9999` with the bf16 package and `0.9976`–`0.9999` with the shipped q8_0 package, the spread coming from weight-quantization accumulation over long prefixes.

PR #598, opened 2026-09-17
tests(yue2): add a standalone NAR parity probe

mirek190:
Follow-up to #596, which is closed: the reference-only request options are gone, and parity comes back as a standalone probe.

## What this adds

`tests/yue2/yue2_nar_parity_probe.cpp` drives `Yue2ArRuntime`/`Yue2NarRuntime` directly — the way `yue2_vae_parity_probe` drives the VAE — and reads everything the acoustic flow consumes from files:

| File | Contents |
|---|---|
| `prefix.i32` | token prefix: EOD, text, ABC block, `MUSIC_START` |
| `codec.i32` | codec tokens, one per acoustic frame |
| `noise.f32` | `[frames, 64]` fp32 noise the reference integrated |
| `latents_ref.f32` | `[frames, 64]` reference latents |
| `metadata.json` | seed / ode_steps / context / latent_dim |

`tests/yue2/yue2_nar_reference_dump.py` records those tensors from the Python reference. ABC and semantic sampling run greedily so token choice carries no RNG, the noise draw mirrors `yue2.nar.song_chunks`, and the steps/context come from the pipeline generation config. Nothing model-specific is added to the request surface.

## Verification

```bash
python tests/yue2/yue2_nar_reference_dump.py --reference-root reference/YuE \
    --model models/YuE2-3B --vae models/YuE2-Vae --out-dir /tmp/yue2nar
build/bin/yue2_nar_parity_probe --model models/Yue2-3B-GGUF \
    --reference /tmp/yue2nar --backend cuda
```

| Reference | Weights | cosine | rmse |
|---|---|---:|---:|
| 734-token prefix, 1670 frames | q8_0 | 0.999887 | 0.0153 |
| 734-token prefix, 1670 frames | bf16 | 0.999932 | 0.0118 |
| 4144-token prefix, 384 frames | q8_0 | 0.997595 | 0.0680 |
| 4144-token prefix, 384 frames | bf16 | 0.999785 | 0.0203 |

The q8_0/bf16 gap is weight-quantization accumulation over long prefixes rather than a port difference, so the defaults (`--min-cosine 0.995`, `--max-rmse 0.1`) accept the shipped q8_0 package while still catching structural breakage — a wrong prefix assembly measured 0.611 on the same inputs during the investigation.

Tested on an RTX 3090 (CUDA) with both packages, plus a short CPU smoke run. The target sits behind the existing test option and only builds when the model composite links `yue2`, because it uses the model runtime rather than framework code.


0xShug0:
Merged! Thanks!

PR #608, opened 2026-09-19
feat(confucius4_r2t2): add bounded VAD endpointing for long streaming ASR

davidxifeng:
Long streaming dictation currently accumulates audio until Confucius4-R2T2 reaches its audio-tower position limit. This adds opt-in Silero VAD endpointing: pauses finalize the current transcript segment, and a strict audio-length cap splits continuous speech before decoding.

This follow-up builds on upstream main after #604 and adds VAD endpointing and streaming boundary fixes to `confucius4_r2t2`, preserving the merged graph-reuse implementation and metadata filtering. The changes are consolidated into a single commit.

- Process transport input as 32 ms VAD frames so multiple boundaries within one packet are handled separately and segment offsets do not depend on packet size.
- Preserve continuous VAD state across recognizer resets. Include retained gap audio in the segment cap and consume short tails at finalization.
- Publish authoritative segment text and input-rate spans through CLI events and the live server's `transcript.segment.end` event with `reset:true`.
- Add model options, WebUI controls, server example, long-stream probe and regression coverage. Declare `vad_gap_keep_ms` in the spec as well as the runtime.

Validation on Apple M3 / Metal with `models/Confucius4-R2T2` and `assets/resources/sample_16k.wav`:

```sh
cmake -S . -B build/macos-metal-release
cmake --build build/macos-metal-release --target test_confucius4_r2t2_transcription test_confucius4_r2t2_graph_reuse audiocpp_cli audiocpp_server -j 6
build/macos-metal-release/bin/test_confucius4_r2t2_transcription --backend metal
build/macos-metal-release/bin/test_confucius4_r2t2_graph_reuse --backend metal
```

Both model tests pass. Offline and ordinary streaming final text match the golden; default endpointing produces four segments with the expected normalized words. A 3.013 s cap with a 5 s requested gap window produces identical text and exact spans for whole-recording and 17 ms input packets. Assertions cover ordered in-range spans, strict segment caps, event/final-result agreement and metadata-free deltas. The existing graph-reuse growth/shrink and joint token parity tests also pass. JSON configuration and runtime/spec option alignment checks pass.

Endpointing remains disabled by default. Segment text is authoritative and can revise provisional deltas at each boundary. Timestamps describe decoded audio spans, including retained context and VAD detection latency, rather than word alignment. Historical performance numbers in the documentation are labeled as predating graph reuse and VAD-frame scheduling; this PR does not claim new performance measurements. CUDA and Vulkan were not tested.

HTTP live-route validation also passes: a local Metal server received the fixture as chunked s16le PCM via `POST /v1/audio/transcriptions/live?model=r2t2-live&sample_rate=16000&channels=1&sample_format=s16le`, returned HTTP 200, emitted four `transcript.segment.end` events with `reset:true` and valid sample spans, and their joined text exactly matched `transcript.text.done`. Local logs: `/tmp/r2t2-followup-transcription.log`, `/tmp/r2t2-followup-graph.log`, `/tmp/r2t2-followup-server.sse`.


0xShug0:
@davidxifeng I think the PR breaks the existing streaming contract. The fix can stay entirely within the model, without changing the server endpoints, which actually introduces other issues.

Before this PR, long R2T2 streams kept accumulating audio until they exceeded the model’s audio-length limit. Splitting the audio and resetting the recognizer at speech boundaries addresses that problem.

The PR introduces a different problem at those boundaries. In session.cpp, it sends only the shortened text as normal deltas, puts the complete segment text in a separate event, and resets the text offset. A client that appends deltas can therefore miss the end of a segment. The new event in runtime.cpp asks live clients to repair their transcript, but file-backed streaming does not send that event, and existing clients should not need a new response format.

Please keep the VAD-based reset. Before each VAD reset, emit any remaining committed text, including necessary spaces, through `partial_text`. Then reset only the model’s segment state.  Do not change words that were already sent. The test should join all text updates across several resets and compare them with the final transcript. Checking the completed segments alone will not catch missing text.

Issue #503, opened 2026-09-10
[Feature Request] new model: AuK


kooks28:
"AuK is a 1.5B foundation model for speech generation and editing. Trained on millions of hours of diverse audio data, AuK supports zero-shot and instruction-based TTS, content and acoustic editing, paralinguistic editing, speech enhancement, and source separation through a unified natural-language instruction interface."
https://github.com/Tencent-Hunyuan/AuK
https://huggingface.co/tencent/AuK
https://huggingface.co/tencent/AuK-Flash

0xShug0:
@kooks28 Thanks for the pointer. There is a catch: the model uses Qwen2.5-Omni-3B as MLLM encoder so the actual size is much larger.  I also had a bad experience with a Tencent’s model (SongGeneration 2 / LeVo 2 model), which apparently was published just for KPI and turned out to be a complete disaster. Let’s hold off and see more feedback on AuK.

HaujetZhao:
It's not for KPI. It's a revolutionary model! @0xShug0 

Go check out https://auk-project.github.io/

0xShug0:
@HaujetZhao How does Auk compare to FireRed Audio? [FireRedAudio](https://github.com/FireRedTeam/FireRedAudio)? FireRedAudio is in audio.cpp.

HaujetZhao:
I think you should try them both yourself; there's a significant difference between them.

- https://huggingface.co/spaces/tencent/AuK
- https://huggingface.co/spaces/hugging-apps/fireredaudio-tts-demo

In my opinion, AuK excels in editing, to the point where the voice actors no longer need to re-record when lines changes; they simply need to modify the voice using AuK. That's why I consider it revolutionary. It's not a TTS (Text-to-Speech) model, but an audio editing model capable of various adjustments, such as modifying dialogue, extracting clear human voices (removing background noise), isolating specific speakers, and converting regular speech into whispers. I had been eagerly anticipating an open-source version of Adobe Podcast Speech Enhance, and now AuK has arrived, and the ability to extract and enhance voices is just one of its many features.

[...]

PR #612, opened 2026-09-19
AuK: a working port, and evidence for NOT adopting it (draft, closes #503?)

christopherthompson81:
Draft, and deliberately so: **I am recommending against merging this**, and opening it
mainly so the evidence behind that recommendation is reviewable rather than asserted.

#503 asked for AuK. It is ported and verified here. Then I ran its own advertised
capabilities, using its own instruction templates, against audio built so each task had
something to act on. Of the sixteen capabilities on the model card, **four work, five
work partly, five fail, one is inconclusive and one I could not test** — measured on
AuK's own PyTorch implementation, not on this port.

## What the port is

AuK is three models, not one: Qwen2.5-Omni's Thinker as a frozen conditioner, a
Flux2Edit MMDiT, and a causal BigVGAN flow VAE. Each stage was diffed against a PyTorch
dump of the same computation before the next was built on it.

| stage | agreement with the reference |
|---|---|
| VAE decode (causal) | cosine 1.00000000 |
| VAE encode | cosine 1.00000000 |
| Conditioner, 36 layers | cosine 1.00000000 |
| Audio tower, 32 layers | cosine 1.00000000 |
| MMDiT, 10 double + 20 single | v_pred cosine 0.99999925 |
| CFM sampler | cosine 0.99998447 |
| End-to-end waveform | cosine 0.994248 |

Runs on CPU and CUDA; ~29x faster on a 3090 (measured idle, three runs each). Framework
impact is one addition: causal BigVGAN, in its own commit, defaults unchanged.

## The capability results

Instructions are AuK's own, rendered from its `pe.config.yaml` rather than invented.
Fixtures are built per task — noise added at 5 dB SNR for denoising, a 3 kHz band
removed for bandwidth extension (the task is audio super-resolution: reconstruct a lost
top end, as with telephone-band or low-bitrate material), two overlapping speakers for
separation — so each task has something real to do, and for the synthesized degradations
the correct answer is known exactly because we removed it ourselves. Judged by listening, with ASR and signal measurements as
supporting evidence.

| capability | verdict | note |
|---|---|---|
| Zero-shot TTS | works | correct words in the reference voice |
| Instruct TTS | partial | speech is fine, the requested style is not matched |
| Speech content editing | partial | replace/delete/delete-after work; **insert duplicates** ("flowers, of fresh flowers") |
| Lyric editing | **fails** | |
| Pitch editing | **fails** | "worse than changing playback rate" |
| Speed editing | works | full sentence in 4.0 s from a 6.0 s source |
| Volume editing | partial | 3.9 dB delivered of 6 dB requested |
| Emotion | **fails** | audio changes, no emotional alignment |
| Timbre | inconclusive | asked for "a deeper male voice"; the result was judged neither good nor bad |
| De-accent | **fails** | |
| Nonverbal editing | **not testable here** | the fixture has no audible breaths to remove, so the request had nothing to act on — a defect in my harness, not a result |
| Whisper conversion | works | |
| Speech enhancement | partial | denoise and dereverb work; **bandwidth extension fails and degrades** — restores 5.5% of a band deleted entirely (clean source: 54.1%) and sounds worse than its input |
| Speech separation (by order) | works | second speaker removed, ASR-confirmed |
| Music separation | partial | all-voices works, singing-only fails |
| Target speaker extraction | **fails** | keeps the wrong speaker |

⚠ Where both implementations were run — de-accent, target speaker, separation, speed,
the content edits — **they agree, including on the failures.** These are properties of
the model, not of the port.

## Two things worth keeping regardless of the decision

**Duration is an unstated caller contract, and getting it wrong is silent.** The caller
sets output length and the model cannot infer it (#503's known flaw). At exactly the
source length, a deletion gets padded with invented syllables ("a provisionable
provision") and an insertion is declined outright. Give the deletion 10% less and the
insertion 20% more and both act. The reference behaves identically. Two capabilities I
had first scored as broken were my own budgeting errors.

**5 of the 16 task types have no English template**, only Chinese: `enhance_speech`,
`separate_speech`, `improve_quality`, `whisper_edit`, `accent_edit`. Two of the five
behave near-identically in English anyway; three diverge sharply.

## Recommendation

Do not adopt. The cost is 3,900 lines plus a ~14 GB three-model stack; what is reliably
delivered is content editing (minus insertion), denoise/dereverb, separation-by-order,
whisper conversion and TTS — and most of that is already covered by models in this tree.

The **causal BigVGAN** change is independent of that judgement and useful for any causal
decoder; it is a separate commit and could be taken alone.

## Scope of the claim

One fixture set, one English speaker, one seed, 32 steps, CFG 2.0, and no prompt
enhancer (`pe.py` normalizes phrasing into these same templates, so its absence should
not change which capability functions — but that is untested). Nonverbal
editing could not be tested, because the fixture I built for it contains no audible
breaths — the instruction had nothing to act on, which is a harness defect rather than a
model result. Happy to re-check any specific case: the harness makes a
single task about five seconds on GPU, and the fixtures and task runner are included.


0xShug0:
@christopherthompson81 I've finished AuK. All that's left is quant testing. I may merge the code tomorrow.

0xShug0:
@christopherthompson81 Good catch. I actually only checked parity and missed the quality checks. Yeah, some of them indeed aren't working. I'll upload the outputs from both Python and C++ and add a disclaimer. I'm not surprised that this type of feature-rich model has such issues (not to mention it's from Tencent 😅). I think we only need to ensure the logic matches during integration, but can't really help with the model quality.

HF repo is up!

0xShug0:
I actually lost momentum to spend more time on this model.

christopherthompson81:
@0xShug0 My thoughts were that it would be a support nightmare. People would file bugs saying something is broken, The reply would be "Upstream never worked either" and they'd likely be annoyed that we're not fixing the bug they're reporting.

0xShug0:
> @0xShug0 My thoughts were that it would be a support nightmare. People would file bugs saying something is broken, The reply would be "Upstream never worked either" and they'd likely be annoyed that we're not fixing the bug they're reporting.

Exactly...It's a hard call. I decided to keep it for the novelty. I think it'll soon become a retro model, and it'll be fun to see how the model evolves

I'm going to move the model to the community models section for now and list the known issues in the FAQ.

Issue #484, opened 2026-09-07
Feature Tracker and FAQ


0xShug0:
Use this thread to:

Request new features
Pick up open features TODOs
Ask or find answers to common questions

PRs for these features are greatly appreciated!  If you want to work on a TODO, please comment first to avoid duplicated effort.

# Features

- [x] Support mp3 and other inputs/output **(PR ready)**
- [x] HTTPS support **(PR ready)**
- [x] VibeVoice Streaming ASR 1.5B (extending 7B config to support 1.5B)
- [ ] Extend MuScriptor to other variants
- [x] YuE2 AR/NAR LoRA support in cli/server
- [ ]  YuE2 NAR LoRA support in UI
- [ ] Server: cli flag or env variables to set the default model root.
- [ ] NixOS Cachex support (https://github.com/0xShug0/audio.cpp/issues/580)
- [ ] Name based backend selection
- [ ] Other server metrics
- [ ] Multi-GPU support for pooling VRAM https://github.com/0xShug0/audio.cpp/issues/505. May need a careful model-by-model rollout.
- [ ] Better support for Intel GPU and Adreno GPU (https://github.com/0xShug0/audio.cpp/issues/489)

UI:
- [ ] Adding a History of music generations (currently any new generation removes the previous one in the right panel.)
- [ ] Possibility to attach metadata (prompt, lyrics etc) to the audio file, or
- [ ] Exporting the metadata as *.json file, full for reproducibility.
- [x] In HeartMuLa having a default 'pop' in tags would serve better than throwing error: tags needed.
- [x] In ACE-step having a default for prompt 'upbeat ...' would also serve better than throwing error: prompt needed.
- [ ] Any new run stops the play, it would be better if it leaves play alone and only stops it if the new generation done and the new song ready.
- [x] The newly added -1 = auto label could move next to Duration seconds similar to existing Seed -1 = random. 
- [ ] Auto-save gens on computer, instead of clicking save.

# FAQ

### Docker can't run 

This is usually a CUDA driver compatibility issue.  See https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cuda-toolkit-and-corresponding-driver-versions 
<img width="1038" height="116" alt="Image" src="https://github.com/user-attachments/assets/a2d9c556-15c1-4b59-9a55-04aaa5d32f0b" />

### Vulkan
- Useful build tips (from @Dev0Lab1) : https://github.com/0xShug0/audio.cpp/issues/475
[...]

PR #613, opened 2026-09-19
vibevoice_asr_streaming: add the 1.5B size, which the loader already handles

christopherthompson81:
Adds the **VibeVoice ASR Streaming 1.5B** size to the existing `vibevoice_asr_streaming` family. Picked up from #484.

## No code change was needed

The 1.5B checkpoint runs through the existing loader as-is. Layer count, hidden size and head counts are read from the checkpoint's own `config.json`, and the tensor names are identical to the 7B's, so there was nothing size-specific to extend. This PR is package entries, docs, and a scoring tool.

The one structural difference is the vocabulary: `embed_tokens` is `(151936, 1536)` here against the 7B's padded `(152064, 3584)`. It is harmless, because the shape is validated against `vocab_size` from the same config. It would only matter to something that treats the 7B's padded count as a constant — a shared tokenizer bundle, or a logits slice sized for the 7B — so I noted it in the model guide for anyone adding code that spans both sizes.

## Measurements

WER over the four LibriSpeech clips already in `assets/asr_validation/librispeech/`, greedy decode, all rows through one binary:

| package | GGUF size | WER (CUDA) | WER (CPU) |
|---|---:|---:|---:|
| `vibevoice_asr_streaming_7b_q4_k` | 5.86 GB | **4.35%** | 4.35% |
| `vibevoice_asr_streaming_1_5b_bf16` | 5.64 GB | **4.35%** | 4.35% |
| `vibevoice_asr_streaming_1_5b_q8_0` | 3.34 GB | 5.80% | 4.35% |
| `vibevoice_asr_streaming_1_5b_q4_k` | 2.12 GB | 7.25% | 5.80% |

**Please read these as a sanity check, not a benchmark.** Four clips is 69 words, so the entire spread above is two substitutions and one of them moves the number 1.4 points. The like-for-like pair is the two `q4_k` rows, where the 7B leads. The tie between 7B Q4_K and 1.5B BF16 is a coincidence of different clips summing to the same total, not parity — I would not want that row quoted as "the 1.5B matches the 7B".

Worth noting that **1.5B BF16 is smaller on disk than 7B Q4_K** (5.64 vs 5.86 GB). "1.5B" names the decoder; the audio encoder is a large fixed cost that does not shrink with it.

### CPU and CUDA disagree on quantized packages, and that is upstream

The two columns above are not a mistake. A quantized package scores differently per backend, deterministically — six repeat runs per file, and thread counts of 1/4/16, are all stable within a backend. On these clips the entire effect is one fragile word:

```
REF:  I AM FROM THE CUTTER LYING OFF THE COAST
cpu:  I'm from the cutter,  lying off the coast.
cuda: I'm from the country, lying off the coast.
```

Ruled out: it is not a structural quantization problem (both Q4_K models are `{Q4_K: 319, BF16: 857, F16: 1}` with zero rows misaligned to the 256-element k-quant block), and it is not cuBLAS FP16 accumulation (`GGML_CUDA_FORCE_CUBLAS_COMPUTE_32F=1` and `..._16F=1` both leave the output unchanged, so these matmuls are not going through cuBLAS).

It is upstream ggml quantizing the **activations** differently per backend:

| weights | CPU `vec_dot_type` | CUDA (mmvq) |
|---|---|---|
| Q4_K | `GGML_TYPE_Q8_K` — one scale per 256 | `quantize_row_q8_1_cuda` — scale+sum per 32 |
| Q8_0 | `GGML_TYPE_Q8_0` — scale per 32 | `quantize_row_q8_1_cuda` — scale+sum per 32 |
| BF16 | none | none |

Different block size *and* different block format, deliberately. That predicts the table exactly: **BF16 quantizes no activations and its two columns agree; every quantized package diverges.** The 7B not flipping here is four clips, not immunity.

**Nothing to fix in this PR** — I mention it because it means a WER figure for a quantized package is not interpretable without its backend, which is now stated everywhere this branch prints one. Happy to move this note somewhere more permanent if you think it belongs in `docs/gguf.md` at greater length than the sentence I added.

## ⚠ The download repo does not exist yet

The new package ids point at `audio-cpp/VibeVoice-ASR-Streaming-1.5B-GGUF`, following the naming of the 7B repo:

```
audio-cpp/VibeVoice-ASR-Streaming-1.5B-GGUF -> 401
audio-cpp/VibeVoice-ASR-Streaming-7B-GGUF   -> 200
```

So these three ids resolve correctly through the model manager but will 404 on download until that repo exists and the GGUFs are uploaded — which needs your account, not mine. `tools/prepare_vibevoice_asr_streaming_gguf.py --size 1.5b --upload` does the whole thing in one command (see below), and I have the converted BF16/Q8_0/Q4_K files here if you would rather not spend the conversion time.

**This is the one thing that needs a decision from you.** If you would rather not add another repo, say so and I will re-point the entries into the existing 7B repo or `audio-cpp/audio.cpp-gguf` — it is a three-line change to the spec. Everything else in the PR is independent of that choice.

## Surface sync

Per `docs/maintainers/loader_and_catalog.md`:

- `audiocpp_cli --list-loaders --json` emits `vibevoice_asr_streaming`
- `tools/model_manager_v2.py list --json` now shows six packages under that family
- `tools/model_manager_v2.py info vibevoice_asr_streaming_1_5b_q8_0` resolves the repo, file and target directory
- `tools/check_loader_catalog_sync.py` reports `ok` (the remaining warnings are pre-existing `format=safetensors` notices on other families)
- README supported-model row and `docs/asr.md` / `docs/gguf.md` tables updated
- `webui/configs/models_catalog.json` gains a 1.5B entry at `min_vram_gb: 6` against the 7B's 18, since multi-size families there already list each size and low VRAM is the main reason to reach for this one

## Publishing is a script, not a procedure

`tools/prepare_vibevoice_asr_streaming_gguf.py` does the whole thing for either size:

```bash
# convert only, into ./out
tools/prepare_vibevoice_asr_streaming_gguf.py --size 1.5b --out out

# convert, score each package, then create the repo and upload
tools/prepare_vibevoice_asr_streaming_gguf.py --size 1.5b --out out \
    --verify-with build/bin/audiocpp_cli \
    --upload audio-cpp/VibeVoice-ASR-Streaming-1.5B-GGUF
```

It downloads the source from HF (or takes a local `--source`), stages the 8 sidecars, converts to BF16/Q8_0/Q4_K, optionally scores each result, and `create_repo(exist_ok=True)` + uploads. **Nothing is uploaded unless `--upload` is passed.** Bare `--upload` uses the repo id from the model spec for that size.

It stages a curated sidecar set rather than pointing `--root` at the raw snapshot, because `--root` embeds every file under 64 MB that is not itself weights — the raw checkout would put `README.md` and the safetensors index inside the GGUF.

I have run it end to end here apart from the upload step, which needs your token. If you would rather publish by hand, the script is still the record of exactly which flags produced the validated files.

## `tools/asr_wer.py`

Also included, because it produced the numbers above and the assets it scores against are already in the tree. It runs a family over the LibriSpeech clips and reports WER. Two notes from building it, in case they are useful beyond this PR:

- It handles both output shapes — the plain `text_output=` line and this family's multi-line diarized output with `Speaker N:` prefixes. Reading only the first line of the diarized form scored a perfectly working 7B at 100% WER.
- It prints stderr and the exit code whenever the transcript comes back empty, rather than reporting an empty hypothesis as a model result.

Drop it from the PR if you would rather not carry it; nothing else here depends on it.

## Second commit is self-review

I read the diff back before marking this ready and fixed eight things, which is why there is a second commit. Three were doc-convention slips found by comparing against neighbouring rows — the README row rendered its description clause as a third model variant, `docs/asr.md` got two rows for a family where every other family has one, and the webui catalog was missing the 1.5B entirely. The rest were in `asr_wer.py`: a `--threads` default of 16 that was one machine's core count, a per-clip timeout that discarded every already-scored clip, a bare `FileNotFoundError` on a missing reference, and a diagnostic that only printed on a non-zero exit — so a model exiting 0 with no output would still have reported a silent 100% WER, the very thing the parsing fix above was for. Squash if you would rather not carry the history.

Verified the fixes changed nothing measured: 1.5B Q4_K re-scores 7.25%, and a deliberately bad model path now reports `model path does not exist` per clip.


0xShug0:
@christopherthompson81 Do you want to host the GGUFs? Running out of SSD space now, and I've reserved what's left for AuK and YuE2 LoRA training 😅

christopherthompson81:
@0xShug0 Sure, give me a few and I'll put them up.

christopherthompson81:
@0xShug0 Done — they're up here:

**https://huggingface.co/christopherthompson81/VibeVoice-ASR-Streaming-1.5B-GGUF**

| File | Size |
|---|---:|
| `vibevoice-asr-streaming-1.5b-bf16.gguf` | 5.64 GB |
| `vibevoice-asr-streaming-1.5b-q8_0.gguf` | 3.34 GB |
| `vibevoice-asr-streaming-1.5b-q4_k.gguf` | 2.12 GB |

Mirrors the 7B repo: same card metadata and structure, same file naming, flat layout, MIT carried over, upstream revision pinned (`4262d23d8a539a6530cf64fbd0b1751ef9a30853`, verified against what I actually converted from rather than copied). The card adds the WER table and a note that the 1.5B needs no loader changes; trim that if you'd rather it stay a straight mirror.

**I've left the spec pointing at `audio-cpp/VibeVoice-ASR-Streaming-1.5B-GGUF`**, on the assumption you'd rather these live in your namespace than have audio.cpp depend on a personal repo. So: copy them across whenever you have room, or say the word and I'll re-point the three `packages[]` entries at mine instead — either is a one-line change and I'm happy with whichever you prefer. If it's easier, I can also transfer the repo to an org you own, given access.

One thing worth knowing before you copy: **these GGUFs embed the model spec as it stood before this PR added the 1.5B `packages[]` entries**, so their embedded spec lists only the 7B packages. It has no effect on loading — the family, config and sidecars are all correct, and these are the exact files the WER numbers in the PR were measured on. I've deliberately not re-converted for that reason: regenerating would give you bytes nobody has tested. If you'd prefer the embedded spec be current, I'll re-convert and re-validate rather than just swapping the files.

0xShug0:
@christopherthompson81 Appreciate it if you could update the spec to point to your repo. I'm more than happy to see more model repos providing audio.cpp-compatible weights.

christopherthompson81:
@0xShug0 Done on both counts.

**Spec now points at the repo**, and the model guide names it the way the 7B section does. `check_loader_catalog_sync.py` is ok, and `model_manager_v2 install vibevoice_asr_streaming_1_5b_q4_k` pulls it down cleanly.

**I also fixed the embedded-spec wrinkle I flagged rather than leaving it.** The files were rebuilt from the pristine upstream snapshot after this branch added the 1.5B `packages[]` entries, so their embedded spec now carries all six packages with the 1.5B entries pointing at the repo they actually ship from, instead of only the 7B's three.

Rebuilding meant the measured numbers no longer described the published bytes, so I re-scored all six on the new files:

| package | WER (CUDA) | WER (CPU) |
|---|---:|---:|
| BF16 | 4.35% | 4.35% |
| Q8_0 | 5.80% | 4.35% |
| Q4_K | 7.25% | 5.80% |

Identical to the table in the PR. And I ran the loop a user actually takes — install from the repo, confirm the downloaded file carries the six-package spec, transcribe with it — which came back at the published 7.25%.

Happy to move the repo into an org later if you ever want these under `audio-cpp`; the spec change to follow it is one line.

0xShug0:
Merged! Thanks!

PR #614, opened 2026-09-19
webui: make YuE2 LoRA adapters usable — NAR control, persistent store, HF download

christopherthompson81:
Makes YuE2's LoRA adapters usable from the WebUI. This is the open UI item on the feature tracker in #484 (`YuE2 NAR LoRA support in UI`); the cli/server line above it is already checked.

## Three commits

**`d6c855ae` — the missing NAR control.** #586 split YuE2's single LoRA into independent AR and NAR adapters. In `webui/configs/model_params.json` that landed as a *rename* — `lora`/`lora_scale` became `ar_lora`/`ar_lora_scale` — rather than a rename plus an addition, so `yue2.nar_lora` and `yue2.nar_lora_scale` have been reachable from the CLI and server, and documented, but not from the WebUI.

**`b03b444d` — the part that makes it actually usable.** Adding a second text field would have matched AR only by copying AR's defect. Both fields asked for a **server filesystem path**, which is the CLI's answer wearing a browser's clothes: someone using the WebUI is by definition someone without a shell on that machine. And the one browser-native route, `/v1/ui/upload`, writes to a per-process temporary directory that is **removed when the server exits** — so an uploaded adapter left a dangling path in the form after a restart.

**`65e7f7b6` — the engine fix, without which none of this is reachable.** YuE2 could not be loaded from *any* caller that passes the package directory, which is what the WebUI does (`catalog.ts` special-cases `yue2` and `minimax_music3` to send the directory). It was a catch-22:

| directory contents | result |
|---|---|
| main + VAE (the documented combo) | `model directory contains 2 GGUF files …` |
| main only | `yue2.vae_gguf file does not exist` |

The VAE must be present for YuE2 to run, and its presence is what broke directory resolution. `audio-cpp/Yue2-3B-GGUF` ships no `model.gguf`, so nothing rescued a real install either.

The ambiguity check exists for two **variants** of one model — its own comment cites `vevo2-q8_0` beside `vevo2-f16`. YuE2's two GGUFs are different **components** that `yue2.model_gguf`/`yue2.vae_gguf` exist to choose between, but the generic resolver fired before the family got a say. A source that reads "the" GGUF declares a `$gguf` root, and **83 of 87 specs do**; the four that do not — `liveavatar`, `yue2`, `minimax_music3`, `minimax_h3` — name their component files themselves. The check now applies only to sources that need a standalone GGUF.

I also reworded the hint that fires when no spec is found *and* the directory holds several GGUFs: it reported the GGUF count as the problem, which sent me hunting for an ambiguity that was not there.

Verified both directions: YuE2 loads from its directory with no `model.gguf` present, and two variants of a `$gguf` family are still refused with the original message. `model_spec_system_test` passes.

## How it works now

Adapters live in `<model_root>/loras/` and are referenced as `loras/<file>`. That needs **nothing from the engine**: `session.cpp:119` already resolves a relative adapter path against the model root.

**Adapters are identified by reading their safetensors header, never by filename.** The header is a `u64` length followed by that much JSON, so a ranged request reads it without downloading the file — about **40 KB to identify a 140 MB adapter**. The tensor names are the authority, and they are the same names `src/models/yue2/lora.cpp` keys on, so a pre-flight check cannot disagree with the loader:

| signal | meaning |
|---|---|
| `lora_A` / `lora_B` suffixes | unfused — loadable here |
| `diffusion_model.` root, or `lora_up`/`lora_down` | ComfyUI layout — **rejected** |
| `nar_self_attn` / `nar_mlp` | NAR branch |
| no `lora_*` pairs | not an adapter (e.g. a tokenizer head) |

This matters because the public adapter repos ship all of these side by side. On `Mothersuperior/yue2-…-tokenizer-v4`, **8 of 22 files are loadable**; the rest are ComfyUI layouts and tokenizer heads that a filename heuristic would happily offer you. The `_comfyui` files are real LoRAs — they would fail at load time, and only the tensor names tell them apart.

## What changed

- **`app/server/lora_store.{h,cpp}`** — header probing, classification, listing, and a single-file HF download.
- **Five endpoints** under `/v1/ui/loras`: `list`, `upload`, `delete`, `browse`, `download`. Uploads and downloads are **classified before being published**, so a file that would not load is deleted rather than left in the picker.
- **The UI** — both free-text fields become branch-filtered dropdowns labelled with size and rank, plus a field taking a short `owner/model` repo id that lists only loadable adapters with a Download button per file.
- **Build gating** — `lora_store.cpp` builds only with `AUDIOCPP_BUILD_NATIVE_MODEL_MANAGER`, sharing the manager's opt-in HTTP/TLS dependency. A plain server build gains no new dependency; **both configurations were compiled**.

## Two things the review caught

**A write primitive.** `resolve_ui_model_path()` passes absolute paths through and does not confine relative ones. That is fine for the existing handlers, which *read* an operator-configured model — but these handlers *create and delete* files at a path the browser chooses, and `--ui-management` has no authentication of its own. On a LAN-bound server that was an arbitrary-write. Every adapter path is now confined to the models root; an absolute path or a traversal is refused with a 400.

**Browsing a 22-file repo took 28.7 seconds.** Hugging Face redirects `/resolve/` to a CDN host, so connection reuse buys nothing and every probe paid a fresh TLS handshake. One ranged request now covers length and header together, and the probes run concurrently: **2.5s for an identical result**, with the probe count capped so one large repo cannot hold a request open indefinitely.

## Verification

Driven in a real browser, not just typechecked:

- paste a repo id → *"2 of 3 files in … can be loaded here"*, with the ComfyUI file excluded and counted separately
- click Download → the adapter appears in the matching dropdown, already selected
- load with `loras/<file>` → `[info][yue2] yue2.nar_lora: …, projections=196, scale=1.000000`
- a ComfyUI file is refused: *"ComfyUI fused layout; this engine needs the unfused file"*
- path confinement: absolute and `../` model paths both refused
- `npm run check` — 277 files, 0 errors, 0 warnings

I also generated an A/B with and without the NAR adapter at matched seed and settings. It is audible but small — bass up, mids scooped, correlation 0.917 — which is what it should be, since by the time the NAR branch runs the composition is already fixed. Happy to attach samples if useful.

## One thing I found but did not fix here

- **Adapter licensing.** Both YuE2 adapter repos are CC-BY-NC-4.0. One-click download of NC-licensed weights is different from a user placing a file themselves; say the word if you want a notice in the UI.


0xShug0:
@christopherthompson81 My main concern is that the shared server should store, list, and download adapters, not contain logic specific to a model’s LoRA layout. The "loadable" check is misleading. The PR actually marks a public [Whisper LoRA](https://huggingface.co/Be-win/whisper-small-mal-en-lora/blob/main/adapter_model.safetensors) as `"loadable": true, "branch": "ar". ` If we add compatibility checks in the server, they’ll quickly become hard to maintain as more models gain LoRA support.


Also maybe just make the confinement claim more accurate: The new file endpoints can escape the models root through a symlink. The guard compares lexical paths, not where they resolve on disk. I created a symlinked model directory inside a temporary root, then called /v1/ui/loras/delete; it returned HTTP 200 and deleted my test file outside that root. The UI/server are intentionally minimal and don’t include a built-in security layer, so this scenario may simply be out of scope.

christopherthompson81:
Both points were right, and both are fixed in `f04f295f` (rebased onto current main).

## The compatibility check is gone

You're right that it wasn't just wrong on that file — it was wrong by construction. `lora_A`/`lora_B` and `self_attn`/`mlp` are PEFT conventions, not YuE2's, so **every** PEFT adapter for any transformer passed. I read the header of the file you linked to confirm: 288 tensors, `lora_A`/`lora_B` throughout, names like `base_model.model.model.decoder.layers.0.encoder_attn.k_proj.lora_A.weight`, and no metadata to contradict them. Nothing in the classifier ever checked that the names belonged to YuE2, so `"loadable": true, "branch": "ar"` was the only possible answer.

The server now reports the file's **shape**, not its fit:

- `unfused_lora` — true when a file has the `lora_A`/`lora_B` pairs every LoRA-consuming family here reads (`yue2` and `vibevoice` both do), false for a ComfyUI fused layout, which none of them can read.
- The branch sniffing is gone, along with the `yue2_lora_branch` and `lora_rank_separate` metadata keys. There is no model-specific logic left in `app/server`.

Whether an unfused adapter suits a given model needs that model's tensor layout, so the loader answers it — which is where a mismatch already surfaced anyway.

## Which dropdown, without a compatibility check

Dropping the branch verdict would have left both pickers offering every adapter, so the slot is **recorded when the file is stored** rather than inferred afterwards. Adapters are filed under `<model>/loras/<group>/`, with the group named by the caller. The panel passes `ar` or `nar` when it uploads or downloads; each slot offers its own folder.

The server treats the group as an opaque directory name — it neither invents one nor reads meaning into it. A family with one adapter slot passes none and gets the flat layout, which is also what a file dropped in by hand gets, and those are offered in every slot. Relative options already resolve as `model_root / path` (`yue2/session.cpp`), so `loras/ar/<file>` needs nothing new from the loader.

That keeps the maintenance cost you were worried about at zero: a new model with LoRA support picks its own folder names and the server learns nothing about it.

## Path confinement

Fixed rather than scoped out — it was a few lines, which seemed better than arguing about the threat model. The check compared lexical paths, so `<root>/link/loras` read as inside the root however `link` was defined. It compares resolved paths now (`weakly_canonical`, which still tolerates a store that hasn't been created yet).

Two things beyond your reproduction: upload checks the destination and its `.partial` with `symlink_status()` rather than `exists()`, because `exists()` follows a dangling symlink and the write lands on its target; and group names are validated as a single path component.

Your exact scenario, against the built server:

```
delete via symlinked model dir  -> 400, file outside the root survives
upload via symlinked model dir  -> 400, nothing planted
group "../../../escape"         -> 400
model "<root>/../outside"       -> 400

upload group=ar    -> 200  loras/ar/good.safetensors
upload group=nar   -> 200  loras/nar/good.safetensors
ComfyUI fused file -> 400  rejected: this engine needs the unfused file
```

And the Whisper adapter from your comment now returns `unfused_lora` with no claim about which model or branch it belongs to.

One note for anyone verifying: these endpoints are behind `AUDIOCPP_BUILD_NATIVE_MODEL_MANAGER`, so a default build compiles none of this. I only caught that because my first build "succeeded" without compiling `lora_store.cpp` at all.

0xShug0:
@christopherthompson81 I think this PR should focus on the immediate task: adding the NAR LoRA controls to the WebUI.

Persistent uploads are a separate usability feature, and server-side Hugging Face downloads are another substantial feature. Both need more review and validation. This is also a question of ownership and responsibility boundaries. Say if audio.cpp handles LoRA classification, it assumes responsibility for making that classification robust and covering the relevant edge cases. For example the current PR accepts a safetensors file containing only lora_A or only lora_B, even though a valid LoRA requires matching A/B pairs. The model loader already has the tensor layout and remains the actual authority on compatibility. Adding a weaker classifier to the shared server creates two definitions of validity that can disagree: the WebUI may offer and download an adapter as usable, only for the loader to reject it later. 

I suggest limiting this PR to exposing `yue2.nar_lora` and` yue2.nar_lora_scale`, along with any small UI changes required to pass those options correctly, so we can make this feature available soon.

christopherthompson81:
Narrowed to what you asked for: `yue2.nar_lora` and `yue2.nar_lora_scale`, nothing else. **+5 / −3 across 2 files** — down from +1154 / −102 across 11. No C++, no new endpoints. Rebased onto current `main`, so the conflict is gone.

Gone: the persistent store, the HF download, the two `model_spec/package.cpp` commits that only existed to serve the download path, and the classifier.

On the classifier specifically — your A/B-pair point is right, and it is the third wrong answer that thing gave. First the branch verdict, then the Whisper adapter, now accepting a file with only one side of the pair. That is a pattern rather than three bugs, and "two definitions of validity that can disagree" is the accurate description of why. The loader has the tensor layout and is the only thing that can answer the question, so it should be the only thing that tries.

I parked the removed work on [`yue2-lora-store-parked`](https://github.com/christopherthompson81/audio.cpp/tree/yue2-lora-store-parked) rather than deleting it, but I am not proposing it — and the path-confinement fixes in there only mattered for endpoints that no longer exist.

One thing worth your attention in what remains: `dist/index.html` is edited surgically rather than regenerated. A full rebuild in my environment rewrites the whole bundle — 8,683 differing chunks against the committed one, from a different minifier and per-build hashes — so a four-option change would arrive as an unreviewable diff. I built from source and compared the embedded `yue2` block: byte-identical to what a real build produces. Happy to push a genuine rebuild instead if you would prefer the artefact to be authentically generated; it will just be large and none of it meaningful.

christopherthompson81:
Confirmed working. Exercised in the browser against `audiocpp_server --ui` on CUDA with a YuE2 3B package and both adapters present: the four controls render, the pickers work on a server without UI management, and the options reach the session. The "not verified end to end" caveat from the earlier revisions is discharged, and the PR body is updated to say so.

Worth noting how the config-only revision was caught: it was only running it that showed the two options sitting in the bundled catalog with nothing drawing them. Reviewing the diff would not have found it.

0xShug0:
@christopherthompson81 PR merged. I'm confused about the discussion of index.html. Does this mean that, in the normal PR workflow, rebuilding `index.html` and including it in the PR will still cause conflicts? Does that mean #546 will not work unless we manually edit `index.html`?

christopherthompson81:
#546 is working. My earlier explanation needs a correction, though. There can be desync because the build includes model specs and their doc paths. webui/native/src/lib/catalog.ts:29 inlines every spec at build time. Every update to model_specs will impact dist/index.html. So, a rebuilt index.html in a normal PR will frequently conflict.

I think the fix is two parts. 1. Stop inlining the specs. A prebuild step emits one small derived JSON (catalogue families × the four used fields) and catalog.ts imports that instead of globbing. The bundle stops depending on model_specs/ entirely, so adding a model no longer invalidates it unless the model is actually added to the UI catalogue. Keeps the build working without a JS toolchain. 2. Stop committing the bundle, building it on demand in CMake. That ends the conflict class rather than shrinking it. The catch is the fallback at CMakeLists.txt:2477: when the file is absent the build succeeds with a placeholder page, so this has to fail loudly when the UI is requested and node isn't available.

PR #618, opened 2026-09-20
Accept a prompt query parameter on the live transcription route

scriptease:
The multipart transcription route takes a `prompt` field for recognition context, but the live PCM route had no way to pass hotwords. This reads a URL-encoded `prompt` from the query string and forwards it as `text_input` alongside the language, matching the CLI's `--text` behaviour. The `language` parameter is now URL-decoded as well so multi-word values survive.

Changes:
- `app/server/runtime.cpp`: read and decode `prompt` and `language` query params on the live route, pass `prompt` as `text_input`
- `app/server/README.md`: document the `prompt` query parameter
- `docs/community_models/r2t2.md`: note hotword support via `prompt` for R2T2 live transcription

Tested against a local Confucius4-R2T2 server on Apple Silicon (Metal) with domain vocabulary passed via `?prompt=`; terms were recognized that were previously misheard without it.


0xShug0:
@scriptease Thanks! PR merged.

scriptease:
thanks!

PR #622, opened 2026-09-20
fix(voxcpm2): decode streaming patches with left context

mduka:
Fixes #620.

## Summary

With the streaming model id (`"mode": "streaming"`, family `voxcpm2`) the audio clicks at every 160 ms patch boundary wherever the waveform is not near zero — for the PCM/SSE stream at any `stream_batch_size`, and for plain WAV requests to the same id. The offline id is clean.

`generate_once` handed every generated patch (`patch_size` 4 latent frames = 7 680 samples = 160 ms) to the callback on its own, and `run_streaming_request` / `emit_chunk` decoded it alone with `decode_features(patch, 1)`. The AudioVAE decoder is not causal: decoded without its neighbours a patch deviates at both edges from the full-sequence decode, and the hard concatenation is a step at every boundary. The offline path decodes the whole sequence with the prompt's last `kStreamingPrefixLen - 1` patches as context (trimmed via `decode_trim_patches`), so it does not show this. Same class of bug as #429 / #436 for the Higgs Audio codec chunking.

Full analysis and reproduction script in #620.

## Change

Each emitted patch is decoded inside a window of up to `stream_left_context` earlier patches — the prompt's context rows count, exactly as in the offline decode — and up to `stream_right_context` later ones. The session trims the context samples off the chunk audio again via `trim_front_patches` / `trim_back_patches` on `VoxCPM2StreamingChunk`.

Both are request options, defaults **3 / 0**:

| option | default | effect |
|---|---|---|
| `voxcpm2.stream_left_context` | `3` | earlier patches decoded as left context, then trimmed |
| `voxcpm2.stream_right_context` | `0` | later patches waited for as right context, then trimmed; each costs one patch of audio in latency |

`0 / 0` restores the previous behaviour exactly.

The change is contained in the `voxcpm2` family plus its `docs/tts.md` entry — no framework, runtime or backend code is touched. 4 files, +96/−11.

## Validation

**Build**

```
cmake -S . -B build -G Ninja \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo -DGGML_VULKAN=ON -DGGML_NATIVE=ON \
  -DBUILD_SHARED_LIBS=OFF -DAUDIOCPP_MODEL_SET=full \
  -DAUDIOCPP_BUILD_NATIVE_MODEL_MANAGER=ON -DAUDIOCPP_BUILD_SERVER_FRONTENDS=OFF
cmake --build build -j16
```

[...]

Issue #620, opened 2026-09-20
[Bug] VoxCPM2 streaming: audible click at every patch boundary (each 160 ms patch is decoded without context and hard-joined)


mduka:
**Model:** VoxCPM2 (`voxcpm2-q8_0.gguf`), family `voxcpm2`, server model with `"mode": "streaming"`
**Version:** v0.8.1 (f2b49373); the code path is unchanged on current `main`
**Backend:** Vulkan (AMD Strix Halo, gfx1151), Ubuntu 26.04 - the cause is backend-independent

## Symptom

Audio from the streaming model id clicks at every 160 ms patch boundary wherever the waveform is not near zero; in speech it is heard as a click every few seconds. It happens for the PCM/SSE stream (`stream_format: audio`, any `stream_batch_size`) **and for plain WAV requests to the same streaming model id**. The offline model id (same GGUF, `"mode": "offline"`) is clean for the same text and seed.
The seams scale with the signal level: quiet voice designs hide them, a loud voice or a cloned voice shows them clearly.

## Cause

`src/models/voxcpm2/generator.cpp`, `generate_once`, streaming branch: every generated patch 
(`patch_size` = 4 latent frames = 7 680 samples = 160 ms at 48 kHz) goes to the callback on its own:

```cpp
VoxCPM2StreamingChunk chunk;
chunk.decode_features = patch;
chunk.decode_patches = 1;
```

`src/models/voxcpm2/session.cpp`, `run_streaming_request` / `emit_chunk`: the patch is decoded alone
(`decoder_->decode_features(chunk.decode_features, chunk.decode_patches)`) and appended to the previous chunk.

The AudioVAE decoder is not causal. Decoded without its neighbours, a patch deviates at both edges from the full-sequence decode, and the hard concatenation is a step at every boundary. The offline path decodes the whole sequence (with the prompt's last `kStreamingPrefixLen - 1` patches as context, trimmed via `decode_trim_patches`), so it does not show this. Same class of bug as #429 / #436 (Higgs Audio codec chunking).

## Reproduce

No reference voice, no style prefix, fixed seed:

```bash
TEXT="The weather forecast for tomorrow calls for light rain in the morning and clearing skies by noon. Please remember to take an umbrella if you leave the house early."
for MODEL in voxcpm2-stream voxcpm2; do
  curl -s http://127.0.0.1:8082/v1/audio/speech -H 'Content-Type: application/json' \
[...]

PR #623, opened 2026-09-20
fix(voxcpm2): correct the streaming decode rationale, drop right context

mduka:
Follow-up to #622 / #620.

## Summary

#622 fixed the clicks at every 160 ms patch boundary of the VoxCPM2 streaming path, but I explained the cause wrongly there: the PR says the AudioVAE decoder is not causal. It is. `causal_conv1d` pads on the left only and `causal_conv_transpose1d` crops to the causal output length. The clicks came from invoking that causal decoder **statelessly per patch**, so its convolution history restarted from zero padding at every boundary. Decoding the preceding patches in the same window rebuilds that history. Right context cannot change anything for a causal decoder, and the measurements in #620 already showed that ("same as with 0" for every left value).

This PR corrects the rationale where #622 wrote it into the tree and removes what only made sense under the wrong explanation. Nothing in it has shipped in a release yet (last tag is v0.8.1).

## Change

- Comments in `types.h` and `generator.cpp` and the `docs/tts.md` entry state the real cause. Benchmark numbers are gone from the header.
- `voxcpm2.stream_right_context` is removed, together with the tail flush and the back trim it needed. An unknown option is ignored by the server, so old configs that still set it keep working.
- `VoxCPM2StreamingChunk::generated_patches` is again the number of patches generated so far, as before #622. #622 assigned the emitted-patch index, which only coincided with it for right context 0.
- `stream_left_context` is bounded to `[0, 8]` (`kVoxCPM2MaxStreamLeftContext`), checked in `validate_generation_options` and in the request-option parsing. Every emitted patch re-decodes the whole window, so the cost grows with the value while the deviation sits at its floor from 5 on; 16 is already slower than real time on the hardware below.

Default `3` is unchanged. `voxcpm2` family plus its docs entry only, no framework code. 4 files, +58/−79.

## Validation

Build as in #622 (Vulkan, RelWithDebInfo, `AUDIOCPP_MODEL_SET=full`), `voxcpm2-q8_0.gguf`, AMD Strix Halo (gfx1151), served through `audiocpp_server` with one streaming and one offline id on the same weights. Same method as #620: plain English text, `seed` 7, `temperature` 0.7, 8.80 s of audio, 54 patch boundaries, streaming WAV against offline WAV of the same fixed-seed features. The offline reference has one natural step > 0.1 on the grid.

| left | max deviation | samples off by > 0.01 | boundary steps > 0.1 | RTF |
|---|---|---|---|---|
| 0 | 0.5466 | 39.5 % | 6 of 54 | 0.56 |
| 1 | 0.0835 | 2.7 % | 1 of 54 | 0.59 |
| **3 (default)** | **0.0015** | **0.0 %** | **1 of 54** | **0.64** |
| 4 | 0.0005 | 0.0 % | 1 of 54 | 0.71 |
| 8 | 0.0004 | 0.0 % | 1 of 54 | 0.95 |
| 16 (before the cap) | 0.0004 | 0.0 % | 1 of 54 | 1.44 |
| 9, −1 | rejected: `stream_left_context must be between 0 and 8` | | | |

Offline reference: RTF 0.51. The default's output is bit-identical to #622's. The CLI path (`--mode streaming --request-option voxcpm2.stream_left_context=9`) rejects the same way through `validate_generation_options`.

Backend tested: Vulkan. CPU build of the family compiles; CUDA / HIP / Metal untested.

## Follow-ups, not in this PR

- The upstream VoxCPM release 2.0.3 replaced its own 4-patch overlap decode, which is what #622 does, with a stateful `StreamingVAEDecoder` that carries the causal-conv state between patches and decodes one patch per step (OpenBMB/VoxCPM#212: 2× faster VAE decode, half the deviation). That is the long-term shape for this path; in ggml it means state tensors per causal convolution and priming from the prompt latents, so it is a separate change.
- `community_models/voxcpm1` has the same streaming structure; separate PR as suggested in #622.


0xShug0:
Thanks for following up and continuing to improve this! PR merged.

PR #624, opened 2026-09-20
fix(voxcpm1): use the stateful streaming decode on Vulkan too

mduka:
Follow-up to #622 for `voxcpm1`, as suggested there.

## Summary

`voxcpm1` does not need the overlap window from #622: it already has a **stateful streaming decode** (`decode_streaming_step` with `AudioVAEStreamingDecodeState`) that carries the AudioVAE decoder's causal-convolution state from one patch to the next. That is the same design OpenBMB moved to in VoxCPM 2.0.3 (OpenBMB/VoxCPM#212). It was only enabled on CPU, CUDA and HIP. On Vulkan `supports_streaming_decode()` returned false and the session fell back to `decode_features(patch, 1)`, which restarts the causal convolutions from zero padding at every 80 ms patch and clicks at the seams, the defect #620 describes for `voxcpm2`.

## Change

1. `supports_streaming_decode()` adds Vulkan. The stateful graph runs the same decoder ops as the offline graph; its state handling adds only concat, view/cont and backend tensor copies, and it runs unchanged on Vulkan (see below). Metal is untested and stays on the per-patch fallback; the docs entry says so.
2. Lifecycle of the streaming graph (second and third commit). The graph is built against the tensors of the request-local `AudioVAEStreamingDecodeState`, but it lived in the decoder object: neither `~Impl()` nor `release_runtime_memory()` dropped it, so its allocation leaked when the decoder was destroyed, stayed allocated between requests, and referenced the freed state until the next request rebuilt it. Now `release_streaming_decoder_graph()` runs in both, and `run_streaming_request` declares `streaming_state` before its request-scoped release guard, so on scope exit the guard's `decoder_->release_runtime_memory()` drops the graph first and the state is destroyed after it, on return and on exceptions. No extra guard or API. Pre-existing on CPU / CUDA / HIP; fixed here because this PR widens its reach.

`community_models/voxcpm1` plus its docs page only. 4 files.

## Validation

**Build**

```
cmake -S . -B build -G Ninja \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo -DGGML_VULKAN=ON -DGGML_NATIVE=ON \
  -DBUILD_SHARED_LIBS=OFF -DAUDIOCPP_MODEL_SET=full \
  -DAUDIOCPP_BUILD_NATIVE_MODEL_MANAGER=ON -DAUDIOCPP_BUILD_SERVER_FRONTENDS=OFF
cmake --build build -j16
```

**Run.** Package `voxcpm1_0_5b_q8_0` (`voxcpm-0.5b-q8_0-audiovae-f16.gguf`, 16 kHz), backend **Vulkan** (AMD Strix Halo, gfx1151), served through `audiocpp_server` with one streaming and one offline model id on the same weights (`retry_badcase=false` on the streaming id). Plain English text, no reference voice, `seed` 7, `temperature` 0.7, 10.96 s of audio, 136 patch boundaries (1 280 samples = 80 ms each). Same method as #620: streaming WAV against offline WAV of the same fixed-seed features, so every difference is the decoder. First audio from the SSE stream (`speech.audio.done` timing).

| | max deviation | samples off by > 0.01 | boundary steps > 0.1 | RTF | first audio |
|---|---|---|---|---|---|
| before (Vulkan fell back to per-patch decode) | 0.7076 | 47.1 % | 28 of 136 | 0.38 | 90 ms |
| **after (stateful decode on Vulkan)** | **0.0012** | **0.0 %** | **2 of 136** | **0.40** | **90 ms** |

The offline reference itself has two natural steps > 0.1 on the grid, so 2 of 136 is the floor. Offline RTF 0.31–0.32 in both runs. Same audio length in all runs (175 360 samples). Re-measured after the lifecycle changes (third commit): max deviation 0.0012, 0.0 %, 2 of 136, RTF 0.37, first audio 90 ms; five further back-to-back streaming requests on the same session and a CLI streaming run (exercises the destructor path) completed normally.

**Voice clone** (`--voice-ref assets/resources/b.wav` with its transcript, CLI, `--mode streaming` vs `--mode offline`, 7.20 s, 89 boundaries): boundary steps 3 of 89, equal to the offline reference; deviation 0.0022 from the second patch on. The first patch deviates by up to 0.0225 because the streaming state starts from zero rather than from the prompt latents (the offline path decodes the prompt's last patches as context). That is inherent to the existing stateful design and also how the upstream `StreamingVAEDecoder` starts; not changed here.

Backends tested: **Vulkan** (this change) and, for the unchanged behaviour, nothing else. CPU / CUDA / HIP already used this path before. Metal untested.

## Note for `voxcpm2`

With the stateful path confirmed on Vulkan, the natural next step for `voxcpm2` is to port this decoder from `voxcpm1` (the two AudioVAE implementations share the block structure) and retire the overlap window from #622 / #623. That is a larger change and will come as its own PR with before/after numbers.


mduka:
Pushed a second commit: the streaming decode graph is now released together with the request-local state it was built against (`~Impl()`, `release_runtime_memory()`, and at request end in `run_streaming_request`). Before, it leaked on decoder destruction, stayed allocated between requests, and pointed into the freed state until the next request rebuilt it. Pre-existing on CPU / CUDA / HIP, fixed here because this PR widens its reach. Wording in the comment and docs tightened (same decoder ops as offline, Metal untested rather than measured, boundaries "closely" match). Re-measured, numbers in the description.

mduka:
Correction: the second commit as first pushed did not compile (the new public wrapper called a private `Impl` member), which is what the Linux CPU, macOS and Windows checks caught. My local incremental build had failed silently and the numbers I quoted in the previous comment came from the first commit's binary. Amended in place: the wrapper now goes through a public `Impl::release_streaming_decode_graph()`. Rebuilt and re-measured with the amended commit on Vulkan: max deviation 0.0012, 0.0 % of samples off by more than 0.01, boundary steps 2 of 136, RTF 0.37, first audio 92 ms; five back-to-back streaming requests and a CLI streaming run through the destructor path completed normally. The description carries the corrected numbers.

mduka:
Third commit, a simplification with no behaviour change: `streaming_state` is now declared before the request-scoped release guard in `run_streaming_request`. On scope exit the guard's `decoder_->release_runtime_memory()` drops the streaming graph first and the state is destroyed after it, on return and on exceptions. The dedicated guard and the public `release_streaming_decode_graph()` from the second commit are gone; the session no longer needs to know about the decoder's streaming graph. The ordering was already correct before, this only replaces the special construction with the existing cleanup chain. Rebuilt (Debug CPU and Vulkan RelWithDebInfo) and re-measured on Vulkan: max deviation 0.0012, 0.0 %, boundary steps 2 of 136, RTF 0.37, first audio 90 ms; repeated streaming requests and the CLI destructor path clean.

0xShug0:
Merged! Thanks!

PR #626, opened 2026-09-20
kokoro_tts: report the durations it already predicts as word timings

christopherthompson81:
`kokoro_tts` reported no timings — `supports_timestamps` false, `word_timestamps`
empty — so a caller following the audio with a highlight had to spread words
across the buffer by a proxy for length.

The numbers were already there. Kokoro predicts a per-token frame count before the
decoder runs and `expand_tc_by_durations` upsamples the encoder features by exactly
those counts, so a token's share of the output is what the decoder was driven with
rather than an estimate. `PredictorOutputs::durations` sits in the chunk loop beside
`input.input_ids` and the decoded audio — everything needed to say where each spoken
unit landed. **No ABI change**: `TaskResult::word_timestamps` and
`audiocpp_result_word()` already exist and were simply never populated by this family.

## The unit is a phoneme group, not a written word

The label carries the group's phonemes, deliberately. Nothing in this family maps
tokens back to input text: `build_kokoro_synthesis_input` keeps a phoneme string and
ids and no span, and on the supplied-phoneme path (#577) there is no text being
spoken at all. A group — a run between the space tokens Kokoro's vocabulary carries —
is one spoken word on both paths, since eSpeak separates words with a space and a
caller's stream does the same.

A caller whose own G2P produced the stream knows which of its words became which
group and can join the two; one that does not still gets the boundaries. Threading a
word map through eSpeak would be a much larger change and is not needed for either.

Rejected alternatives: per-token timings (~500 entries per chunk, and every caller
would group them the same way), and zipping groups onto the text's whitespace words
on the text path (breaks exactly where it matters — numbers and abbreviations that
expand to several groups).

## `word_timestamps` had to become a TTS capability

It was allowed for `asr` and `align` only, so the spec validator rejected it under
`tts` and `capabilities_from_spec` could never see it. A family whose architecture
predicts durations before vocoding knows where each unit lands — Kokoro does, and so
does anything FastSpeech-shaped — so the token is admitted for the task rather than
worked around. That is the one shared-table change here and the part most worth a
second opinion.

## Details for review

- **The frames→samples scale comes from the chunk's own audio**, not an assumed hop
  length, so it survives a change to the decoder's upsampling ratio and absorbs the
  rounding when a chunk's sample count is not an exact multiple of its frame count.
  Chunks are offset by the audio already merged, so a list of supplied phoneme entries
  produces one continuous timeline.
- **A duration/token count mismatch throws.** It is an internal invariant — the same
  graph produces and consumes them — and timings that look right while pointing at the
  wrong audio are worse than no timings.
- **Confidence stays 0.** The model does not score its own duration prediction and any
  number here would be read as one.

## ⚠ An installed package will still report false

Capabilities are read from the package's embedded schema-v1 contract, and the published
`kokoro-82m-q8_0.gguf` predates this, so `supports_timestamps` stays false until the
package is regenerated. Same position #577 was in, where the answer was to relax the
*validation copy*. That reasoning does not obviously transfer: a stale capability is not
a hard failure, it reads false and a caller falls back. **Flagging rather than deciding
it** — happy to add the equivalent relaxation if you would prefer the runtime's contract
to win for capabilities.

Separately, and not part of this PR: regenerating the package does not currently work.
`audiocpp_gguf --input … --model-spec … --overwrite` reports `embedded_sidecars=true` and
writes a package whose sidecar manifest has no payloads — zero voices extract, and loading
fails on `voices/zm_yunxia.bin`. Reproduced from two source packages and with `--root`
pointing at a tree that contains the file. Same shape as what #579 fixed once. I can open
that as its own issue.

## Verification

Against the ONNX Kokoro — the same model on a different runtime — over the same text with
the same supplied phonemes. Both joined to source words the same way, and word starts
compared as a fraction of each engine's own duration, since the buffers differ slightly in
length:

```
[0] The button was forgotten on the cotton coat.          8 groups   worst 0.00%
[1] She read the schedule aloud at a quarter past three.  10 groups  worst 0.76%
[2] Uranium and aluminium are both elements.              6 groups   worst 0.76%
[3] A quarter of the students answered correctly.         7 groups   worst 0.76%
[4] The harbour was quiet this morning.                   6 groups   worst 0.76%

rows=37  mismatched sentences=0  worst word-start delta=0.76% of duration
```

Sentence [0] is exact. The worst disagreement anywhere is ~18 ms on a 2.4 s utterance,
and where the totals differ the word starts still agree as fractions — so the difference
is in the vocoder's output length, not the alignment. Two independent implementations
landing on the same answer is not something a bug in either would produce.

Driven through the C ABI on CUDA (RTX 3090, sm_86) with `--model-spec-override`, since the
installed package's contract predates the capability.

https://claude.ai/code/session_01KfTCTjC7FTnNm3aS2zszKE


christopherthompson81:
Self-reviewed at `high` before asking anyone else to read it; six findings, four needed code. Pushed as `7024d162` and `1a567ece`.

**The one that mattered: the schema change broke an existing unit test and I had not run it.**
`typed_rejects_unknown_capability` feeds `"capabilities": {"tts": ["word_timestamps"]}` and asserts the validator rejects it — its comment reads *"ASR timestamp capabilities cannot be attached to TTS"*, which is precisely the rule this PR reverses. Confirmed red on the original commit:

```
model_spec_system_test failed: typed_rejects_unknown_capability
  should reject with: unknown capability 'word_timestamps'
```

The negative case now uses `partial_results` (still genuinely ASR-only) so the test keeps testing that capability values are typed per task, and a positive case asserts a `tts` spec declaring `word_timestamps` validates. That needed an `expect_accepts` helper — only `expect_rejects` existed, which is why widening the set had nothing to trip over.

**A punctuation-only group was reported as a word.** The G2P spaces a mark that followed a space in the source, so the opening quote in `She said "hello" loudly.` became its own group — four words, five groups — and a caller joining words to groups in order is one out from there to the end of the chunk. Measured, text path, `af_heart`:

```
before:  words=4 groups=5  [ʃi | sˈɛd | “ | həlˈO” | lˈWdli.]
after:   words=4 groups=4  [ʃi | sˈɛd | həlˈO” | lˈWdli.]
before:  words=5 groups=5  [hi | pˈɔzd | — | ðˈɛn | lˈɛft.]
after:   words=5 groups=4  [hi | pˈɔzd | ðˈɛn | lˈɛft.]
```

Skipped now; the frames are still consumed so the next group starts after the pause rather than on top of it.

**The token/duration check threw.** Wrong severity: the audio is the product and an empty word list is a state every caller already handles, so a defect in a supplementary feature should not cost the render. It reports nothing and traces `kokoro.word_timings_token_mismatch`. The invariant does hold by construction (`predictor.cpp:734`, `:778`), so it was dead code that could only ever fire after a later refactor — i.e. exactly when losing the audio would be most surprising.

**None of it was testable.** `append_kokoro_word_timings` was in an anonymous namespace. Declared in the header, narrowed to take the vocabulary rather than the whole `KokoroAssets`, and given `tests/unittests/test_kokoro_word_timings.cpp` — pad/space boundaries, duration-proportional spans, the cross-chunk offset, both punctuation cases, and the three degenerate inputs. The punctuation case is there because it is the finding a test would have caught and there was none.

**Two I did not change**, and documented instead:

- A sentence-final mark's pause falls inside the last word's span. Fair point that a highlighter keeps that word lit through the silence — but ending the group at the last non-punctuation token would make this engine disagree with the reference implementation and with the ONNX path at every comma. Stated in `docs/models/kokoro_tts.md` rather than changed; happy to flip it if you would rather.
- The capability/report inconsistency against a published package — still the open question above. It is at least no longer silent: the docs state it, and I have documented that `--words-out`'s stdout `word_timestamps=` line is new for TTS runs, since that is a visible change for existing CLI users.

`ctest` 37/37. Cross-engine agreement, speaking rate (0.75/1.0/1.5) and the multi-chunk offset all re-verified after the grouping change — 37 word rows at worst 0.76% delta, and 260 groups across 5 merged entries monotonic and ending at 100.0% of the buffer.

0xShug0:
@christopherthompson81 I'll check the PR later. will be AFK for a few hours...

0xShug0:
@christopherthompson81 I don’t think the PR should make a global schema change, given that it applies only to Kokoro and the returned data is not general --- it's Kokoro-specific phoneme alignment semantics. Since no other TTS models currently need this, it is unclear what the appropriate normalization/contract should be. So it would be better to keep phoneme-group timing output local to Kokoro instead of exposing it as the generic global `word_timestamps` capability.

BTW `--words-out` automatically injects `return_timestamps=true`, so there should be a compatibility layer for the distributed GGUFs; otherwise, it will throw `unknown Kokoro TTS request option: return_timestamps`.

christopherthompson81:
Both taken, and the second one turned out to be worse than reported. Pushed as `58d0b13d`, rebased onto current `main`.

## The schema change is gone

`src/framework/model_spec/schema.cpp` and `tests/unittests/test_model_spec_system.cpp` are back to their upstream state — `git diff upstream/main --name-only` no longer lists either, and the spec no longer declares `word_timestamps`. The diff is kokoro files, its own test, docs and the CMake entry for the test.

The feature stays as an opt-in `return_timestamps` request option on `kokoro_tts`'s own spec. That needed nothing shared: `return_timestamps` is already a registered `bool` in the option registry (`src/framework/model_spec/options.cpp:67`), so declaring it is a per-family change. Off by default, and `audiocpp_model_supports_timestamps()` stays 0 for the family — which also disposes of the stale-package question I flagged, since there is no capability to go stale.

## `return_timestamps` — you were right, and it was broader than distributed GGUFs

It threw on **every** package including the repo's own spec, because I never added the option anywhere. `--words-out` was documented in `docs/usage.md` by this PR while being the one way to make it fail:

```
$ audiocpp_cli --task tts --family kokoro_tts --model .../Kokoro-82M-GGUF --backend cpu \
    --language en --text "..." --voice-id af_heart --out out.wav --words-out words.json
audiocpp_cli failed: unknown Kokoro TTS request option: return_timestamps
```

My verification only ever read the stdout `word_timestamps=` line, which needs no flag — so the flag was never once exercised. Fixed by extending the existing #577-shaped relaxation in `validate_request_options` to the key. Re-run against the published `kokoro-82m-q8_0.gguf`, whose embedded contract predates the option: `--words-out` writes the file, no flag reports nothing, and with `--seed` pinned the audio is byte-identical either way.

## A correction to the PR body

The claim that a phoneme group is one spoken word "on both paths" is **false on the text path**, and it supports your point about the semantics rather than just the process. eSpeak-ng merges function words:

| text | words | groups | |
|---|---|---|---|
| The button was forgotten **on the** cotton coat. | 8 | 7 | `ɔnðə` |
| She read the schedule aloud **at a** quarter past three. | 10 | 9 | `æTə` |
| He said it was **in the** box under the table. | 10 | 9 | `ɪnðə` |
| Uranium and aluminium are both elements. | 6 | 6 | — |
| I went to the store and bought a loaf of bread. | 11 | 11 | — |

Three of five ordinary sentences. The cross-engine check that reported 0.76% worst-case agreement ran on the **supplied-phoneme** path, where the caller's own G2P chose the spacing and the question cannot arise — so it agreed on exactly the path where agreement was guaranteed, and said nothing about the other one.

A caller zipping these onto whitespace-split words is therefore right on some sentences and one out for the rest of the chunk on others. That is now the documented reason the option is opt-in and no capability is declared, and it is pinned by `test_merged_function_words_are_one_group` so it cannot quietly come back. `docs/models/kokoro_tts.md` leads with it instead of the old "on ordinary prose it is one spoken word".

## Verification

`ci/run-local.sh linux-cpu` PASSED, 40/40 including the new case. `check_loader_catalog_sync.py` and `model_spec_system_test` both exit 0.

0xShug0:
@christopherthompson81 Thanks, PR Merged! The GGUF update may need to wait.

0xShug0:
@christopherthompson81 GGUFs updated.

PR #627, opened 2026-09-20
ci: run the CI jobs locally, and actually run the unit tests on linux

christopherthompson81:
## Why

This came out of #540. That fix is for a Vulkan shader that compiles to nothing and only surfaces twenty minutes later as an undefined reference at link — a failure that has never reproduced outside CI. Checking any change to it meant pushing and waiting, and `glslc` is not installed on a typical dev box, so the shader path was only ever exercised through stubs.

## What

`ci/run-local.sh` replays the workflow steps in the same base image.

```
ci/run-local.sh                 # list the jobs
ci/run-local.sh linux-vulkan    # one job
ci/run-local.sh linux           # every job whose name starts with "linux"
ci/run-local.sh all
ci/run-local.sh --clean nix     # discard the build trees and the nix store first
```

Docker is the only requirement — nix runs from the `nixos/nix` image, so it does not have to be installed on the host. Build trees live in `build/ci-local-*` (already gitignored) and are reused between runs; the nix store lives in a named volume so a second run does not refetch the closure.

All four jobs verified:

| job | result |
|---|---|
| linux-cpu | passed, 364s |
| linux-vulkan | passed, 430s cold, 122s warm |
| nix-cpu | passed, 401s |
| nix-vulkan | passed, 440s |

`nix-vulkan` is the job that failed on #539, and it passes against #540's branch: 1932 shader symbols declared, every one defined, no undefined generated symbols in the linked binary.

## Second commit: the linux test step was running nothing

`linux-build.yml` has a step named "Build and run unit tests" that ends in `No tests were found!!!`. `ctest` exits 0 on an empty test set, so the job has been green while running no tests.

The suites are gated behind `ENGINE_BUILD_TESTS`, which the Configure step never sets. `mac-build.yml` has the same step word for word and does set it, so this looks like a dropped flag rather than a decision.

Nothing was hiding behind it — 35 tests, all passing, in both the cpu and vulkan configurations, in about 4 seconds. Worth having in place before that stops being true.

## Notes for reviewers

Two things the workflows depend on that are invisible until the steps are run one at a time, both commented in the script:

- `CC`/`CXX` are job-level `env:` in `linux-build.yml`. The Vulkan shader generator is an ExternalProject that configures itself during the *build* step, so setting them only on the configure command fails with `No CMAKE_C_COMPILER could be found`.
- The nix container runs as root against a checkout owned by someone else, so the flake's git input is refused with `repository path '/src' is not owned by current user` until the path is marked safe.

The cmake flags in `run_linux_job` mirror `linux-build.yml` by hand and have a comment saying so; they need updating together.

https://claude.ai/code/session_01TMmMgd5xNGnjsQgybuUiK3


christopherthompson81:
Self-review before marking this ready. The script had the exact failure it exists to catch, so it's worth writing down.

**It reported a pass when a step failed.** The job functions are invoked as `run_linux_job ... || ok=0`, and a shell function on the left of `||` runs with `set -e` suppressed for its entire body. Every step ran regardless of the one before it, and the verdict came from whichever command happened to be last. With a stubbed `docker` that fails only the loader check:

```
--- check_loader_catalog_sync
FAKE: loader catalog desync
... configure, build and test all run anyway ...
================ linux-cpu PASSED (0s)      exit 0
```

Now:

```
--- check_loader_catalog_sync
FAKE: loader catalog desync
================ linux-cpu FAILED (0s)      exit 1
```

A local runner that fails open is worse than no local runner, because it gets trusted.

**It skipped a step the nix job actually has.** `run_nix_job` never ran `check_loader_catalog_sync`, although `nix-build.yml` has that step — and `run_loader_check`'s own comment claimed both workflows run it. A catalog desync passed locally and failed in CI, which is the thing this PR is meant to prevent. It runs in the ubuntu deps image, since `nixos/nix` has no `python3`.

**Four smaller ones**, all fixed in `cdc3a90`:

- `ubuntu_image_with_deps` ended in `echo "$tag"`, so its exit status was the echo's — a failed `docker build` read as success and the caller ran against a tag that didn't exist.
- The build context was the whole repo for a Dockerfile that copies nothing, tarring `.git` and the multi-gigabyte `build/ci-local-*` trees the script itself creates. It's stdin-only now.
- `--clean` didn't touch the deps image, so editing the apt list to mirror a new package in `linux-build.yml` silently kept the stale image.
- `--clean` wiped the nix store inside each job, so `nix-vulkan` refetched the closure `nix-cpu` had just fetched. Both are discarded once now, before any job runs.

Re-verified after the changes against real docker: `linux-cpu` 293s and `nix-cpu` 379s, 35 tests passing, and the nix job now runs the loader check.

0xShug0:
Merged! Thanks.

